PR #12, opened 2018-08-02
add routes macro

cksac:
It would be great to have some utilities help to compose endpoint filters. This PR added routes macro to do so, thought?

### Steps
- Defined endpoint URL
- Supported method e.g. POST, GET, PUT, DELETE
- Handling logic

### Example
```rust
let api = routes! {
    ["todos"] => |p| {
        // `GET /todos`
        get {
            p.and(db.clone()).map(list_todos)
        };
        // `POST /todos`
        post {
            p.and(warp::body::json()).and(db.clone()).and_then(create_todo)
        };
    }
    ["todos" / u64] => |p| {
        // `PUT /todos/:id`
        put {
            p.and(warp::body::json()).and(db.clone()).and_then(update_todo)
        };
        // `DELETE /todos/:id`
        delete {
            p.and(db.clone()).and_then(delete_todo)
        };
    }
};
// View access logs by setting `RUST_LOG=todos`.
let routes = api.with(warp::log("todos"));
```

### DIFF
--- a/examples/todos.rs
+++ b/examples/todos.rs
@@ -1,14 +1,16 @@
 #![deny(warnings)]
-#[macro_use] extern crate log;
+#[macro_use]
+extern crate log;
 extern crate pretty_env_logger;
 extern crate serde;
-#[macro_use] extern crate serde_derive;
+#[macro_use]
+extern crate serde_derive;
+#[macro_use]
 extern crate warp;
 
 use std::env;
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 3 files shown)


seanmonstar:
Hm, it's interesting what you've done here! Personally, I've wanted to stay away from this style, though. Unless there's overwhelming desire for something like this, I'm leaning towards things like this should be external.

seanmonstar:
I'm going to close for now, per my comments before. Thanks for sharing the idea, though!

PR #15, opened 2018-08-02
add warp::path::tail() filter

### LINKED ISSUE #13: Allow getting request path tail
Like this: https://rocket.rs/guide/requests/#multiple-segments

seanmonstar:
Closes #13 

Something like this, @lnicola?


lnicola:
Yeah, looks good.

lnicola:
I wonder if a segment iterator would make the user more mindful of path traversal attacks. I guess not?

PR #17, opened 2018-08-02
Update README.md example to fix example

jeromebenois:
It seems this example doesn't work without specify #[macro_use] and remove warp:: before macro to fix 'non-ident macro paths are experimental'.


seanmonstar:
Oh right, this was using the new syntax for Rust 2018. Thanks for the fix!

PR #18, opened 2018-08-03
add Handlebars templating engine example usage, closes #1

### LINKED ISSUE #1: HTML support built-in
Needing to output HTML using a templating system is required by many applications. warp should provide an easy default way to do so (clearly, the filter system allows someone to plug in any output they want). It seems like handlebars is the most popular.

jxs:


seanmonstar:
Awesome, this was easier than it looked! With #1, I figured warp should provide handlebars support natively. Do you think with the ease of doing this, it perhaps should not do so?

whitfin:
@seanmonstar I'd agree with that. Since it appears simple enough, and now there's a documented example, it should be enough for anybody without the need for explicit support (which might actually make it more complicated?).

PR #20, opened 2018-08-04
dir: urldecode route before creating path from it

hdhoang:
I think hyper already rules out the `Err` branch by validating the whole path earlier.


seanmonstar:
Awesome, thanks!

Issue #14, opened 2018-08-02
Rationale behind wrapper filters for methods?


kamalmarhubi:
I'm curious about why we write

```rust
warp::get(
    warp::path("a")
       .and(warp::path("b"))
);
```

instead of

```rust
warp::get()
    .and(warp::path("a"))
    .and(warp::path("b"))
;
```

seanmonstar:
Yea, this inconsistency is a little annoying, I agree. The method filters were some of the first ones I wrote, and back then, I hadn't figured out a nice solution to ditching the `()`, so `and`ing them caused this nastiness:

```rust
warp::get()
    .and(warp::path("hello"))
    .map(|((), ())| foobar())
```

It's probably clearer if they just worked the same as all other filters, huh? I knew I'd forget to fix *something* before releasing 0.1...

kamalmarhubi:
Ah! That makes sense. I still haven't looked into the "magic" you're doing in here, mostly just playing around to see what the library is like.

The original title of this issue was going to be about the implementation escaping in the return types of those filters as well, but then I realised they were simply "different". That's probably also something worth fixing at the same time.

What's your position on breaking changes? Do you have a rough timeline for a 0.2 that would batch things like this?

seanmonstar:
The implementation "escaping" was actually because since those functions take a `Filter` argument, whether that filter was `Clone` or `Copy` was lost if the the return value is just `impl Filter`. By returning concrete types, those qualities still exist thanks to `derive(Copy, Clone)` on the combinators. Since the combinators aren't publicly exported, and thus not nameable, I'd like to think that the return type here can change, as long as the returned type still fulfills the same traits.

I'll probably hold off on a breaking change for at least a little bit, so that these papercuts can be grouped up. Fixing these filters could probably be done by putting the new ones in `warp::method::v2` or something...

kamalmarhubi:
Ah I didn't mean `impl Filter`, I just meant a per-filter newtype struct that could still have `derive(Copy, Clone)` on it. This would mean a boilerplate `impl Filter` for each one, but at least you're guaranteed the type names are not changing. The unnameablity is something I didn't consider, :thinking:...

seanmonstar:
Newtypes here can't be done easily, since they take advantage of the combinators and closures. I suppose they could stop using those, but may not be worth the hassle.

PR #21, opened 2018-08-04
Add method::v2 wrapping filters

kamalmarhubi:
This commit adds `Wrap`-based method filters, mirroring the existing
methods.

refs #14

### DIFF
--- a/src/filters/method.rs
+++ b/src/filters/method.rs
@@ -137,3 +137,144 @@ where
     })
 }
 
+pub mod v2 {
+    //! HTTP Method Filters
+    //!
+    //! These filters deal with the HTTP Method part of a request. They match
+    //! the request `Method`, and if not matched, will reject the request with a
+    //! `405 Method Not Allowed`.
+    //!
+    //! These "filters" behave a little differently than the rest. Instead of
+    //! being used directly on requests, these filters "wrap" other filters.
+    //!
+    //!
+    //! ## Wrapping a `Filter` (`with`)
+    //!
+    //! ```
+    //! use warp::Filter;
+    //! use warp::filters::method::v2 as methodv2;
+    //!
+    //! let route = warp::any()
+    //!     .map(warp::reply);
+    //!
+    //! // This filter rejects any non-GET requests before they get to the
+    //! // wrapped `route`.
+    //! let get_only = route
+    //!     .with(methodv2::get());
+    //! ```
+    //!
+    //! Wrapping allows adding in conditional logic *before* the request enters
+    //! the inner filter. In the example, it rejects non-GET requests before
+    //! they get to the wrapped `route` filter.
+    use futures::{future::FutureResult, IntoFuture};
+    use http::Method;
+
+    use filter::{And, Filter, FilterBase, MapErr, WrapSealed};
+    use reject::{CombineRejection, Rejection};
+    use reply::Reply;
+
+    /// Wrap a `Filter` to require that the request method to be `GET`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use warp::Filter;
+    /// use warp::filters::method::v2 as methodv2;
+    ///
+    /// let route = warp::any().map(warp::reply);
+    /// let get_only = route.with(methodv2::get());
+    /// ```
+    pub fn get() -> WithMethod {
+        WithMethod(&Method::GET)
+    }
+
+    /// Wrap a `Filter` to require that the request method to be `POST`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use warp::Filter;
+    /// use warp::filters::method::v2 as methodv2;
+    ///
+    /// let route = warp::any().map(warp::reply);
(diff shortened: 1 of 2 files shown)


kamalmarhubi:
@seanmonstar after a bunch of futzing around figuring out the shape of restrictions, here's a PR! My questions:
- [ ] is the module placement right?
- [ ] should `MethodIs` have an exposed construction function?

seanmonstar:
I'm on mobile right, will look more later, but thanks for jumping on this! I too have wondered about making it work with `and`, you likely noticed that that didn't work currently because of the rejection cancelation. I'm not sure the way rejections work internally are the best, if a better way is found that allows these to work with `and`, that'd probably be the best!

src/filters/method.rs
```diff
@@ -137,3 +137,144 @@ where
...
+    /// let route = warp::any().map(warp::reply);
+    /// let delete_only = route.with(methodv2::delete());
+    /// ```
+    pub fn delete() -> WithMethod {
+        WithMethod(&Method::DELETE)
+    }
+
+    /// Wrap a `Filter` to require a specific HTTP method.
+    #[derive(Debug, Clone, Copy)]
+    pub struct WithMethod(&'static Method);
```
seanmonstar: I don't know how important this is, but I thought I'd explain just for context: the previous `MethodIs` thingy took a closure so that the size of it could be 0, instead of holding a static reference. In this case, `size_of::<WithMethod>()` is 1 word, and cloning (copying) it probably means a copying that pointer? I didn't put any measuring into whether this matters at all, just figured I'd explain the *why* of the older one.

kamalmarhubi: Ah, that's useful context. This wasn't even remotely something I was thinking about. If there is a performance impact, I guess these can be implemented as individual filters, one per method, or by using a phantom type + associated const instead of storing the method in the filter struct.

kamalmarhubi:
I'll be honest, I'm kind of lost on the rejection / cancellation stuff, and didn't spend too much time trying to figure it out. I was trying to work within the confines of the APIs that currently exist, to get a feel for them more than anything else. That said, I definitely saw that it prevented the "normal" filter approach for these.

I'm wondering: if using `and` with method filtering is necessarily the best idea: depending on where it's added in the chain, a lot of processing could happen before it gets a chance to reject. The thing I like about using a `Wrap` type for this is that you can't "get it wrong": wherever you call it, method checking will be inserted at the start of the chain. Not sure if this is actually important, but you could imagine a filter verifying auth against some remote service. That work would be unnecessary if it was a `PUT` request to a `GET`-only route.

seanmonstar:
> I'm kind of lost on the rejection / cancellation stuff, and didn't spend too much time trying to figure it out.

Yea, it's only used in this one set of filters, and isn't well commented or likely well designed. I copied the concept over from Akka, while trying to remove the performance cost (in Akka, it allocates much more freely, because that's just how the JVM does things). I'm not even sure the way it works right now is always correct... I can think of some situations where the canceling of the rejection could not match what the user wanted...

Here's how it works currently: say you have a few routes that allow different methods, but both reject for other reasons.

- `get(path!("todos" / u64))` to fetch the details of a single Todo.
- `post(path!("todos"))` to post a new Todo.

A request like `GET /todos` would pass the first `get` filter, but then reject on the path. It'd then try to the `post` route, and reject because the method is wrong. With cancellation, this will result in a 404, since warp figures that the method *is* allowed, just the path is wrong. Without cancellation, it'd return a 405 because of the `post` filter. Again, perhaps this behavior is wrong in some cases... I can certainly see the argument that it actually **should** be a 405, since that specific path does *not* allow `GET`...

Writing this comment up, I'm starting to lean towards "rejection cancellation" is not needed and/or confusing, and if it were just removed, these wouldn't be *required* to be `Wrap`s.

>  depending on where it's added in the chain, a lot of processing could happen before it gets a chance to reject.

That's true, but it also seems advantageous to allow a user to determine for themselves which predicates should be evaluated first. For example, if there is a long path and header chain that would be nice to "mount" a whole section of the API, it might be more natural for the method filters to be near the end.

```rust
let details = get().map(|| reply_details());
let update = post().and_then(update_details);

let routes = my_mounting.and(details.or(update));
```

seanmonstar:
As an update, I've filed #30, which removes cancelation from warp.

seanmonstar:
I've merged #30, so I believe this could now be updated to do return simpler filters:

```rust
pub fn get2() -> impl Filter<Extract=(), Error=Rejection> { .. }
```

I'd suggest starting with `_2` versions of the current ones, and not exposing any thing else publicly. It's easy to make things public, it's hard to change things once they are.

kamalmarhubi:
@seanmonstar is your preference `get2` vs `v2::get`? I guess so since there's a desire to expose things directly at the top level?

seanmonstar:
Maybe it's better to be in a v2 module in the method module, and only suffixed with a 2 at the top level? Then you could import the v2 module instead if that was your preference.

seanmonstar:
Continued and merged in #41.

Issue #29, opened 2018-08-06
Providing parameters to websocket closure


caolan:
I'd like to use a path parameter from the update request inside the websocket handling code, but the current API doesn't provide an easy way to do that:

```rust
warp::path("example")
    .and(warp::path::param())
    .and(warp::ws(|websocket| {
        // cannot access path parameter easily here
    }))
```

seanmonstar:
Hm, possibly the design of the `ws` filter could be changed, to not use a closure when building, but rather to attach how to handle the upgraded websocket in `map`...

```rust
warp::path::param()
    .and(warp::ws())
    .map(|param: MyParam, ws: Ws| {
        ws.on_upgrade(move |websocket| {
            // moved `param` in here...
        })
    })
```

That'd change the `Ws` to not `impl Reply`, and instead, calling `Ws::on_upgrade` would return the `impl Reply`.

seanmonstar:
Done in #34

PR #34, opened 2018-08-07
add warp::ws2() as a more flexible websocket filter

seanmonstar:
Motivated by #29 

cc @caolan

### DIFF
--- a/examples/websockets.rs
+++ b/examples/websockets.rs
@@ -7,20 +7,22 @@ use warp::{Filter, Future, Stream};
 fn main() {
     pretty_env_logger::init();
 
-    // The `ws()` filter will do the full Websocket handshake,
-    // and call our function if the handshake succeeds.
-    let ws = warp::ws(|websocket| {
-        // Just echo all messages back...
-        let (tx, rx) = websocket.split();
-        rx.forward(tx)
-            .map(|_| ())
-            .map_err(|e| {
-                eprintln!("websocket error: {:?}", e);
+    let routes = warp::path("echo")
+        // The `ws2()` filter will prepare the Websocket handshake.
+        .and(warp::ws2())
+        .map(|ws: warp::ws::Ws2| {
+            // And then our closure will be called when it completes...
+            ws.on_upgrade(|websocket| {
+                // Just echo all messages back...
+                let (tx, rx) = websocket.split();
+                rx.forward(tx)
+                    .map(|_| ())
+                    .map_err(|e| {
+                        eprintln!("websocket error: {:?}", e);
+                    })
             })
-    });
+        });
 
-    let path = warp::path("echo");
-    let routes = path.and(ws);
 
     warp::serve(routes)
         .run(([127, 0, 0, 1], 3030));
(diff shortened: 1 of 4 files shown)


scrogson:
@seanmonstar What is the rationale behind the name `Ws2` and `ws2`?

seanmonstar:
@scrogson I'd expect them to replace the existing `ws()` filter, which returns a `Ws`. Changing them in 0.1 however is a breaking change, so I added a `2` suffix (similar to things like `futures::Stream::concat2()`). Do you think a different name is better?

kellytk:
@seanmonstar FWIW I've seen `next` and `ng` used in identifiers to indicate what you describe.

scrogson:
> I'd expect them to replace the existing ws() filter, which returns a Ws.

Does this mean that `ws2` and `Ws2` would become `ws` and `Ws` in `v1.0`?

> Changing them in 0.1 however is a breaking change, so I added a 2 suffix (similar to things like futures::Stream::concat2()).

In my opinion (and I could be completely wrong or mean to suggest this, but...), anything `< 1.0` is fair game for breakage. Given this is early days, I assume most people are still in the exploration phase and might not be upset about a small breaking change. I understand if you don't agree with me on that. 😉 

> Do you think a different name is better?

I guess my biggest complaint is seeing version numbers in the types rather than a module namespace.

Coming from the Erlang/Elixir world, it's not completely obvious if it's a versioning thing or the arity of the function. In Erlang the arity of a function is often used to document functions: `module:function/arity` due to the fact that functions can share the same name but take different arguments.

seanmonstar:
> Does this mean that ws2 and Ws2 would become ws and Ws in v1.0?

Yes, though rather, in 0.2.0.

In the Rust community (and the way cargo interprets semver), it's expected that upgrading a patch version of 0.x.y to 0.x.z will continue to compile. Cargo will eagerly download the newest patch version.

> Coming from the Erlang/Elixir world, it's not completely obvious if it's a versioning thing or the arity of the function.

I've seen it far more often as "this is the newer version", with several examples in the futures crate. Though, it also happens to have a couple examples of including arity in the name, like `join3`, `join4`, etc... Yet even those are kind of lies, since it's not the actual arity (`fut.join3(one, two)`).

I've also occasionally seen `ng` (for next generation), like there was `std::vec_ng` for a while before Rust 1.0...

caolan:
Thanks @seanmonstar, this API will do what I need.

scrogson:
> I've seen it far more often as "this is the newer version", with several examples in the futures crate.

Good to know, thanks for taking the time to explain 👍

jnicholls:
You could always go straight to 0.2.0 now, rather than getting people to switch to a `ws2()` and then back to `ws()` later. You can go to 0.13483024824.0, it doesn't matter. Make the API the way you want before 1.0.

kellytk:
@jnicholls FWIW @seanmonstar described his rationale in https://github.com/seanmonstar/warp/issues/14#issuecomment-410028563 "I'll probably hold off on a breaking change for at least a little bit, so that these papercuts can be grouped up."

jnicholls:
@kellytk Indeed, fair enough! :) Thanks.

Issue #22, opened 2018-08-04
nested querystrings


hwchen:
Hi!

Thanks for your hard work on this! I'm enjoying the filters api.

I have a use case for nested querystrings (or some way to create a list out of repeated querystring keys). I don't believe that serde_urlencoded supports this, while I think [serde_qs](https://github.com/samscott89/serde_qs) does.

Would you want to support this in warp, or is there a way for me to write a custom filter?

seanmonstar:
I hadn't heard of serde_qs, does it just support more use cases than serde_urlencoded? I'm certainly interested in providing the best support in warp...

caolan:
Another problem is that serde_urlencoded doesn't support unit enums, which would be handy for strict query parameters.

Example:
https://play.rust-lang.org/?gist=627092ecc8fb4be62189f3997dbd97c8&version=stable&mode=debug&edition=2015

It will serialize them but not deserialize.

There's an old pull request for serde_urlencoded to support it, but it's not been approved: https://github.com/nox/serde_urlencoded/pull/30

seanmonstar:
Having taken a look at serde_qs, the README suggests that unless wanting nested query parameters, to prefer serde_urlencoded. Should the better course of action here be to try to get those two features merged into serde_urlencoded?

hwchen:
Sorry it took me a while to get back to this thread.

I can open an issue for serde_urlencoded and see if there's any interest in supporting nested querystings.

hwchen:
It looks as though the maintainer of serde_urlencoded is not interested in adding nested querystrings.

I understand wanting to use serde_urlencoded, which is probably much more battle tested.

What's the most ergonomic way for me to access the query string? Then I could parse it however I like.

seanmonstar:
We could add a `string()` (naming is hard) filter to the query module, returning a `String`. Then it could be parsed however one sees fit. (This is another case where it can't return the `&str`, but that with `async`/`await`, it probably could.)

hwchen:
would `raw_query()` make sense?

I'm happy to try a pull request if you like.

seanmonstar:
Sure, that'd be great!

PR #36, opened 2018-08-08
feature: filter for raw query strings

hwchen:
A new filter `raw_query()` creates a `Filter` that returns the raw query
string as type String.

- new fn `filters::query::raw_query()`
- expose fn publicly as `warp::raw_query()` in lib.rs
- simple test in `tests/query/rs`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,6 +130,8 @@ pub use self::filters::{
     query,
     // query() function
     query::query,
+    // raw_query() function
+    query::raw_query,
     ws,
     // ws() function
     ws::ws,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,8 +130,6 @@ pub use self::filters::{
     query,
     // query() function
     query::query,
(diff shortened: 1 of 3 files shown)


hwchen:
This would resolve https://github.com/seanmonstar/warp/issues/22 by allowing custom query string parsing.

src/lib.rs
```diff
@@ -130,6 +130,8 @@ pub use self::filters::{
     query,
     // query() function
     query::query,
+    // raw_query() function
+    query::raw_query,
```
seanmonstar: This is just a thought, but what do you think if the function were just `raw`, and not re-exported, just left in the `query` module? That'd mean typical usage would then be `warp::query::raw()`...

hwchen: Makes sense to me, this way it won't seem like a top-level filter. I'll make the change.

seanmonstar:
Great, thanks!

hwchen:
No problem! Thanks!

PR #39, opened 2018-08-08
Add an example of a file upload service.

stephanbuys:
This was based on https://github.com/abonander/multipart/blob/master/examples/rocket.rs by @abonander.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,10 @@ urlencoding = "1.0.0"
 pretty_env_logger = "0.2"
 serde_derive = "1.0"
 handlebars = "1.0.0"
+bytes = "0.4.9"
+multipart = "0.15.2"
+regex = "1.0.2"
+
 
 [profile.release]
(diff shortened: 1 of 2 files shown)


seanmonstar:
I opened #40 to discuss a multipart API.

stephanbuys:
@seanmonstar should I keep this PR open, wait for filter support? Or resubmit when multipart support lands in warp proper?

seanmonstar:
I'm not sure, to be honest. It might be useful for people to see this while support is missing? But I'd definitely rather be able to add an example using some sort of `warp::multipart::upload()` (or something) filter, since doing it the way in this example can make it pretty easy for someone to accidentally put an enormous file upload into memory...

abonander:
Yeah that's definitely a concern.

When Warp gets a body-stream filter this example can be implemented properly, though it will need the `multipart-async` crate instead.

stephanbuys:
Closing for now...

PR #41, opened 2018-08-09
Add method::v2 filters that act like all the others

seanmonstar:
refs #14

----

@kamalmarhubi I took what you started in #21 and applied the comments that I had left there, this is what it looks like. What do you think?

### DIFF
--- a/examples/headers.rs
+++ b/examples/headers.rs
@@ -21,15 +21,10 @@ fn main() {
     // Match when we get `accept: */*` exactly.
     let accept_stars = warp::header::exact("accept", "*/*");
 
-    let index = host.and(accept_stars)
+    let routes = host.and(accept_stars)
         .map(|addr| {
             format!("accepting stars on {}", addr)
         });
-
-    // Only allow GETs in this example
-    let routes = warp::get(
-        index
-    );
 
     warp::serve(routes)
         .run(([127, 0, 0, 1], 3030));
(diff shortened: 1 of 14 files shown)


examples/headers.rs
```diff
@@ -21,16 +21,11 @@ fn main() {
     // Match when we get `accept: */*` exactly.
     let accept_stars = warp::header::exact("accept", "*/*");
 
-    let index = host.and(accept_stars)
+    let routes = host.and(accept_stars)
```
kamalmarhubi: did this example lose its get check intentionally?

src/filters/method.rs
```diff
@@ -137,3 +114,69 @@ where
...
+    //!
+    //! These filters deal with the HTTP Method part of a request. They match
+    //! the request `Method`, and if not matched, will reject the request with a
+    //! `405 Method Not Allowed`.
+    use http::Method;
+
+    use filter::Filter;
+    use reject::Rejection;
+
+    use super::method_is;
```
kamalmarhubi: while here, could you add a short comment on `method_is` above explaining why it takes an Fn? copying https://github.com/seanmonstar/warp/pull/21#discussion_r207746190 would be enough.

examples/headers.rs
seanmonstar: Yea, the deprecation lint trigger, and as I started to adjust it, I figured showing method filters in the **headers** example was just noise.

PR #43, opened 2018-08-09
Add basic information to README.md

cdvv7788:
The information at the repository is too short. While there are more in-depth docs, the README should communicate the purpose of the package.
Currently, it says something like: We have a web framework and we use Filters to do some things.
If it says what are the things the framework does, it could be more easy for newcomers to understand its capabilities.

Also added an explicit link to docs and examples. There are just too many repositories with broken badges, so in my case, I tend to ignore them and look for links in the readme.


seanmonstar:
Thanks!

PR #47, opened 2018-08-10
Clarify project scope

kellytk:


seanmonstar:
Thanks!

README.md
```diff
@@ -5,7 +5,7 @@
 [![crates.io](https://meritbadge.herokuapp.com/warp)](https://crates.io/crates/warp)
 [![Released API docs](https://docs.rs/warp/badge.svg)](https://docs.rs/warp)
 
-A super-easy, composable, web framework for warp speeds.
+A super-easy, composable, web backend framework for warp speeds.
```
scrogson: `s/backend/server/` ?

kellytk: A frontend could implement a server and a backend could implement a client, albeit from a standpoint other than an end-user, so I personally think that server in this context is still somewhat ambiguous.  IMHO server/client is more a description of the receiver/initiator of connections whereas backend/frontend is more a description of the dual running contexts of a 'web app.'  What do you think @scrogson?

scrogson: Sure..but this is a framework for building web servers, not clients...unless I'm missing something?

kellytk: A web server could be implemented in the frontend of a p2p 'web app.'

PR #51, opened 2018-08-11
boxed: Use Into in place of From in error bound

kamalmarhubi:
This will make it possible for users to use `Filter::boxed` with their
custom filters and error types once the traits are unsealed.

### DIFF
--- a/src/filter/boxed.rs
+++ b/src/filter/boxed.rs
@@ -26,11 +26,11 @@ impl<T: Tuple + Send> BoxedFilter<T> {
         F: Filter<
             Extract=T,
         > + Send + Sync + 'static,
-        Rejection: From<F::Error>,
+        F::Error : Into<Rejection>,
     {
         BoxedFilter {
             filter: Arc::new(BoxingFilter {
-                filter: filter.map_err(Rejection::from),
+                filter: filter.map_err(Into::into),
             }),
         }
     }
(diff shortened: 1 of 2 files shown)


seanmonstar:
Curious then, what is gained by the change?

kiljacken:
As far as I understand, the gain is the ability for a user to use their own error type that `impl Into<Rejection> for MyErrorType` instead of having to use a rejection outright. I use a similar design on my current project, but I currently have to resort to use a macro to generate a boilerplate wrapper functions for each endpoint.

Ideally the end goal would be allowing passing a function to `filter.and_then()` that returns `Result<impl warp::Reply, Something>` where `Into<Rejection>` is implemented for `Something`. This would allow easier integration with using the default Rust error trait, and even the `Failure` trait from the failure package, as long as the user `impl`s the corresponding Into for their error type.

seanmonstar:
Yea that seems fine, but the current released code already allows that. You can `impl From<MyFoo> for warp::Rejection` instead of `impl Into<warp::Rejection> for My Foo`. It's generally better to implement `From` than `Into`, `Into` is meant to be used via its blanket implementation over all `From`.

kiljacken:
Ahh, I've seemingly had a misunderstanding of what's allowed when implementing traits for foreign types. With that in mind, yes, this PR does seem somewhat pointless.

Issue #11, opened 2018-08-02
Expose more HTTP methods as convenience filters, eg PATCH, HEAD


kamalmarhubi:
I'm not sure what the right line is for which ones merit convenience filters, but at least `PATCH` I'd think. Alternatively, given the small number, all could be exposed.

List of methods:
- :heavy_check_mark: `GET`
- :grey_question: `HEAD` this could be promoted, I think
- :heavy_check_mark: `POST`
- :heavy_check_mark: `PUT`
- :heavy_check_mark: `DELETE`
- :grey_question: `CONNECT`(pretty niche for use cases of warp?)
- `OPTIONS` unsure of this one
- :x: `TRACE` 
- :grey_question: `PATCH` I think this one should be promoted

(I'm ignoring webdav and any other such protocols on top of HTTP here.)

kamalmarhubi:
Ah, looks like these are exactly the methods that `http` considers first class: the `PATCH` extension is included as a variant, but all others are grabbed by `Extension` ([source](https://github.com/hyperium/http/blob/c9358881db8b2054bee95905657be4b3b3ffc64c/src/method.rs#L115)). This makes me think that including these makes sense.

seanmonstar:
Yea, could probably promote a few. I wanted the most common ones to be easy to use, and the current 4 are definitely the the 95ish percent.

PR #52, opened 2018-08-11
method: Add filters for HEAD, OPTIONS, PATCH

kamalmarhubi:
Add `method::v2`-style filters for these methods, exposing them at the
top level without a `2` suffix. Among HTTP methods defined in RFC 7231
methods, this leaves `TRACE` and `CONNECT` without filters.

refs #11


seanmonstar:
Woot!

PR #55, opened 2018-08-13
add `hyper::Service` and `hyper::NewService` trait implementations for `WarpService`

dmexe:
As I see, a creation of `hyper::NewService` hardcoded inside `Server::bind_ephemeral` method,
so I've created a separate module `service`, that provides `hyper::NewService` trait implementation for wap services. It would be useful if warp uses together with customized tokio runtime or hyper stacks.

### DIFF
--- a/src/filter/service.rs
+++ b/src/filter/service.rs
@@ -4,7 +4,7 @@ use ::{Filter, Request};
 use ::reject::Reject;
 use ::reply::{Reply};
 use ::route::{self, Route};
-use ::server::{IntoWarpService, WarpService};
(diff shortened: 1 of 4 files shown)


seanmonstar:
Thanks for working on this! I do want to allow turning `Filter`s into a `Service`, I just hadn't decided on if that should be `hyper::service::Service`, or the newly released [`tower::Service`](https://docs.rs/tower-service/0.1.0/tower_service/trait.Service.html).

Do you have situation you were trying to use warp filters directly with a `hyper::Server`?

PR #57, opened 2018-08-14
Expand documentation on returning filters from functions

blm768:
This PR adds an example that shows the two main ways to return a filter from a function. It also adds a short example to the `BoxedFilter` docs in case someone stumbles across that before they find `Filter::boxed()`. I don't know whether the redundancy is worth it, but I added the example before I realized that the existing examples were there (since they're not in the docs.rs version yet), so they're in there, but I can also remove them.


seanmonstar:
Cool, thanks!

Issue #37, opened 2018-08-08
Error message with rejection


kiljacken:
It would be nice if the rejections allowed passing an error message, so the end user of an API has a chance of knowing what they did wrong (for example on bad request).

hjr3:
I have been playing around with warp and have run into this as well. The `warp::Reply` trait allows [a number](https://github.com/seanmonstar/warp/blob/45f8535662afde031c04d0a96c8d490c72212495/src/reply.rs#L11) of different ways to create a reply. This allows for simple use cases (using only a string or number) and complex use cases that want to build up complex responses using `http::Response`.

The `warp::Rejection` type is a dead end in that it only accepts a reason code. There are use cases where I want to return a body along with the reason code and even other use cases where I want to set a `Content-Type` header as well.

I think `warp::Rejection` should to allow for an `http::Response` or `ReplySealed` should to be implemented for `Result<impl warp::Reply, impl warp::Reply>`.

kamalmarhubi:
I believe the intent with `Rejection` is to be a very small error type that covers the most common errors. In the hypothetical 1.0 release, I think that the [`Reject` trait](https://github.com/seanmonstar/warp/blob/283a5463d7548a24afc69bf2f2521148aacb7061/src/reject.rs#L174-L177) would be unsealed, and you'd have your custom error implement that.

dmexe:
Proposal: Error reform
======================

Current State
-------------

`Reject` holds only a status `Reason: u8` which can be composed with another rejection to determine final `Filter` result. (as I understood after digging a code)

So `Reject` usable for flow control during `Filter`execution and lacks an error description which critical for debugging and ops. In many cases, an error cause just lost (for example https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs#L187)

Proposal
--------

It would be  better using same error handling mechanism as finch, (see https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/EndpointResult.scala#L111-L128). Which means:

1) Using separate enum variant which indicates that current filter's branch isn't matched.
2) Fail fast, no error composition, trying next branch only if first returns `NotMatched`
3) An error keeps inside another enum variant

`Reject` and `Rejection` may be changed to:

```rust

pub enum Rejection {
  // signals that current `Filter`'s branch is not valid and cannot be matched.
  NotMatched, 
  // holds an internal error which maybe happens during filter execution or user handler's result
  Err {
    status: StatusCode, 
[...]

Issue #33, opened 2018-08-07
Custom Failure / error handling


kamek-pf:
Hey,
Really like what you did !

Say I want to define very rich, specific errors, I usually do this with `failure`.
I'd like to be able to return my own custom failures in the `and_then` combinator, and then either simply log them or present a detailed error page.

Do you currently have a way to do this ? Is it a goal at all ?
Maybe I just have the wrong approach, what would you recommend to deal with this ?

seanmonstar:
As of 0.1.0, warp only allows errors back into the filter chain that are `warp::Rejection`s. You can create your own futures, and your own errors, definitely, you'll just need to map your custom errors into an appropriate `Rejection`. Take look at the [errors example](https://github.com/seanmonstar/warp/blob/master/examples/errors.rs).

I do think it'd probably be nice to expand the rejection system to allow for better typed application errors, but I don't yet know the best way to do so.

PR #60, opened 2018-08-16
Add cause to Rejection (WIP)

dmexe:
Discussions: #37, #33 

It doesn't break anything, I've just added an error cause to a Rejection

- [x] Add an error cause to Rejection
- [x] Serialize Rejection into JSON
- [x] Combine rejections with causes
- [x] Use an rejection's error inside filters

### DIFF
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -25,8 +25,9 @@ pub(crate) fn body() -> impl Filter<Extract=(Body,), Error=Rejection> + Copy {
             .take_body()
             .map(Ok)
             .unwrap_or_else(|| {
-                warn!("request body already taken in previous filter");
-                Err(reject::server_error())
+                let err = "request body already taken in previous filter";
+                warn!("{}", err);
+                Err(reject::server_error().with(err))
             })
     })
 }
@@ -55,8 +56,9 @@ pub fn content_length_limit(limit: u64) -> impl Filter<Extract=(), Error=Rejecti
             if length <= limit {
                 Ok(())
             } else {
-                debug!("content-length: {} is over limit {}", length, limit);
-                Err(reject::payload_too_large())
+                let err = format!("content-length: {} is over limit {}", length, limit);
+                debug!("{}", err);
+                Err(reject::payload_too_large().with(err))
             }
         })
         .unit()
@@ -117,7 +119,7 @@ pub fn json<T: DeserializeOwned + Send>() -> impl Filter<Extract=(T,), Error=Rej
             serde_json::from_slice(&buf.chunk)
                 .map_err(|err| {
                     debug!("request json body error: {}", err);
-                    reject::bad_request()
+                    reject::bad_request().with(err)
                 })
         })
 }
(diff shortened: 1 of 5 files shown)


src/filters/path.rs
```diff
@@ -186,15 +186,19 @@ pub fn index() -> impl Filter<Extract=(), Error=Rejection> + Copy {
 ///         format!("You asked for /{}", id)
 ///     });
 /// ```
-pub fn param<T: FromStr + Send>() -> impl Filter<Extract=One<T>, Error=Rejection> + Copy {
+pub fn param<T>() -> impl Filter<Extract=One<T>, Error=Rejection> + Copy
+where
+    T: FromStr + Send,
+    T::Err: Into<::reject::Cause>
```
seanmonstar: This would be a breaking change, though it definitely seems worth doing as part of 0.2!

src/reject.rs
```diff
@@ -107,14 +116,47 @@ impl Rejection {
...
+    {
+        let cause = Some(err.into());
+        Self {
+            cause,
+            .. self
+        }
+    }
+}
+
+impl PartialOrd for Rejection {
```
seanmonstar: We might not want ordering to be part of the public API of a `Rejection`... Mostly because I don't know if it's obvious to a user *why* it has one. What would they assume the ordering means? The ordering used on the `reason` field can just be an implementation deal in `CombineRejection`.

src/reject.rs
```diff
@@ -157,23 +203,62 @@ impl Reject for Rejection {
...
     fn into_response(self) -> ::reply::Response {
-        let code = self.status();
+        use http::header::{CONTENT_TYPE, HeaderValue};
+        use hyper::Body;
 
+        let code = self.status();
         let mut res = http::Response::default();
         *res.status_mut() = code;
+
+        match serde_json::to_vec(&self) {
```
seanmonstar: This may make sense in JSON APIs, but maybe less so in other cases. In other cases, it may be better to just print a plain text message.

The best case would probably be if we could convert this somehow to just `Response<Cause>`, and a solution to #4 would allow a user to automatically add converting rejections into JSON (or some other format)...

src/reject.rs
```diff
@@ -220,3 +311,75 @@ mod sealed {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn combine_rejections() {
```
seanmonstar: Thanks so much for adding tests about this! It was sorely missing.

src/filters/body.rs
```diff
@@ -25,8 +25,9 @@ pub(crate) fn body() -> impl Filter<Extract=(Body,), Error=Rejection> + Copy {
             .take_body()
             .map(Ok)
             .unwrap_or_else(|| {
-                warn!("request body already taken in previous filter");
-                Err(reject::server_error())
+                let err = "request body already taken in previous filter";
+                warn!("{}", err);
+                Err(reject::server_error().with(err))
```
seanmonstar: This is admittedly pretty cool!

dmexe:
Updates summary:
1) created a new `warp::path::param2`, it would be become `warp::param::param` in next major release.
2) removed `PartialCmp` trait implementations, comparison now inside `CombineRejection`
3) replaced an error cause default format from json to text.

seanmonstar:
Thanks, this is pretty great!

PR #68, opened 2018-08-22
Add rustfmt

dmexe:
It would be better to format source files according to the default 'rustfmt' code style.

**Note**: `rust fmt` has a bug 

The block in `lib.rs`
```rust
pub use ::self::filters::{
  // any() function
  any::any
}
```

becoming to 
```rust
pub use ::self::filters::{
  // any() function any::any
}
```

so, I've removed comments from the block.


thomaseizinger:
It would be good to report this bug to the rust-lang team, since `rustfmt` is entering RC mode and is about to be stabilized for the Rust 2018 edition. It would be pretty bad if a bug like that makes it into 1.0!

remexre:
Is this stance "permanent" now that rustfmt is stable?

seanmonstar:
@remexre not permanent, if the above bug has been fixed. If it has, I think it'd be fine to start using rustfmt on warp...

PR #69, opened 2018-08-22
Update serde_urlencoded to 0.5.3

kiljacken:
Fixes a long-standing issue with decoding unit enums from query strings


seanmonstar:
Thanks! For anyone needed this fix immediately, `cargo update` will pull it in, since it's a semver patch change.

Issue #56, opened 2018-08-13
Filter and Service conversions


seanmonstar:
It's a goal to allow turning a `warp::Filter` into a `tower::Service`, and to allow for `warp::Server` to be given a `Service` (or `NewService`) as its argument. This would facilitate making use of the wider tower middleware directly.

seanmonstar:
Closed via #367

PR #70, opened 2018-08-23
Add tower_service::Service implementation for Filter

dmexe:
I've created the `LiftService` that implements both tower and hyper services. A filter could be turned into service by calling the function `filter.lift()`

- [x] create the `LiftService`
- [x] add tower and hyper service implementations for the `LiftService`
- [x] `Server` can consume  `LiftService`. 
- [x] replace `filter::FilteredService` by `LiftService`
- [x] remove `WarpService` and `IntoWarpService` traits
- [x] add an example which uses `LiftService` along with `hyper::Server`

Topic: #56


carllerche:
:heart:

I might recommend to name the fn `into_service()` and the wrapper type `FilterService`. The `lift` terminology tends to be used for a specific pattern:

```rust
trait MainTrait {}

trait SpecializedTrait {
    fn lift(self) -> Lift<Self> { ... }
}

impl<T: MainTrait> SpecializedTrait for T {}
```

It is mostly a hack around the lack of specialization.

toidiu:
i was wondering what the status of this was since its been some time?

Issue #2, opened 2018-08-01
Add CORS filter


seanmonstar:

digeratus:
👍🏾+1

manifest:
Is there any concept already how it will look in code?

manifest:
To support CORS we need to add specific headers only if `Origin` header is provided.

For preflight requests it's easy to implement using the filter:

```rust
warp::options().and(warp::header("origin")).map(|origin| {
    Ok(Response::builder()
        .status(StatusCode::OK)
        .header("access-control-allow-methods", "HEAD, GET")
        .header("access-control-allow-headers", "authorization")
        .header("access-control-allow-credentials", "true")
        .header("access-control-max-age", "300")
        .header("access-control-allow-origin", origin)
        .header("vary", "origin")
        .body(""))
});
```

What would be the right way to implement handlers for regular requests? Some kind of response extension to check on `Origin` header?

```rust
// If "Origin" header is presented
warp::get2().map(|| {
    Ok(Response::builder()
        .status(StatusCode::OK)
        .header("access-control-allow-origin", origin)
        .header("vary", "origin")
        .body("foobar"))
})

// If it isn't
warp::get2().map(|| {
    Ok(Response::builder()
        .status(StatusCode::OK)
        .body("foobar"))
})
```

manifest:
Probably a better idea would be to implement a filter that initialize response with

```rust
Response::builder()
    .header("access-control-allow-origin", origin)
    .header("vary", "origin")
```

if "origin" header is presented and just `Response::builder()` if isn't.

```rust
warp::get2()
    .and(warp::cors())
    .map(|response| {
        Ok(response
            .status(StatusCode::OK)
            .body("foobar"))
    })
```

Is there any documentation on how filters parsing data from requests can be built?

seanmonstar:
> Is there any concept already how it will look in code?

What I've had in mind is to make use the sealed `Wrap` trait, allowing usage with `Filter::with`:

```rust
let cors = warp::cors()
    .allow_origin("my.foo.local")
    .allow_max_age(60 * 10)
    .allow_methods(&["GET", "POST", "DELETE"]);

// chained on any route:
let route = warp::path("api")
    .map(generate_response)
    .with(&cors);
```

> borrowed value does not live long enough

[...]

PR #73, opened 2018-08-26
WIP: Add cors filter

manifest:
PR for the [issue](https://github.com/seanmonstar/warp/issues/2). Work in progress.

### DIFF
--- a/src/filters/cors.rs
+++ b/src/filters/cors.rs
@@ -1,0 +1,124 @@
+//! CORS Filters
+
+use ::filter::{Filter, WrapSealed};
+use ::reject::Reject;
+use ::reply::{Reply, Response};
+use ::route;
+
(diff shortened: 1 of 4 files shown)


shepmaster:
It'd be wonderful to cross-reference with [the implementation in tower-web](https://github.com/carllerche/tower-web/tree/master/src/middleware/cors). Since the plan is to merge the projects, having "close enough" code is probably a desirable goal.

manifest:
@shepmaster Thanks for sharing. It isn't clear for me at the point how warp and tower-web should relate to each other.

carllerche:
One thing to figure out: What should be a filter vs. what should be a middleware?

manifest:
I'm closing this for now until there would be a better understanding of warp relation to tower-web.

PR #76, opened 2018-08-28
Allow content-type to specify charset in body filters

### LINKED ISSUE #74: body::form rejects if content-type includes charset
I'm seeing seemingly-good requests be rejected. Looks like this might happen when the client specifies a charset in the content-type. Relevant log output:
```
 TRACE warp::filters::body        > is_content_type "application/x-www-form-urlencoded"? "application/x-www-form-urlencoded; charset=utf-8"
 DEBUG warp::filters::body        > content-type doesn't match "application/x-www-form-urlencoded"
```

I'm not an http expert and am unaware whether it's technically correct for charset to be included here, but random googling indicates it's not rare.

Maybe [`is_content_type`](https://github.com/seanmonstar/warp/blob/v0.1.2/src/filters/body.rs#L97) should just check what the header starts with? or split on semicolons?

SSheldon:
Previously, the form and json body filters would reject requests where the content-type included a charset. This change uses mime to parse the content-type and perform a smarter comparison.

This fixes #74.


seanmonstar:
Great, thank you!

PR #78, opened 2018-08-28
Add warp::reject::forbidden

pietroalbini:


src/reject.rs
```diff
@@ -105,7 +111,8 @@ bitflags! {
         const LENGTH_REQUIRED        = 0b00000100;
         const PAYLOAD_TOO_LARGE      = 0b00001000;
         const UNSUPPORTED_MEDIA_TYPE = 0b00010000;
-        const SERVER_ERROR           = 0b10000000;
+        const SERVER_ERROR           = 0b00100000;
```
seanmonstar: The `SERVER_ERROR` being the biggest flag is relied upon for combining rejections together (so any route that triggers a server error won't have it lost).

pietroalbini: Oh, ok. Reordered and added a comment explaining this.

PR #87, opened 2018-09-02
Treat no query string as empty query string

### LINKED ISSUE #86: Accept missing query string for optional argument
Hi!  I have a handler that should accept an optional query parameter, so both `http://myhost/path/?arg=foo` and `http://myhost/path/` should be handled by the same method.

For that I use a route that looks like this (get() is really use warp::get2 as get, to be prepared for coming versions).

```rust
        .or(get().and(path("path")).and(index()).and(query()).map(myhandler))
```

```rust
fn myhandler(param: ArgQ) -> Response<Vec<u8>> { ... }

#[derive(Debug, Deserialize)]
struct ArgQ {
    arg: Option<String>,
}
```

This works fine as long as there is a query string in the request, i.e. `/path?arg=foo` gives me an ArgQ where arg is `Some("foo")` and `/path?somethingelse=whatever` or even just `/path?` gives med arg = `None`.  But `/path` (without a question mark) gives me a bad request without even entering my handler.  Is this a bug?  Or should I somehow mark `query()` to be optional?

kaj:
If there is optional query parameters to a handler, treat an URL with no question mark as equal to an URL that ends in a question mark.  Technically, the URL ending in a question mark has an empty query string while the URL without a question mark has no query string, but in practice, it should not make any difference and the question mark is rarely present.

If there are required query parameters, I don't think this makes any difference either, as extracting them from an empty string will fail fast (and the fail should behave the same with or without that question mark anyway).

This fixes #86 in the place where it is easy to fix.  Also added a bunch of tests for missing query strings (and query string missing required pieces).

### DIFF
--- a/src/filters/query.rs
+++ b/src/filters/query.rs
@@ -13,6 +13,7 @@ pub fn query<T: DeserializeOwned + Send>() -> impl Filter<Extract=One<T>, Error=
     filter_fn_one(|route| {
         route
             .query()
+            .or(Some(""))
             .and_then(|q| {
                 serde_urlencoded::from_str(q)
                     .ok()
--- a/tests/query.rs
+++ b/tests/query.rs
@@ -1,5 +1,7 @@
 #![deny(warnings)]
 extern crate warp;
+#[macro_use]
+extern crate serde_derive;
 
 use std::collections::HashMap;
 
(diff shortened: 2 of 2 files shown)


seanmonstar:
Seems like a great improvement to me, thanks!

PR #92, opened 2018-09-15
Add a flatten() function to the Filter trait.

Emm:
This adds a lot of flexibility. I use it like this:

```rust
fn with_host(host: Arc<Host>) -> BoxedFilter<(impl Reply,)> {
    let h = host.clone();
    path!("host").map(move || format!("{}", h.0)).boxed()
}

fn with_host_or_static(host: Arc<Host>) -> BoxedFilter<(impl Reply,)> {
    with_host(host).or(static_files()).boxed()
}

fn app(domain: String) -> BoxedFilter<(impl Reply,)> {
    host(domain)
        .map(|host| with_host_or_static(host))
        .flatten()
        .boxed()
}
```


Emm:
Is that really different from a faulty `FromStr` implementation which panics? That's also going to crash the Tokio thread, isn't it?

Emm:
In my case, this was a dependency injection issue. Essentially, I have what
you would call an app (a set of routes with an `or`) where some of them
require a structure built from parameters captured earlier in the query.
So, the app itself is a filter, but it needs the result of a previous
filter to operate. I have not found another way of doing this using
existing filters, and I imagine this is not an isolated case.

euclio:
I believe I have a use case for this as well. I'm trying to write a server that uses `warp::fs::dir` to serve static files, but I want to be able to change the directory that the `dir` filter uses at runtime. I tried writing something like

```rust
fn serve_relative_to(path: Arc<Mutex<PathBuf>>) -> impl Filter<Extract=(File,), Error = Rejection> {
    let dir = path.lock().unwrap();
    warp::fs::dir(*dir)
}

fn main() {
    let dir = Arc::new(Mutex::new(PathBuf::from("/some/path")));

    let routes = warp::any()
        .map(move || Arc::clone(&dir))
        .and_then(serve_relative_to);
}
```

but I can't get it to compile successfully.

udoprog:
Just to chime in, I would have appreciated this feature today.

I have two case: I either serve files which are statically compiled into the binary, or I allow a developer to specify a web-root from where to serve those files (and those files are reloaded on changes).

I worked around conditional file serving at the end of the setup, by copying a bunch of code for now:
https://github.com/udoprog/setmod/blob/master/bot/src/web.rs#L315

L-as:
I get why this is problematic for warp, but even if this was not to be merged, I do not think that panicking when arguments to the filter creation functions are bad is good behavior. There are other cases besides this where I think that it would be problematic, as it's *still* possible to create filters after start-up, just not like this, in addition to it not being obvious that the functions are fallible.

Emm:
Going to close this, that's way too old.

sariug:
but is there a workaround currently for this? I am also having the same trouble:
- my rust code is generating files and then sends them therefore the name of the file is always changing.

PR #93, opened 2018-09-16
Add optional cookie value filter

algermissen:
An additional cookie filter that I needed to implement inside the warp crate because the necessary building blocks, such as header::optional_value are only crate-public.

Rather than accepting this PR, I guess it will be more helpful to understand what you mid-term goal is to support writing own headers (which I think will be a very common thing to do).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.idea/
+warp.iml
--- a/src/filters/cookie.rs
+++ b/src/filters/cookie.rs
@@ -26,6 +26,21 @@ pub fn optional(name: &'static str) -> impl Filter<Extract=One<Option<String>>, 
     })
 }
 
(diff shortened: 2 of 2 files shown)


seanmonstar:
> I guess it will be more helpful to understand what you mid-term goal is to support writing own headers (which I think will be a very common thing to do).

I've been spending a bit of time recently on re-designing the typed headers that were in hyper pre-0.12, and I expect to eventually allow warp to use those directly (replacing the `header::<T: FromStr>(name)` API for one of `header<T: Header>()`, and then likely also adding `header::optional<T: Header>()` as well.

seanmonstar:
> Rather than accepting this PR

So would you rather I don't merge this?

algermissen:
If it fits for you, please go ahead. I just wanted to emphasise the general question on exposure of the necessary elements (which `header::optional<T: Header>()` would do)

PR #94, opened 2018-09-17
Improve error handling.

dmexe:
Solve https://github.com/seanmonstar/warp/issues/37#issuecomment-422039319

* New method `Rejection::into_cause` which use to match an error cause.
* New method `Rejection::json()` which return a json response.
* Update a `errors.rs` example to show how to handle custom errors.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -127,7 +127,7 @@ impl Rejection {
         }
     }
 
-    /// Returns a json responder for this rejection.
+    /// Returns a json response for this rejection.
     pub fn json(&self) -> ::reply::Response {
         use http::header::{CONTENT_TYPE, HeaderValue};
         use hyper::Body;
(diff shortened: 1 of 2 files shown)


src/reject.rs
```diff
@@ -125,6 +126,51 @@ impl Rejection {
...
+    /// Returns a json responder for this rejection.
+    pub fn json(&self) -> ::reply::Response {
+        use http::header::{CONTENT_TYPE, HeaderValue};
+        use hyper::Body;
+
+        let code = self.status();
+        let mut res = http::Response::default();
+        *res.status_mut() = code;
+
+        let empty = "{}".to_string();
```
seanmonstar: This will always allocate on the heap. I'd write it like this:

```rust
*res.body_mut() = match serde_json::to_string(&self) {
    Ok(s) => Body::from(s),
    Err(_) => Body::from("{}"),
};
```

dmexe: @seanmonstar 
> This will always allocate on the heap. I'd write it like this:

fixed

seanmonstar:
Thanks!

Issue #58, opened 2018-08-14
Relation to tower-web


alexreg:
I just read your blog post that introduces the warp library, so I'm quite intrigued now as to how things will develop now with tower and tower-web quickly entering the stage. I am certainly attracted by the design philosophy of warp, but I do want to know how you see it in relation to tower-web, which you mentioned in passing. From an issue you have open here, it would seem the `Filter` and `Service` traits are somehow analogous? I'm presuming neither tower-web nor warp will ever include the other as a dependency, but what will the relation between them be, going forwards? Direct competitors?

seanmonstar:
We've been talking about how to reduce duplicate efforts, and find pieces that can be combined. It's possible the projects (tower-web and warp) merge at some point, exposing both APIs as "flavors". If you know the Scala world at all, here is how I see the projects:

- tower is like finagle
- warp is like finch
- tower-web is like finatra

So, similar to how in finch, an `Endpoint` can be converted into a `Service`, I think a `warp::Filter` can be converted into a `tower::Service` (see #56).

> does warp support async workflows like Gotham or such?

Yep, the README and the docs both point out that warp is built on top of hyper and thus supports asynchronous request handling.

alexreg:
Thanks for the clarification. Unfortunately I don't know Scala, so those analogies don't help much... I'm still left unsure of whether one framework is better for certain use cases.

> Yep, the README and the docs both point out that warp is built on top of hyper and thus supports asynchronous request handling.

I just looked at the docs closer, and I think I see how you do it now... I wanted to check, because some web frameworks use hyper internally, but don't expose any async stuff really (even if they use it internally?). But I'm glad!

seanmonstar:
> one framework is better for certain use cases.

The goal is that functionality is available in both, and the choice comes down to personal taste of how you prefer structure web apps.

dmexe:
> We've been talking about how to reduce duplicate efforts, and find pieces that can be combined. It's possible the projects (tower-web and warp) merge at some point, exposing both APIs as "flavors". If you know the Scala world at all, here is how I see the projects:

The current state of these projects is going to the wrong direction (as I think). 
1) https://github.com/carllerche/tower-web/tree/master/src/middleware
2) https://github.com/seanmonstar/warp/pull/73

[...]

PR #99, opened 2018-09-23
Add gzip example

GuillaumeGomez:
As promised, here's the gzip example.

I'll try to write up what gave me issues writing this sample of code but clearly it rotates around Filter vs Futures... A bit more examples in the documentation would be **really** useful as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ urlencoding = "1.0.0"
 pretty_env_logger = "0.2"
 serde_derive = "1.0"
 handlebars = "1.0.0"
+flate2 = "1.0.2"
 
 [profile.release]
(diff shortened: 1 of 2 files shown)


seanmonstar:
Cool, thanks! This could definitely work, and for smaller bodies, shouldn't cause much of an issue. For large bodies, and especially if trying to compress files from disk, this could cause the futures thread to block.

With the eventual integration/merge with tower-web (#58), it should become something as simple as this: 

https://github.com/carllerche/tower-web/blob/f12bd745b67aca4bbfb648fe61b31ab6ae7a10ac/examples/middleware.rs#L56

GuillaumeGomez:
That'd be awesome. :O

Issue #46, opened 2018-08-09
Add compression Filters


seanmonstar:
These would likely need to be `with` filters.

Some hypothetical example:

```rust
let route = warp::fs::dir("static")
    .with(warp::compress::gzip());
```

seanmonstar:
As an update, quoting from #103:

> I'm hopeful with the changes to support tower (#58), we can just make use of the existing [deflate middleware](https://github.com/carllerche/tower-web/tree/master/src/middleware/deflate).

jxs:
related to #58 
cc @luciofranco

ParkMyCar:
If we can't integrate with tower's deflate middleware right now, I'd take a stab at implementing this, seems pretty interesting to me!

asonix:
@ParkMyCar if you're going to look into this, here's a PR that may be of interest to you: https://github.com/seanmonstar/warp/pull/397

I've been thinking about @seanmonstar 's comments on there and tbh I'm not sure a whole lot can be done re: the second point, but I feel the first point is easily solved

jxs:
@ParkMyCar if you are looking for a deflate crate with AsyncRead/Async Write there's https://docs.rs/async-compression/0.2.0/async_compression/

PR #103, opened 2018-09-29
Introduce a gzip filter

Kerollmops:
This is a work in progress related to the issue https://github.com/seanmonstar/warp/issues/46.

I let you some comments in the code to help me, if you want, of course 😃

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ keywords = ["warp", "server", "http", "hyper"]
 base64 = "0.9.2"
 bitflags = "1"
 bytes = "0.4.8"
+flate2 = "1.0"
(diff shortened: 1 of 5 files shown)


seanmonstar:
Very cool, thanks for getting started on this!

As your comments mention, it'd be best if we could do this asynchronously. I'm hopeful with the changes to support tower (#58), we can just make use of the existing [deflate middleware](https://github.com/carllerche/tower-web/tree/master/src/middleware/deflate).

Kerollmops:
Ok so I will wait the support of tower in warp, can we close this PR ?

seanmonstar:
Alright, thanks for looking into this though!

PR #105, opened 2018-10-02
Add last-modified header to file and dir filters (#45)

hoggetaylor:
This pull request sets the last-modified header when serving responses from the file system.

This implementation uses [httpdate](https://github.com/pyfisch/httpdate) to format the date.
It appears that Hyper uses the deprecated [time](https://github.com/rust-lang-deprecated/time) crate; which is also an option.


seanmonstar:
Awesome, thanks!

I'm working on [a typed headers redesign](https://github.com/hyperium/headers/pull/4) that I've been testing out in warp to add this kind of stuff, but since it's not ready yet, and this PR is, let's go!

PR #106, opened 2018-10-02
Made file response public

madmaxio:


seanmonstar:
Thanks for the PR! I'd like to keep the fields of the struct private for now, but to allow the use case you have, I've added `warp::reply::with_header(reply, name, value)` to master.

PR #107, opened 2018-10-02
add warp::redirect(as_location) reply helper

### LINKED ISSUE #100: How to return HTTP 301
The docs and examples are not clear on how to do redirect.

seanmonstar:
Closes #100

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -108,6 +108,7 @@ mod filter;
 pub mod filters;
 mod generic;
 mod never;
+pub mod redirect;
 pub mod reject;
 pub mod reply;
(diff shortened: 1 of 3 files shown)


stevensonmt:
I love the ease of this redirect method, but it does feel awkward that the argument is not a warp filter type. For instance, if I've defined `home = warp::path("/home")` and want to do a redirect from `/login/callback` to `/home` it would be intuitive to put `callback = login.and(warp::path("callback").map(...process callback payload...).redirect(home)`. Bonus points if I could do `...redirect(home.and(warp::path("?callback_payload_params")))`.

PR #110, opened 2018-10-03
 Add automatic insert for directory routes 

FreeMasen:
Addresses issue #108


seanmonstar:
Thanks! That was quick!

PR #111, opened 2018-10-04
Fix typo in `with_status` documentation

jeffparsons:
Previously referred to headers; should refer to status codes.


seanmonstar:
Woops, thanks!

PR #114, opened 2018-10-09
Reject with timeout

xoac:
add `request_timeouted` to `reject` module.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -64,6 +64,12 @@ pub fn not_found() -> Rejection {
     Reason::empty().into()
 }
 
+/// Rejects a request with `408 Request Timeout`
+#[inline]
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks for the PR! I'm currently working through a refactor of the rejection system, to allow arbitrary status codes and whatever to be set, so that there doesn't need to be special support like in this PR.

xoac:
so feel free to close this if it's not useful

shenshouer:
Any docs or example for how to timeout when request?

jxs:
you just have to send a timeout [`StatusCode`](https://docs.rs/http/0.2.1/http/status/struct.StatusCode.html#associatedconstant.REQUEST_TIMEOUT)

PR #115, opened 2018-10-11
rework Rejection system

seanmonstar:
The `Rejection` type can now nest and combine arbitrary rejections,
so it is no longer bound to a small set of meanings. The ranking of
status codes is still used to determine which rejection gets priority.

A different priority can be implemented by handling rejections with
a `Filter::recover`, and searching for causes in order via
`Rejection::find_cause`.

- Adds `warp::reject::custom()` to create a `Rejection` with
  any `Into<Box<std::error::Error>>`. These rejections should be
  handled with an eventual `Filter::recover`. Any unhandled
  custom rejections are considered a server error.
- Deprecates `Rejection::with`. Use custom rejections instead.
- Deprecates `Rejection::into_cause`, as it can no longer work. Always
  returns `Err(Rejection)`.
- Deprecates `Rejection::json`, since the format needed is too generic.
  The `errors.rs` example shows how to send custom JSON when recovering
  from rejections.
- Deprecates `warp::reject()`, since it current signals a `400 Bad
  Request`, but in newer versions, it will signal `404 Not Found`.
  It's deprecated simply to warn that the semantics are changing,
  but the function won't actually go away.
- Deprecates `reject::bad_request()`, `reject::forbidden()`, and
  `reject::server_error()`. Uses custom rejections instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,6 @@ categories = ["web-programming::http-server"]
 keywords = ["warp", "server", "http", "hyper"]
 
 [dependencies]
-bitflags = "1"
 bytes = "0.4.8"
 futures = "0.1"
 #headers = { path = "../headers" }
--- a/examples/errors.rs
+++ b/examples/errors.rs
@@ -1,18 +1,25 @@
 #![deny(warnings)]
-
+extern crate serde;
+#[macro_use] extern crate serde_derive;
 extern crate pretty_env_logger;
 extern crate warp;
 
 use std::error::Error as StdError;
 use std::fmt::{self, Display};
 
-use warp::{Filter, reject, Rejection, Reply};
+use warp::{Filter, Rejection, Reply};
 use warp::http::StatusCode;
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug)]
 enum Error {
     Oops,
     NotFound
+}
+
+#[derive(Serialize)]
+struct ErrorMessage {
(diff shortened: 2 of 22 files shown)


seanmonstar:
Here's the updated change:

----

The `Rejection` type can now nest and combine arbitrary rejections,
so it is no longer bound to a small set of meanings. The ranking of
status codes is still used to determine which rejection gets priority.

A different priority can be implemented by handling rejections with
a `Filter::recover`, and searching for causes in order via
`Rejection::find_cause`.

- Adds `warp::reject::custom()` to create a `Rejection` with
  any `Into<Box<std::error::Error>>`. These rejections should be
  handled with an eventual `Filter::recover`. Any unhandled
  custom rejections are considered a server error.
- Deprecates `Rejection::with`. Use custom rejections instead.
- Deprecates `Rejection::into_cause`, as it can no longer work. Always
  returns `Err(Rejection)`.
- Deprecates `Rejection::json`, since the format needed is too generic.
  The `errors.rs` example shows how to send custom JSON when recovering
  from rejections.
- Deprecates `warp::reject()`, since it current signals a `400 Bad
  Request`, but in newer versions, it will signal `404 Not Found`.
  It's deprecated simply to warn that the semantics are changing,
  but the function won't actually go away.
- Deprecates `reject::bad_request()`, `reject::forbidden()`, and
  `reject::server_error()`. Uses custom rejections instead.

PR #120, opened 2018-10-20
Rename `warp::path::index` to `warp::path::end`

### LINKED ISSUE #118: Using `path` vs `index` for files
I might not understand the use of `index()` correctly. In the `files.rs` example the route to the file is defined with 
```
let readme = warp::get2()
        .and(warp::index())
        .and(warp::fs::file("./README.md"));
```
but what if you have multiple files that should be routed to, for instance, `/file1` and `/file2`? When I tried to use index for this:
```
let file1 = warp::get2()
              .and(warp::index())
             .and(warp:fs::file("./file1"));

let file2 = warp::get2()
              .and(warp::index())
             .and(warp:fs::file("./file2"));
```
only the first defined route in `let routes = file1.or(file2)` is ever rendered and navigating to `/my/project/root/file2` returns file not found.

If instead I define the routes with 
```
let file1 = warp::path("file1")
        .and(warp::fs::file("./file1"));
let file2 = warp::path("file2")
        .and(warp::fs::file("./file2"));
let routes = file1.or(file2);
```
I get access to both files. It might be helpful to adjust the example to show this explicitly. Also please correct me if I'm not using the `index()` filter correctly which caused my trouble.

seanmonstar:
Closes #118

### DIFF
--- a/examples/errors.rs
+++ b/examples/errors.rs
@@ -43,7 +43,7 @@ impl StdError for Error {
 
 
 fn main() {
-    let hello = warp::path::index()
+    let hello = warp::path::end()
         .map(warp::reply);
 
     let oops = warp::path("oops")
--- a/examples/file.rs
+++ b/examples/file.rs
@@ -8,7 +8,7 @@ fn main() {
     pretty_env_logger::init();
 
     let readme = warp::get2()
-        .and(warp::index())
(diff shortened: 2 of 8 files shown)


theduke:
Just as a side note, this was a breaking change and should probably have been saved for `0.2`.

PR #157, opened 2019-01-21
add graceful shutdown example

daaku:
Provides a more complete example using an external `signal-hooks` crate that shows how to trigger a clean shutdown for the common case of unix signals.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ urlencoding = "1.0.0"
 pretty_env_logger = "0.2"
 serde_derive = "1.0"
 handlebars = "1.0.0"
+signal-hook = { version = "0.1", features = ["tokio-support"] }
 
 [features]
 tls = ["rustls"]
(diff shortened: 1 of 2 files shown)


daaku:
Just fyi, I won't feel offended at all if you're not interested in merging this especially considering it involves a third party crate.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ urlencoding = "1.0.0"
 pretty_env_logger = "0.2"
 serde_derive = "1.0"
 handlebars = "1.0.0"
+signal-hook = { version = "0.1", features = ["tokio-support"] }
```
ricochet1k: At the very least, this should be an optional dev-dependency required by only the graceful_shutdown example.

seanmonstar:
I'd rather just point at [the hyper guide](https://hyper.rs/guides/server/graceful-shutdown/) instead of adding another dev-dependency, personally.

PR #158, opened 2019-01-21
rustfmt

remexre:


examples/errors.rs
```diff
@@ -41,27 +42,20 @@ impl StdError for Error {
...
-    let hello = warp::path::end()
-        .map(warp::reply);
+    let hello = warp::path::end().map(warp::reply);
 
-    let oops = warp::path("oops")
-        .and_then(|| {
-            Err::<StatusCode, _>(warp::reject::custom(Error::Oops))
-        });
+    let oops =
+        warp::path("oops").and_then(|| Err::<StatusCode, _>(warp::reject::custom(Error::Oops)));
```
seanmonstar: Huh, that's a weird choice... Is it possible to tell rustfmt to line break after the function call instead after an equals sign?

remexre: Looking through the rustfmt docs, I don't see anything that looks like it'd do it; I agree this is a suboptimal formatting, though...

PR #162, opened 2019-01-30
impls ReplySealed for Vec<u8> and &'static [u8].

remexre:
what it says on the tin.

tbh, would it make sense to `impl<T: Into<hyper::Body>> ReplySealed for T`?

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -414,8 +414,10 @@ impl Rejections {
             Rejections::Known(ref e) => {
                 let mut res = http::Response::new(Body::from(e.to_string()));
                 *res.status_mut() = self.status();
-                res.headers_mut()
-                    .insert(CONTENT_TYPE, HeaderValue::from_static("text/plain"));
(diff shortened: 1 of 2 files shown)


seanmonstar:
I've figured that the `Content-Type` should probably be set automatically for `String` and `Vec<u8>` replies... What do you think of that?

remexre:
Set automatically to what? ftr, I'm using the `&'static [u8]` with [packer](https://github.com/iptq/packer) to serve a bunch of statics; I'm setting the content-type based on file extension (since it's known to be "true" at compile time)

seanmonstar:
I would assume that:

- `String` implies `text/plain; charset=utf-8`
- `Vec<u8>` implies `application/octet-stream`

seanmonstar:
Looks good, though the CI failures are legit.

remexre:
whoops, yeah, I wrote it from class :P should be good now

seanmonstar:
Cool, thanks!

PR #166, opened 2019-02-01
Add a `multi_param` filter constructor for multiple-valued path parameters

g-s-k:
I ran into a spot where I wanted to add an endpoint that looked like `/foo,bar,.../do_baz` - this seemed like a pretty simple logical extension of accepting one parameter. What I had to do with the current API is sort of clunky, and needed to be done for each endpoint that required it:

```rust
let do_foo = warp::path::param().and_then(|thing: String| {
    match thing.split(',').map(MyType::from_str).collect::<Result<Vec<_>, _>>() {
        Ok(ok) => Ok(ok),
        Err(err) => Err(warp::reject::custom(err)),
    }
});
```

This feels like a common enough pattern to include as a filter, and it doesn't end up being all that much code.

What I'm not toooootally sure about is the macro syntax. I added a branch to the `path!` macro with the syntax `[<type>]` to represent a multiple-valued segment - I'd like to know if that's valuable, if it should be changed to a different idiom, or if I should just throw it in the trash. (Small aside: I know that `[type]` means an array of unspecified size, but that's not an issue because those can't be passed as function params.)

Thanks! I really like working with this library, and I'm excited to have something to contribute.

### DIFF
--- a/examples/path.rs
+++ b/examples/path.rs
@@ -1,0 +1,32 @@
+#![deny(warnings)]
+extern crate warp;
+
+use warp::{path, Filter};
+
+fn main() {
+    // Match a single parameter
+    // e.g. `/add_3/39` returns "42"
+    let one = path!("add_3" / usize).map(|num| format!("{}", num + 3));
+
+    // Match two parameters in different segments
+    // e.g. `/pythag/5/and/12` returns "13"
+    let two =
+        path!("pythag" / f64 / "and" / f64).map(|num0: f64, num1| format!("{}", num0.hypot(num1)));
+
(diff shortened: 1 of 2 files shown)


g-s-k:
I agree that multi-value params don't have first-class support in other routers - however, I have had to work on several APIs where the concept was used. The list would just be accepted as a single string param, then split and parsed further down the pipeline - like the example above. This is fine, but I feel like the `warp` paradigm is to enforce types as a first step, and to catch errors as early as possible.

It's not a deal breaker to have to implement this in my app, but it would be more convenient to have it available for the next project, and the next, and the next...where I invariably need the ability to act on a set of multiple keys in a single request.

PR #167, opened 2019-02-01
Breaking Changes for 0.2

### LINKED ISSUE #109: Filter::or with a filter's error of Never should also have Never as error
If you chain 2 filters with an `or`, and the latter one can *never* fail, it is clear that the chained filter likewise can *never* fail. However, this is currently not true:

```rust
fn v1() -> impl Filter<Extract=(Option<u32>,), Error=Rejection> {
    warp::path::param::<u32>()
        .map(Some)
        .or(warp::any().map(|| None))
}
```

It **should** be like this:

```rust
fn v1() -> impl Filter<Extract=(Option<u32>,), Error=Never> {
    warp::path::param::<u32>()
        .map(Some)
        .or(warp::any().map(|| None))
}
```

I've made changes locally that makes this work, by modifying `CombineRejection`, but I suspect this change may subtly break some people...

### LINKED ISSUE #117: 0.2 Breaking Changes Tracking Issue
Through usage of 0.1, some breaking changes have been found to be needed. This is to track that all are dealt with when 0.2 is released.

- [x] Replace `get`, `post`, `put`, and `delete` with the new `get2`, `post2`, `put2`, and `delete2` filters.
- [x] Remove `warp::cookie::optional_value`.
- [x] Replace `ws` with `ws2`, `Ws` with `Ws2`.
- [x] Remove deprecated functions from `reject` module.
- [x] Remove `Rejection::cause()`.
- [x] Remove `warp::path::index`.
- [ ] Add `Transport` bound to `Server::serve_incoming` and `Server::run_incoming`.

seanmonstar:
Closes #117 
Closes #109 

The one missing piece from #117 that I'm undecided about is changing `Server::run_incoming` and `Server::serve_incoming` to require `Transport`... It could just be annoying for people who don't care...

### DIFF
--- a/examples/body.rs
+++ b/examples/body.rs
@@ -17,7 +17,7 @@ fn main() {
     pretty_env_logger::init();
 
     // POST /employees/:rate  {"name":"Sean","rate":2}
-    let promote = warp::post2()
+    let promote = warp::post()
         .and(warp::path("employees"))
         .and(warp::path::param::<u32>())
         // Only accept bodies smaller than 16kb...
--- a/examples/errors.rs
+++ b/examples/errors.rs
@@ -51,7 +51,7 @@ fn main() {
     let nope =
         warp::path("nope").and_then(|| Err::<StatusCode, _>(warp::reject::custom(Error::Nope)));
 
-    let routes = warp::get2()
+    let routes = warp::get()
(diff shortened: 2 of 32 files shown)


seanmonstar:
Each commits has been individually rebased onto master.

PR #174, opened 2019-02-25
Emit log messages when query string fails to deserialize

### LINKED ISSUE #173: Log error if query string cannot be deserialized
At the moment, any error produced by `serde_urlencoded` is just swalled by the `.ok()` call. 

https://github.com/seanmonstar/warp/blob/c850b563864aebc28a51b0af9b5490918ae56259/src/filters/query.rs#L18

This makes things hard when trying to debug why your path is not hitted as expected.

I'd propose to add a `.map_err(|e| warn!("query string failed to deserialize: {:?}", e))` before doing the `.ok()` so that the error gets at least logged.

thomaseizinger:
I also added a log message if the implementation defaults to an empty string if there is none provided.

While doing the code, I noticed that the behavior for the 2nd method is different. Is that on purpose?

Resolves #173.

### DIFF
--- a/src/filters/query.rs
+++ b/src/filters/query.rs
@@ -12,12 +12,15 @@ use reject::{self, Rejection};
 pub fn query<T: DeserializeOwned + Send>() -> impl Filter<Extract = One<T>, Error = Rejection> + Copy
 {
     filter_fn_one(|route| {
-        route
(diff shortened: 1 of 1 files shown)


seanmonstar:
Thanks!

PR #192, opened 2019-03-20
Configurable tls support

prasannavl:
Added more flexible tls support.

- **Features**:
  - Configurable tls support

- **Changes**
  - Expose `tls` module
  - Add `tls::config`, `tls::config_from_path`, `tls::config_from_der_data`.
  - Add `tls2` method on `warp::Server` to take `ServerConfig` instead of path only
    `tls` variant.
  - `tls::configure` has been deprecated
  - `tls` method of `warp::Server` has been deprecated.

This is relatively small but breaking change (eventually - currently non-breaking with deprecation notes) for tls.  So, took the liberty to document it and do a version bump. If that isn't acceptable, please let me know and I'll look at how to restructure this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+### v0.2.0 (March 20, 2019)
+
+- **Features**:
+  - Configurable tls support
+
+- **Changes**
+  - Expose `tls` module
+  - Add `tls::config`, `tls::config_from_path`, `tls::config_from_der_data`
+  - Change `tls` method of `wrap::Server` to take `ServerConfig` instead of path only
+  - Removed `tls::configure` function
+
 ### v0.1.14 (March 19, 2019)
 
 - **Features**:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,15 @@
-### v0.2.0 (March 20, 2019)
+### v0.1.15 (March 20, 2019)
 
 - **Features**:
   - Configurable tls support
 
 - **Changes**
   - Expose `tls` module
-  - Add `tls::config`, `tls::config_from_path`, `tls::config_from_der_data`
-  - Change `tls` method of `wrap::Server` to take `ServerConfig` instead of path only
-  - Removed `tls::configure` function
+  - Add `tls::config`, `tls::config_from_path`, `tls::config_from_der_data`.
+  - Add `tls2` method on `warp::Server` to take `ServerConfig` instead of path only
+    `tls` variant.
+  - `tls::configure` has been deprecated
+  - `tls` method of `warp::Server` has been deprecated.
 
 ### v0.1.14 (March 19, 2019)
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "warp"
-version = "0.2.0" # don't forget to update html_root_url
+version = "0.1.15" # don't forget to update html_root_url
 description = "serve the web at warp speeds"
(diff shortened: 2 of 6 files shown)


seanmonstar:
Thanks for the PR! I'm excited to improve TLS support.

I'd probably prefer to not expose the `rustls::ServerConfig` type (or really anything from `rustls`), so that we can upgrade to new versions without it being a breaking change. So, zooming out, what is your goal? Just to configure a few more options?

(Don't worry about any of the version stuff, I do all that right when cutting a release.)

prasannavl:
@seanmonstar - I wasn't very happy to expose the rustls stuff as well. Thought, a similar approach to how `actix_web` does, with a connector is probably more appropriate. However currently, I embed certain binary certificates, and using it right now with warp isn't very obvious creating a TlsServer manually. 

So, though would do this as a stop gap before a more generic abstraction that hides the tls ones. Or (understandably) if you rather think it's better to avoid a stop gap, have you given any thought to how you'd like the tls api to look like?

prasannavl:
Alternatively, my specific problem can also be solved by just adding another function taking a `Vec<u8>`, or `[u8]` on the outer api. But that limits configuration of any other kind until a proper tls redesign.

seanmonstar:
I hadn't thought too hard about the existing TLS API, it was mostly added as people were asking for it, and I focused on convenience over configuration.

What about if there were a new `Tls` type added that were a builder, and supplied to `Server::tls`? It'd be an improvement by having `key_path` and `cert_path` methods, since currently you could mess up the order. But then it could also have methods for supplying the key or cert directly as bytes, like you want...

seanmonstar:
What do you think of my suggested alternative?

prasannavl:
I think that's a far better approach than the current PR :) 
A bit on a time crunch, but let me take it up this weekend.

prasannavl:
I've been mulling over this a bit. Thinking more on this, I feel like the actix model is a probably the best approach. It basically gates `bind_tls`, `bind_ssl` under feature flags, that works with rustls and openssl. 

Considering that rust community has converged around rustls for a rust native solution, and with native-ssl utilising openssl, I think this is probably the most optimal approach. If we go down this path, I think it's just better to expose the rustls, and openssl config directly, as otherwise it's impossible to cover complicated cases and unnecessarily complicates warp. While warp abstracting this sounds nice in theory, I don't think it's very practical to cover all cases, unless this project is ready to take on the nuances of TLS configurations as well. 

Now, this can be abstracted a little bit with the builder pattern above such that this responsibility is pushed on to the builder with the feature gated methods inside it rather than the Server itself exposing the tls configs of upstream libs -- however I think it's inevitable that they have to be exposed, or else warp will have to deal with cipher suites, versions, and the whole shebang.

Thoughts?

seanmonstar:
> While warp abstracting this sounds nice in theory, I don't think it's very practical to cover all cases, unless this project is ready to take on the nuances of TLS configurations as well.

Yea, I get what you mean. However, the *reason* I want to abstract over it for *now*, is so that when rustls needs a version upgrade that would normally require a breaking change, warp can upgrade, and since it was never exposed publicly, it doesn't require warp to *also* issue a breaking change. If/when rustls has a stable 1.0, I'd be more open to exposing the dependency. (This stance is based off experience of when hyper used to expose openssl as a public dependency.)

prasannavl:
Excellent. That sounds reasonable. Will update the PR in the next couple of days in line with the above discussions. :)

seanmonstar:
Going to close up here, feel free to open a PR with new changes!

PR #193, opened 2019-03-21
impls ReplySealed for Cow<'static, str>.

shota-f:
Hi, thanks for the great warp lib!
I think it is useful if the enum Cow which represents string implements Reply(Sealed).

In my case, it is especially nice when using the with_status method to make a impl Reply like a WithStatus.
If the Cow implements Reply, Error(assume enum) which has some static str, String, or Cow, and which is underlying a Rejection can provide first argument of the method without unnecessary allocation.

I implemented it by using existing impl for static str and String.


seanmonstar:
Cool, sounds good to me, thanks!

PR #205, opened 2019-04-15
Skip headers for Not Modified fs responses

xfix:
According to RFC 7232, 304 Not Modified responses should not generate representation metadata other than the following fields: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.

This avoids a warning from hyper when logging is enabled: "content-length value found, but empty body provided". While it is allowed to provide Content-Length for Not Modified responses it is unnecessary.


xfix:
Depends on https://github.com/hyperium/hyper/issues/1797 to be fixed, after actually testing this in practice.

Issue #204, opened 2019-04-14
Add support for custom keepalive comments in SSE


codesections:
The current SSE filter has a `keep` function that sends a keepalive comment every interval.  Right now, that comment is hard-coded as `:`.

I would like to add support for a custom comment as the keepalive message (as a third parameter to `keep`).  The [relevant portion of the spec](https://www.w3.org/TR/2009/WD-eventsource-20090421/#notes) states that "authors can include a comment line (one **starting with** a ':' character) every 15 seconds or so" (emphasis supplied).  Thus, it's compliant with the spec to have a custom message.

My use case is that I'm porting over an existing API, and it's current SSE heartbeat message is `:thump`; I'd prefer to port that over without changing the message.

If you'd be open to a PR along these lines, I would be happy to work on something—I believe it would be a minor change.

If you would prefer, we could add a `keep_with_message` function to avoid changing the function signature of `keep`.  (Or a `keep2`.)

seanmonstar:
That sounds reasonable. Looking over the API again, perhaps `keepalive` (or `keep_alive`?) would be a clearer name over `keep`? In that sense, we could add the second version with the new name (and not need the uglier `keep2`), deprecating `keep`.

What sort of argument would make sense? It'd be nice to be able to easily pass the default...

codesections:
I terms of being able to easily pass the default, it seems like the typical go-to solution is the builder pattern.  That would suggest an API along the following lines:

```rust
keep_alive::from(stream);
```
as the default, and
```rust
keep_alive::from(stream)
    .with_duration(30)
    .with_text("thump".to_string());
```
as the custom.  (One related change implicit in that suggested API is to have the duration take an integer number of seconds rather than an `impl Into<Option<Duration>>`.  It seems like overkill to require users to pull in the `Duration` module given that keep-alive durations will necessarily be some number of seconds rather than, say, hours or days.)

Let me know if you'd like any changes to that API.

seanmonstar:
Sure, a builder pattern seems fine. I'd suggest not having the `from` part, but rather instead of how it currently returns an `impl Stream`, return the `SseKeepAlive` type directly, and add methods that consume `self` to adjust the duration and text.

[...]

PR #206, opened 2019-04-16
Add keep_alive

codesections:
This pull request adds `keep_alive`, which is an updated version of `keep` that
has been modified to use the builder pattern (as discussed in #204).  This removes the need to pass a `None` to get the default interval of 15 seconds and allows users to customize the text of the keep-alive comment.

### DIFF
--- a/src/filters/sse.rs
+++ b/src/filters/sse.rs
@@ -24,14 +24,15 @@
 //!             warp::sse::retry(Duration::from_millis(5000)),
 //!         ).into_b().into_b(),
 //!     ]);
-//!     sse.reply(warp::sse::keep(events, None))
+//!     sse.reply(warp::sse::keep_alive(events))
 //! });
 //! ```
 //!
 //! Each field already is event which can be sent to client.
 //! The events with multiple fields can be created by combining fields using tuples.
 //!
-//! See also [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) API.
+//! See also the [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) API,
+//! which specifies the expected behavior of Server Sent Events.
 //!
 use std::error::Error as StdError;
 use std::fmt::{self, Display, Formatter, Write};
@@ -74,7 +75,8 @@ pub trait ServerSentEvent: SseFormat + Sized + Send + 'static {
 impl<T: SseFormat + Send + 'static> ServerSentEvent for T {}
 
 #[allow(missing_debug_implementations)]
-struct SseComment<T>(T);
+/// Comment field (":<comment-text>")
+pub struct SseComment<T>(T);
 
 /// Comment field (":<comment-text>")
 pub fn comment<T>(comment: T) -> impl ServerSentEvent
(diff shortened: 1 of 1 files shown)


src/filters/sse.rs
```diff
@@ -449,21 +451,36 @@ where
...
+/// Struct for tracking the interval between keep-alive messages, the content
+/// of each message, and the associated stream.
+pub struct SseKeepAlive<S> {
     event_stream: S,
+    comment_text: String,
     max_interval: Duration,
     alive_timer: Delay,
 }
+impl<S> SseKeepAlive<S> {
+    /// Customize the interval between keep=alive messages.
```
seanmonstar: Should it say "keep-alive"?

src/filters/sse.rs
```diff
@@ -449,21 +451,36 @@ where
...
+    pub fn with_interval(mut self, time: Duration) -> Self {
```
seanmonstar: The builder pattern in Rust usually doesn't include `with_` prefixes.

src/filters/sse.rs
```diff
@@ -449,21 +451,36 @@ where
...
+
+/// This is a deprecated version of `keep_alive`, which is maintained for
+/// backwards compatibility.  Refer to the documentation accompanying `keep_alive`
+/// for details about the purpose/usage of these two functions.
 ///
-/// See [notes](https://www.w3.org/TR/2009/WD-eventsource-20090421/#notes).
+/// Unlike `keep_alive`, `keep` requires you to provide an `Option<Durration>` as a
+/// second parameter and interprets `None` as a signal to use a default durration
+/// of 15 seconds.  `keep` also always sends the heartbeat signal `:` whereas
+/// `keep_alive` alows you to customize that signal.
```
seanmonstar: Care to add `#[deprecated(note = "use keep_alive instead")]`?

src/filters/sse.rs
```diff
@@ -483,6 +500,60 @@ where
     let alive_timer = Delay::new(now() + max_interval);
     SseKeepAlive {
         event_stream,
+        comment_text: "".to_string(),
+        max_interval,
+        alive_timer,
+    }
+}
+
+/// Keeps event source connection when no events sent over a some time.
```
seanmonstar: Suggestion: add "alive" between "connection" and "when".

src/filters/sse.rs
```diff
@@ -74,7 +75,8 @@ pub trait ServerSentEvent: SseFormat + Sized + Send + 'static {
 impl<T: SseFormat + Send + 'static> ServerSentEvent for T {}
 
 #[allow(missing_debug_implementations)]
-struct SseComment<T>(T);
+/// Comment field (":<comment-text>")
+pub struct SseComment<T>(T);
```
seanmonstar: I'm guessing this needed to become public since it's exposed in the `Stream` implementation of `SseKeepAlive`? I'd prefer to keep it private...

Things that need to be "public" but still private are in the `sealed` mod down below. (They have a `pub` modifier, but since `sealed` isn't public, they're not accessible outside the crate). We could either move this struct into the `sealed` module, or put some `struct SseKeepAliveItem` there instead. Which seems cleaner?

src/filters/sse.rs
```diff
@@ -449,21 +451,36 @@ where
...
 
-/// Keeps event source connection when no events sent over a some time.
-///
-/// Some proxy servers may drop HTTP connection after a some timeout of inactivity.
-/// This function helps to prevent such behavior by sending dummy events with single
-/// empty comment field (i.e. ":" only) each `keep_interval` of inactivity.
-///
-/// By default this time interval between events equals to 15 seconds.
+    /// Customize the text of the keep-alive message.
+    pub fn with_text(mut self, text: String) -> Self {
```
seanmonstar: I don't have a strong opinion here, but: it looks like `sse::comment` is able to be any `Display` type. Should that be allowed here also?

src/filters/sse.rs
```diff
@@ -449,21 +451,36 @@ where
...
+pub struct SseKeepAlive<S> {
     event_stream: S,
+    comment_text: String,
     max_interval: Duration,
     alive_timer: Delay,
 }
+impl<S> SseKeepAlive<S> {
+    /// Customize the interval between keep=alive messages.
+    pub fn with_interval(mut self, time: Duration) -> Self {
+        self.max_interval = time;
```
seanmonstar: I think this will also need to adjust the `alive_timer`.

seanmonstar:
Thanks for the initial work! I adjusted this a little and merged as #213.

PR #208, opened 2019-04-23
Un-hide reexports

### LINKED ISSUE #207: Confusing missing documentation for beginners
Just a little heads up. I'm just starting with `warp` today, and these items missing from the docs was a little confusing for me at first:

https://github.com/seanmonstar/warp/blob/4a1c67c4ca0a22e263dcf2d94ee64a36a4682e57/src/lib.rs#L127-L132

At first I was thinking that maybe the usage of warp, as explained in the examples, has been deprecated and there is now a different blessed way to use the filters, and I actually downloaded and compiled all the examples before I noticed that comment and that the filters were simply hidden from the docs.

Maybe you'd want to put them back into the docs, so that, while being verbose, it is nice and clear for everybody.

NeoLegends:
This fixes #207. If you don't want to unhide the items, just close this & the corresponding issue. :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,10 +124,7 @@ mod transport;
 
 pub use self::error::Error;
 pub use self::filter::Filter;
-// This otherwise shows a big dump of re-exports in the doc homepage,
-// with zero context, so just hide it from the docs. Doc examples
-// on each can show that a convenient import exists.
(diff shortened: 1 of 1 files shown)


seanmonstar:
I'm going to close for now, discussion in #207. Thanks!

PR #229, opened 2019-05-19
Update tungstenite

Eijebong:


Eijebong:
Sure, done

src/filters/ws.rs
```diff
@@ -260,6 +256,10 @@ impl Stream for WebSocket {
                 protocol::Message::Pong(payload) => {
                     trace!("websocket client pong: {:?}", payload);
                 }
+                protocol::Message::Close(frame) => {
+                    trace!("websocket closed: {:?}", frame);
+                    return Ok(Async::Ready(None));
```
application-developer-DA: I'm not sure that it's correct to return `None` here. `Message::Close` means that we received a `Close` message, so I expect that the user received the `Close` message back (because basically that's what happened).

I think the more correct way to handle it would be to impement it the same way it's done in `tokio-tungstenite` : https://github.com/snapview/tokio-tungstenite/commit/e2cdf11bc8d8d0837d2073ecc3db686c13435523 , i.e.

`Message::Close` -> `Ready(Some(Message::Close))`
`ConnectionClosed` -> `Ready(None)`

application-developer-DA:
@Eijebong I'm not sure that the implementation is correct, so I added some comment.

BTW, we also have `tokio-tungstenite`, which basically wraps the `tungstenite-rs` and makes it available for `tokio`. It seems that `warp` has a similar implementation, perhaps it would make sense to use `tokio-tungstenite` in `warp`. We normally update `tokio-tungstenite` after doing changes in `tungstenite-rs`, which means that in case of any changes, you won't have to do any additional changes in `warp`.

src/filters/ws.rs
Eijebong: Well I don't think I changed anything behavior-wise. The `Err(ConnectionClosed)` was changed to `Ok(Message::Closed)` AFAICT, this is just updating a dependency, not trying to fix any behavior. Unless I'm missing something, it should be another issue / MR to fix that.

application-developer-DA: The `Err(ConnectionClosed)` was not replaced by `Message::Closed`. We changed the semantics a bit. In the previous version, the `ConnectionClosed` was returned after the connection is closed and it contained an optional close code.

Since there were requests to support sending/receiving `Closed` as a message (and also it was required for `tokio-tungstenite` to make it compatible with the "expected semantics" of `Stream`), we decided to make `Closed` as a part of the `Message`.

Because of that, the `ConnectionClosed` does not contain an optional close code in it anymore, but it still can be returned from (see https://github.com/snapview/tungstenite-rs/commit/b8f7d3668ec79c62abb9562ee3b5dd355908b6a5).

Eijebong: Oh ok, my bad then. Thanks for the explanation. I'll fix it tonight when I get home

Eijebong:
@application-developer-DA Does it look better to you now ?

application-developer-DA:
@Eijebong yes, I think that usage of `ConnectionClosed` is correct now, but I don't think that returning `Ok(Async::Ready(None))` when `Message::Close` is received is a good idea. IMO the `Message::Close` should be returned to user. Which unit test fail? Perhaps we need to update some of them.

Basically I see only 2 issues with returning `Async::Ready(None)` instead of `Message::Close` to the user when the `Message::Close` is received from tungstenite-rs:

1. The original enum does contain this message and I think it might be confusing for the user that there is no way to investigate the `Close` message when the server receives it.
2. If the user drops the object right after the stream returns `Async::Ready(None)`, it might pose some issues as the connection on the server side might be dropped earlier than it should.

Other than that, it looks good.

src/filters/ws.rs
```diff
@@ -241,8 +241,7 @@ impl Stream for WebSocket {
                 Err(::tungstenite::Error::Io(ref err)) if err.kind() == WouldBlock => {
                     return Ok(Async::NotReady);
                 }
-                Err(::tungstenite::Error::ConnectionClosed(frame)) => {
-                    trace!("websocket closed: {:?}", frame);
+                Err(::tungstenite::Error::ConnectionClosed) => {
                     return Ok(Async::Ready(None));
```
application-developer-DA: Shouldn't the `trace!("websocket closed: {:?}", frame);` be logged here?

The websocket is closed when we receive `ConnectionClosed`, when the server receivs `Message::Close` it does not automatically mean that the websocket is closed, but rather that we received a "request" to close the websocket connection, i.e. it's like a message from the client: "Hey Server, I want to close the connection, please confirm and drop the TCP connection".

Eijebong:
The inner message isn't exposed to the user, returning a message with `Message::Close` in it would be confusing as the user would have no idea what kind of message was returned as all the `is_` functions would return false. @seanmonstar Should I add a `is_close` method ?

The failing test is the following:

```
#[test]                                                                    
fn closed() {                                                              
    let _ = pretty_env_logger::try_init();                                 
                                                                           
    let route = warp::ws2().map(|ws: warp::ws::Ws2| {                      
        ws.on_upgrade(|websocket| {                                        
            websocket                                                      
                .close()                                                   
                .map_err(|e| panic!("close error: {:?}", e))               
        })                                                                 
    });                                                                    
                                                                           
    let mut client = warp::test::ws().handshake(route).expect("handshake");
                                                                           
    client.recv_closed().expect("closed");                                 
}                                                                          
```

It fails because `test::WsClient::recv_closed()` checks for the stream to return None and not `Message { inner: Message::Close }`.

application-developer-DA:
> The inner message isn't exposed to the user, returning a message with Message::Close in it would be confusing as the user would have no idea what kind of message was returned as all the is_ functions would return false. @seanmonstar Should I add a is_close method ?

Well, the original crate (`tungstenite-rs`) does contain this method https://docs.rs/tungstenite/0.8.1/tungstenite/enum.Message.html#method.is_close . I'm still thinking that it would make sense to use `tokio-tungstenite` directly in `warp`, that would prevent doing the changes twice after each update of `tugstenite-rs`, but I think it's a different topic and should be handled in a separate pull request (most likely some changes in `tokio-tungstenite` would be required in order to make it suitable for `warp`).

> It fails because test::WsClient::recv_closed() checks for the stream to return None and not Message { inner: Message::Close }.

If we're talking about the websocket client, then I think that from the API perspective it will look as follows:

1. The client receives close message from the server when the server triggers a close (so the client receives `Message::Close`).
2. Then it'll get `ConnectionClosed` which `warp` converts into `Async::Ready(None)`.

So if the semantics of `recv_closed()` assumes the server is closing the connection, then the internal implementation should wait for `Message::Close` and then `ConnectionClosed`. Does it make any sense?

seanmonstar:
Is there benefit to the details of the close frame? With the comparison to TCP, we don't show the FIN frame, but instead just Ok(0).

application-developer-DA:
Well, it's just a matter of mathematical completeness. The `Message` enum has a `Close` message, also the `Message` is a type for the item for both `Stream` and `Sink`, I would expect that if I can send `Close`, then the reception of `Close` is also possible (unless otherwise stated in protocol description). Also some apps may take some actions depending on the `CloseReason` contained in the `Close` message.

Also I don't think that it's hard to implement this workflow as there are almost no changes required to support returning of `Message::Close` to the user (it's basically already supported), and the `ConnectionClosed` can be used as an "end of stream" (`Async::Ready(None)`).

As far as I understood, only one function from `test` module should be edited in order to pass all tests, that's basically the only change which is still required, right?

Eijebong:
@application-developer-DA I tried changing the test and rturn the close message. It works but now some thread is panicking from the binary test I guess because the connection gets closed improperly.

application-developer-DA:
@Eijebong do you by chance know what exactly fails? It might point to some other issue. The close of underlying TCP connection normally happens when the object is dropped (at least it's so for `tungstenite-rs`).

Eijebong:
@application-developer-DA I get this

```
[eijebong@bananium] ~/code/warp  (master) >>> target/debug/deps/ws-713b42647f74bebd --nocapture

running 5 tests
test fail ... ok
test upgrade ... ok
test closed ... ok
test binary ... ok
thread 'test binary; warp-test-runtime-0' panicked at 'websocket error: ConnectionClosed', tests/ws.rs:98:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
test text ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

thread 'thread 'binarytest text; warp-test-runtime-0' panicked at '' panicked at 'websocket forward: ()', websocket error: ConnectionClosedsrc/libcore/result.rs', :tests/ws.rs999::985:
17
```

`RUST_BACKTRACE=1` doesn't print any backtrace...

seanmonstar:
For now, I'd suggest that warp continues to interpret a `Message::Close` received from a client as the websocket is closing, and return `Ok(Async::Ready(None))` to the user. I don't believe there's much a server would do with the close frame, and that way there isn't subtle differences between receiving `Some(Message::Close)` vs `None`, which implies the stream is finished anyways.

Eijebong:
There, came back to the first version of the patch

application-developer-DA:
> I don't believe there's much a server would do with the close frame

Hm, in our product we rely on the close codes, depending on the way the websocket client was closed. We handle the close code and use the reason to understand why the client closed the connection and do some actions depending on that.

> and that way there isn't subtle differences between receiving Some(Message::Close) vs None, which implies the stream is finished anyways.

Yeah, but the question is how do we ensure that we comply to the RFC 6455 (The WebSocket Protocol)?

Let's assume that we implement it the way you suggest, i.e. `Message::Close` is converted to the `Async::Ready(None)` and returned to the user, the questions are:

1. What will happen to the WebSocket object after that? Will it be destroyed immediately? - If no, then when will it be destroyed? (when the underlying TCP connection will be closed?)
2. How do we ensure that the server replies with the close code back to the client?

There are 2 important things which we have to keep in mind:

```
If an endpoint receives a Close frame and did not previously send a
   Close frame, the endpoint MUST send a Close frame in response. 
```
```
After both sending and receiving a Close message, an endpoint
   considers the WebSocket connection closed and MUST close the
   underlying TCP connection.  The server MUST close the underlying TCP
   connection immediately; the client SHOULD wait for the server to
   close the connection
```

If we just interpret `Message::Close` as `Async::Ready(None)` and drop the connection immediately, then we can't ensure that server's response (close message) has been successfully sent back to the client.

However if we don't close the connection immediately after receiving `Async::Ready(None)`, then we don't know where and when to drop the connection.

`tungstenite-rs` gives back the `Message::Close` when we receive a close message (and that's why we don't have to treat it in a special way and IMO should return it to the user), also `tungstenite-rs` "schedules" the response to the client to comply with the RFC. So the one should call `read_message()` after that until it returns `ConnectionClosed`, which means that now the WebSocket connection is meant to be closed and that the server must drop the object (the underlying TCP connection should be closed).

TL;DR

In the previous version of `tungstenite-rs` the close message was completely hidden from the user, it was not part of the `Message` enum, which means that the user could neither send, nor receive `Close` as a separate `Message` by sending/receiving it via `Sink`/`Stream`. The only way to investigate the close code was to wait for `ConnectionClosed(Option<Close>)` and investigate the `Close` reason. Since `Close` was not a separate message, the `ConnectionClosed(...)` error was used to indicate that the WebSocket connection should be closed (according to RFC); this error **did not** mean that we received close message, but rather that we've already processed it and did all required actions to prepare for the gracefull connection close.

So after https://github.com/snapview/tokio-tungstenite/issues/51#issuecomment-454260255 and https://github.com/snapview/tokio-tungstenite/issues/45#issue-342153089 we decided to make `Close` a part of the `Message` enum, which means that the user is able to push the `Close` message to the `Sink` like a normal message and initiate a closing by doing that; also the one can now **receive** a `Close` as a normal message and treat it appropriately (if applicable).

This now has its implications: since the `Close` is part of the `Message` enum and is returned to the user when this message is received, there is no reason to include it in `ConnectionClosed`. The overall logic did not change, i.e. you still have to wait for `ConnectionClosed` to be able to gracefully close the connection etc. `ConnectionClosed` is still a good indicator that the connection can safely be dropped.

To be honest, I don't quite understand why it turned out to be more complex than expected to do this change in `warp`, theoretically this should be a small change which barely touches the actual logic, i.e. if `ConnectionClosed` was used to indicate the `Async::Ready(None)` it still should be used for this purpose. And `Close` can be treated like a normal message in the wrapper. If there are some tests which crash after the change, it might be an indicator that something is wrong in some different parts.

seanmonstar:
@bbangert you've done a lot with websockets and tried them in warp, would you mind sharing any thoughts on whether your server would like to the close code?

bbangert:
I've never needed the close code, but it was useful to know if the client gracefully closed the connection vs. a connection drop that was unintentional. Mainly for metrics, but if a ton of clients go from graceful close to sudden drop, it could signal that we shipped a bad client, so that has definitely been handy.

We haven't had a need for the close code thus far, but I could imagine a client library choosing to send different close codes based on the conditions that caused it and that being useful to the server.

Eijebong:
Anything I can do to get this moving again ?

application-developer-DA:
Well... IMHO we have to return the `Message::Close` to the user (and it seems to be trivial to do as everything has already been written for it; AFAICT the problem is only with some unit test which assumes that no `Message::Close` will be received).

But even if you guys have a strong opinion that we don't have to return the `Message::Close` to the user, then [regardless of the implementation] it's not correct to tell that the stream is ended when `Message::Close` has been received (see my comments above), the correct place to return` Async::Ready(None)` is at `ConnectionClosed` (as it had always been done in the previous versions).

seanmonstar:
I don't use websockets often enough, so I'm likely wrong. Let's do it the way others have suggested.

Eijebong:
Ok, now I think it's good to go.

application-developer-DA:
Looks correct to me.

(I did not check in details the changes in tests though)

seanmonstar:
Thanks for all the follow-through! <3

Eijebong:
Sure :) When can we expect this to be published ?

seanmonstar:
In a few minutes, merging in a couple other PRs and then publishing.

PR #232, opened 2019-05-30
Fix doc link to reject module

boxdot:
Also expose re-export of `http` crate in the docs. Rationale: http crate is fundemantal for using the APIs of warp. Examples are using it as re-export, therefore I think it should be marked as such in the docs. It also helps not to confuse users of warp to add `http` to their Cargo.toml.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,7 +86,6 @@ extern crate bytes;
 #[macro_use]
 extern crate futures;
 extern crate headers;
-#[doc(hidden)]
 pub extern crate http;
 extern crate hyper;
 #[macro_use]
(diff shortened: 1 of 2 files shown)


seanmonstar:
Hey, sorry I missed this! Thanks for noticing the broken link. Let's start with that, I'd so far prefer to keep `http` hidden from the main docs page, so as to not distract.

seanmonstar:
Actually, I just manually merged the first commit to master, so you don't need to do anything else. Thanks again!

boxdot:
Thanks for merging! About `http` -- either way has good reasons I guess, so no hard feelings. :)

PR #236, opened 2019-06-03
Unseal `Reply` trait.

### LINKED ISSUE #233: Consider to unseal `reply::Reply` trait
Currenty, `warp::reply::Reply` trait is sealed and is implemented only for 
 - `http::StatusCode`
 - `http::Response<impl Into<hyper::Body>>`
 - `String`
 - `&'static str`

### Why this is a limitation?  

If we could implement `Reply` for custom types, we would be able to offload more of the API definition onto the type system. That is, we could express entirely in the type system what a handler might return. E.g.

* `-> Json<SomeSerializableType>`
* `-> Result<(), StatusBadRequestUnit>`
* `-> Result<Json<User>, Either<StatusBadRequestUnit, StatusNotFoundUnit>>`

This is great for writing and reading APIs. More important, it allows to easily test such handlers. Currently, we have either to split handlers into several functions, or to inspect the value of returned response. Further, it also allow to generate an API spec like OpenAPI from the handler signature. I also think this follows the warp philosophy where an app is defined by a single composed Rust type.

### Example

I unsealed the `Reply` trait and implemented it for `()`, `Result<T,E>` with `T, E: Serializable`, and I also exposed a `Json<T>` wrapper. I change the todos example to show how the API might look. Feel free to take a look at the diff: https://github.com/seanmonstar/warp/compare/master...boxdot:unseal-reply

What do you think?

boxdot:
Closes #233

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -33,9 +33,11 @@
 //!     .or(warp::post2().and(custom));
 //! ```
 
+use std::borrow::Cow;
 use std::error::Error as StdError;
 use std::fmt;
 
+use generic::{Either, One};
 use http::header::{HeaderName, HeaderValue, CONTENT_TYPE};
 use http::{HttpTryFrom, StatusCode};
 use hyper::Body;
@@ -44,9 +46,12 @@ use serde_json;
 
 use reject::Reject;
 // This re-export just looks weird in docs...
-pub(crate) use self::sealed::{ReplyHttpError, ReplySealed, Reply_, Response};
+pub(crate) use self::sealed::Reply_;
 #[doc(hidden)]
 pub use filters::reply as with;
+
+/// Response type into which types implementing the `Reply` trait are convertable.
+pub type Response = ::http::Response<Body>;
 
 /// Returns an empty `Reply` with status code `200 OK`.
 ///
@@ -109,7 +114,7 @@ struct Json {
     inner: Result<Vec<u8>, ()>,
 }
 
-impl ReplySealed for Json {
+impl Reply for Json {
     #[inline]
     fn into_response(self) -> Response {
         match self.inner {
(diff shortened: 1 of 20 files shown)


src/reply.rs
```diff
@@ -44,10 +46,13 @@ use serde_json;
 
 use reject::Reject;
 // This re-export just looks weird in docs...
-pub(crate) use self::sealed::{ReplyHttpError, ReplySealed, Reply_, Response};
+pub(crate) use self::sealed::Reply_;
 #[doc(hidden)]
 pub use filters::reply as with;
 
+/// Response type into which types implementing the `Reply` trait are convertable.
+pub type Response = ::http::Response<Body>;
```
seanmonstar: Oh yea, I had forgotten that part of the reason the trait was sealed was to not expose the exact body type... oh well, not a big deal for 0.1.

src/reply.rs
```diff
@@ -193,15 +198,43 @@ where
...
+/// ```rust
+/// use warp::{Filter, http::Response};
+///
+/// struct Message {
+///     msg: String
+/// }
+///
+/// impl warp::Reply for Message {
+///     fn into_response(self) -> warp::reply::Response {
+///         Response::builder()
```
seanmonstar: This can skip the builder and unwrap: `Response::new(format!("message: {}", self.msg).into())`.

src/reply.rs
boxdot: Ah, indeed. So, you have a plan for 0.2 to change this? E.g. by using an associated type?

boxdot:
Thanks for reviewing! 

The CI is grumpy because of a nightly warning about trait objects without an explicit `dyn`. I fixed this also. However, this requires rustc version 1.27. I guess this is fine since you wanted to move to the 2018 edition anyway.

boxdot:
Ah, I see you already fixed the dyn warnings. Nice! :)

seanmonstar:
Thank you!

PR #244, opened 2019-06-24
Make `Json` public (fixes #243)

### LINKED ISSUE #243: Make `Json` type `pub`lic
I was trying to create some generic conversion, so that each my function which returns a `Future` or a `Result` (i.e. `IntoFuture`) can be converted to a `Future` which is "compatible" with the warp service I'm running, so that each `Err` is converted to the corresponding `Rejection` and each `Ok` value is serialized into a JSON value.

In other words, I tried to turn `T where T: IntoFuture<Item = impl Serialize, Error = impl Into<Cause>>` into `impl IntoFuture<Item = Reply, Error = Rejection>`.

Unfortunately my attempt to do so failed (I did it by defining my own extension trait and implementing it for each `T` which implements already mentioned traits. However, the `warp::reply::json()` function returns a `-> impl Reply` which is actually a `Json` type which is not public, thus I could not specify the return datatype in my trait).

If `Json` were public, I could have written something like this:

```
trait IntoServiceFuture {
    fn into_service_future(self) -> Box<dyn Future<Item = reply::Json, Error = Rejection>>;
}

impl<T, R, E> IntoServiceFuture for T
where
    T: IntoFuture<Item = R, Error = E> + 'static,
    R: Serialize,
    E: Into<Cause>,
{
    fn into_service_future(self) -> Box<dyn Future<Item = reply::Json, Error = Rejection>> {
        Box::new(
            self.into_future()
                .map(|res| reply::json(&res))
                .map_err(|err| reject::custom(err)),
        )
    }
}
```

But if the `reply::Json` is not public, I can't write it like that. And I can't use the `impl` inside the trait definition as it's not yet supported in Rust (I think it has something to do with existential types).

Probably it also makes sense to remove `doc(hidden)` from some types such as `Cause` and other things.

P.S.: I know that a decent documentation is still missing, but maybe it would make sense to add it as well. I think `warp` is really a very nice framework and I like it much more than `actix` or similar crates on a backend; `warp` just feels to be much more flexible, and much more "**Rust**y" if you know what I mean :)

application-developer-DA:


application-developer-DA:
Just a friendly reminder that this small PR awaits the validation ;)

This seems to be a minor thing, but that would facilitate implementation of things mentioned in https://github.com/seanmonstar/warp/issues/243 without the need to use a forked version.

jxs:
thanks! 👍

PR #249, opened 2019-07-04
Expose more basic internal HTTP error types

gferon:
I'm mostly looking for feedback, this is more a proof of concept rather than the actual solution.

I'm trying to render rejections as a JSON object. For custom errors, it already works totally fine. However, in order to reuse all the existing `Filter` implementations, while being able to capture the actual reason in a custom `recover` function, I felt it was needed to define a few of the existing error types as `pub`.

Btw, everything's awesome, thanks for `warp` :cyclone:

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -78,6 +78,30 @@ pub fn not_found() -> Rejection {
     }
 }
 
+// 400 Bad Request
+#[inline]
+pub(crate) fn invalid_query() -> Rejection {
+    known(InvalidQuery(()))
+}
+
+// 400 Bad Request
+#[inline]
+pub(crate) fn missing_header(name: &'static str) -> Rejection {
+    known(MissingHeader(name))
+}
+
+// 400 Bad Request
+#[inline]
+pub(crate) fn invalid_header(name: &'static str) -> Rejection {
+    known(InvalidHeader(name))
+}
+
+// 400 Bad Request
+#[inline]
+pub(crate) fn missing_cookie(name: &'static str) -> Rejection {
+    known(MissingCookie(name))
+}
+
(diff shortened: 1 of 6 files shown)


seanmonstar:
Ah, I see what you want to do. You want to check for specific causes to customize your JSON. That's fair!

I'm for the idea in general, with two thoughts:

1. Where should the individual types live? What seems better, `warp::cookie::MissingCookie` or `warp::reject::MissingCookie`?

2. Any public types should not expose their fields. So like, `LengthRequired` has no fields, which if exposed, means we couldn't add any fields later without it being a breaking change. So I'd change them to have a private unit field, like `LengthRequired(())`.

Whatcha think?

gferon:
I have to admit I have a preference for `warp::reject::MissingCookie` but it's probably just a matter of taste in the end.

I don't have much experience with backwards compatibility and Rust, but it sounds like a sensible thing to do - I'll try it out and commit something right away.

seanmonstar:
I think I also prefer them all being in `warp::reject`... If you want to move them, we can do it now. But it's also easy to move later.

gferon:
Great, I'll take care of it and report back.

gferon:
I think I'm done for this one.

src/reject.rs
```diff
@@ -505,8 +529,25 @@ impl fmt::Debug for Rejections {
     }
 }
 
+/// Invalid query
 #[derive(Debug)]
-struct MethodNotAllowed;
+struct InvalidQuery(());
```
seanmonstar: Missing a `pub`?

gferon: Yep, sorry. I just fixed it.

PR #251, opened 2019-07-05
Remove 'static lifetime restriction from path function

omrerdn:
Removed `'static` lifetime restriction from the parameter of [path function](https://docs.rs/warp/0.1.16/warp/filters/path/index.html), now it accepts `&str` with an arbitrary lifetime.

_`'static `restriction might be added intentionally by the api designer, if it is then this PR is useless._

According to [this question ](https://stackoverflow.com/questions/56895824/static-lifetime-for-a-temporary-parameter) in SO, one might want this kind of dynamic usability.

-  Generating input on the fly. 

It can be done with the generic lifetime parameter.

### DIFF
--- a/src/filters/path.rs
+++ b/src/filters/path.rs
@@ -153,7 +153,7 @@ use route::Route;
 /// let hello = warp::path("hello")
 ///     .map(|| "Hello, World!");
 /// ```
-pub fn path(p: &'static str) -> impl Filter<Extract = (), Error = Rejection> + Copy {
+pub fn path<'a>(p: &'a str) -> impl Filter<Extract = (), Error = Rejection> + Copy +'a {
     assert!(!p.is_empty(), "exact path segments should not be empty");
     assert!(
         !p.contains('/'),


shepmaster:
This doesn't appear to work; I'd suggest adding the test that you used locally to verify the functionality to this PR. That way, the functionality will be automatically tested in the future and shows off how to use it.

I edited `example/hello.rs`:

```rust
extern crate warp;

use warp::Filter;

fn main() {
    let hello = warp::path(&String::from("tmp")).map(|| "hello world");

    warp::serve(hello).run(([127, 0, 0, 1], 8080));
}
```

But get

```none
error[E0716]: temporary value dropped while borrowed
 --> examples/hello.rs:6:29
  |
6 |     let hello = warp::path(&String::from("tmp")).map(|| "hello world");
  |                 ------------^^^^^^^^^^^^^^^^^^^-                      - temporary value is freed at the end of this statement
  |                 |           |
  |                 |           creates a temporary which is freed while still in use
  |                 argument requires that borrow lasts for `'static`
```

Issue #257, opened 2019-08-04
Usability of the WebSocket filter


najamelan:
I'm implementing a lib that provides AsyncRead/AsyncWrite on top of websockets, to enable codec based communication between wasm code running in the browser and a server.

The wasm part is [published](https://crates.io/crates/ws_stream_wasm), but the server part turns out to be more of a challenge. Since applications with a web frontend almost always have to serve static files, it makes sense to write this library as to be compatible with other crates like hyper and warp. This makes it possible for people to upgrade http connections benefiting from the convenience, not having to handle TLS separately and so on.

So it would seem the most appropriate input for my library would be to take in an object that implements `Stream<Result<tungstenite::Message, tungstenite::Error>>` and `Sink<tungstenite::Message, Error=tungstenite::Error>` and return an object that implements `AsyncRead`/`AsyncWrite`. So I had a look at the warp ws module to see how to go about providing compatibility with warp and I'm running into some blockers:

1. warp does not return the tungstenite types like tokio-tungstenite. This means I'm obliged to create my own type which implements `From<warp::ws::Message>` and `From<warp::Error>`. And creating similar conversion from tungstenite types to provide compatibility with raw tungstenite or tokio-tungstenite connections. In itself this is not convenient, but it's not the end of the world, it can be done. Is it important for warp to not expose dependency types in the public API? If not, could we at least allow access to the inner type, or provide `Into<tungstenite::Message> for warp::ws::Message` so that we can access the inner type?

2. `warp::ws::Message` does not provide the `CloseFrame` from tungstenite, making it impossible for the client code to detect that the websocket client disconnects cleanly and read the close code and reason.

[...]

PR #258, opened 2019-08-05
Allow extracting the binary data out of a ws::Message without copying.

najamelan:
Until now the API only allowed getting a reference. For code wanting to work on websocket connections, regardless of whether they're provided by warp, tungstenite, tokio-tungstenite or other websocket libraries this is a problem because the type from warp has to be kept all the way through algorithms to avoid making a needless copy of all the data.

This adds a new method to `warp::ws::Message` called `into_bytes` which will return a `Vec<u8>` of the inner data.

Note that this alleviates this problem for code wanting to operate on binary messages. However, there could and probably should also be a a similar method to extract string messages without copying.

I chose to delegate the call to the inner tungstenite message because I feel tungstenite has [better implementations](https://github.com/snapview/tungstenite-rs/blob/master/src/protocol/message.rs#L261) of these conversions, rather than modifying the `as_bytes` method which can panic. Following a similar implementation as `as_bytes` is a possible alternative. Note that tungstenite will also transform data from message types other than `String` and `Binary`. I figure that if the user wants to avoid that, they have to check the type, which
they would have to do in the alternative impl as well, since otherwise it can panic.

The second commit adds an impl for `Into<Vec<u8>>` for `warp::ws::Message`.

This could be done for string conversions too, but since those are fallible and tungstenite has their own error type, we would just need to convert the error type.

This pull request deals with very specific problem but there are others:
- there is an inconsistency between the `to_str` method which returns an error when called on a message variant for which the operation is not supported and the `as_bytes` method which will panic. Note also the inconsistency in naming.
- the `is_pong` method is missing. Unless there is a specific reason for this, it should probably be added for consistency in case people want to wrap `ws::Message`, eg. to abstract out with differences from other websocket libraries.
- there is currently no way to recover the code and reason of a close message.

See also: #257

All in all, the tungstenite `Message` is superior in both functionality and implementation. I would be in favor of an `Into` implementation or simply an `into_inner` method. I understand that this means that changing from tungstenite to another websocket implementation becomes a breaking change. However at this early stage in development, avoiding this at all costs leads to a degraded API while breaking changes will be necessary in the future anyway to fix it. I understand the desire hide implementation details, but I would propose to postpone that until a stable fully featured and clean API can be offered as an alternative.

If it's desirable to fix some of the other issues mentioned above, I can add more commits to this pull request or file another one. 

There are no unit tests provided with this pull request.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -400,6 +400,11 @@ impl Message {
             _ => unreachable!(),
         }
     }
+
+    /// Destructure this message into binary data.
+    pub fn into_bytes(self) -> Vec<u8> {
+        self.inner.into_data()
+    }
 }
 
 impl fmt::Debug for Message {
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -412,3 +412,10 @@ impl fmt::Debug for Message {
         fmt::Debug::fmt(&self.inner, f)
     }
 }
+
+
+impl Into<Vec<u8>> for Message {
+    fn into(self) -> Vec<u8> {
+        self.into_bytes()
+    }
+}


robeady:
I came across this same issue but for string messages - what do you think of also implementing a function like `into_string` returning an `Option<String>` if the message is text?

najamelan:
I wouldn't mind, but I'm waiting for some reaction from a maintainer to see how they want to deal with this. As soon as there is some green light, I'm happy to rebase this, and add String extraction, np

PR #260, opened 2019-08-06
Update tungstenite and serde_urlencoded

SimonSapin:
This helps avoid duplicating crates in https://bugzilla.mozilla.org/show_bug.cgi?id=1568540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+### v0.1.19
+
+- **Fixes**:
+  - Update some private dependencies.
+
 ### v0.1.18 (July 25, 2019)
 
 - **Features**:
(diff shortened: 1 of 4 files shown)


seanmonstar:
(Just a note that I'm on vacation without my computer for 7 more days, I can handle this then.)

SimonSapin:
Could you publish this on crates.io? Thanks!

SimonSapin:
Actually, it turns out that the addition of the `multipart` crate in https://github.com/seanmonstar/warp/commit/70c45006932ff2d80521c3b3c03c1592379d65f0 causes more duplications since it depends of old versions of `mime` and `mime_guess`, but *that* update is non-obvious: https://github.com/abonander/multipart/pull/121.

What do you think of making `multipart` optional?

PR #262, opened 2019-08-09
Add query parameters to the todos example.

rockwotj:
Previously, none of the examples used query parameters.

### DIFF
--- a/examples/todos.rs
+++ b/examples/todos.rs
@@ -61,6 +61,7 @@ fn main() {
     // (and to reject huge payloads)...
     let json_body = warp::body::content_length_limit(1024 * 16).and(warp::body::json());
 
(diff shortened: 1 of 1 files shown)


examples/todos.rs
```diff
@@ -61,11 +61,14 @@ fn main() {
     // (and to reject huge payloads)...
     let json_body = warp::body::content_length_limit(1024 * 16).and(warp::body::json());
 
+    let list_options = warp::filters::query::query::<ListOptions>();
```
seanmonstar:
```suggestion
    let list_options = warp::query::<ListOptions>();
```

rockwotj:
Good suggestions! I've updated the comments in various places.

PR #263, opened 2019-08-09
Fix depreciation warning from mime_guess create

rockwotj:
See: https://docs.rs/mime_guess/2.0.1/mime_guess/fn.guess_mime_type.html


seanmonstar:
Thanks!

PR #264, opened 2019-08-16
Make multipart and websocket support optional, but supported by default

SimonSapin:


SimonSapin:
This avoids depending (through `multipart`) on old versions of `mime`, `mime_guess`, and `log` which duplicate with more recent versions that are also used. Per https://github.com/abonander/multipart/pull/121 it looks like `multipart` wants to support an old version of rustc, and so may not be open to updating its dependencies.

seanmonstar:
I'm good with this! Thanks!

SimonSapin:
Nice! Could you publish on crates.io?

seanmonstar:
@SimonSapin sure! While I'm getting ready, there's nothing else that needs de-duping, right?

SimonSapin:
I tried it with `[patch.crates-io]` on mozilla-central and I think this is it, thanks!

seanmonstar:
Published!

abonander:
@SimonSapin I don't care so much about old Rust support but I figured breaking on Debian-stretch might be an issue for Servo/Firefox.

SimonSapin:
Firefox generally bumps its Rust version requirement to the latest Stable two weeks after that release: https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox

PR #265, opened 2019-08-26
update to std::future::Future

jxs:


dbcfd:
async tokio-tungstenite for websocket support https://github.com/snapview/tokio-tungstenite/pull/68

src/filter/or.rs
```diff
@@ -58,50 +60,37 @@ where
...
-            Ok(Async::NotReady) => {
-                self.state = State::Second(Some(err1), second);
-                Ok(Async::NotReady)
-            }
-            Err(e) => {
-                self.original_path_index.reset_path();
-                return Err(e.combine(err1));
-            }
+            pin.original_path_index.reset_path();
+            pin.state = State::Second(Some(err1), fut2);
```
zimond: This loop seems to be breaking `tests/path.rs#or`

jxs: yeah, that one is solved thanks
some more to go 😀

zimond:
There is a way to bring the websocket module back I think... [tokio-tungstenite](https://github.com/snapview/tokio-tungstenite/blob/bea935e16f4e2aa0f895ac1e81c082f15811b272/src/compat.rs)

bstrie:
We literally *just* switched to warp to take advantage of its websocket support, so I'd love if that could stay first-class in the long run. :)

Cargo.toml
```diff
@@ -11,25 +11,29 @@ categories = ["web-programming::http-server"]
...
 bytes = "0.4.8"
-futures = "0.1"
+futures-preview = "0.3.0-alpha.18"
+futures-util-preview = "0.3.0-alpha.18"
 headers = "0.2"
 http = "0.1"
-hyper = "0.12.18"
-log = "0.4"
+hyper = "=0.13.0-alpha.1"
+logcrate = { version = "0.4", package = "log" }
```
cynecx: This may be a dump question, but why did you alias the log crate?

dbcfd:
> We literally _just_ switched to warp to take advantage of its websocket support, so I'd love if that could stay first-class in the long run. :)

The tokio-tungstenite branch I put above seems to be working. Would be great to pull it into this branch and do some more testing on it. I do need to add more tests to it.

examples/errors.rs
```diff
@@ -34,49 +30,53 @@ impl Display for Error {
...
 impl StdError for Error {}
 
-fn main() {
+#[warp::tokio_rt]
+async fn main() {
     let hello = warp::path::end().map(warp::reply);
 
     let oops =
-        warp::path("oops").and_then(|| Err::<StatusCode, _>(warp::reject::custom(Error::Oops)));
+        warp::path("oops").and_then(|| async { Err::<StatusCode, _>(warp::reject::custom(Error::Oops))});
```
dbcfd: Rather than doing an async closure, use `futures::future::ready`

examples/errors.rs
```diff
@@ -34,49 +30,53 @@ impl Display for Error {
...
+                message: "oops, you aren't allowed to use this method.".into(),
+            });
+            Ok(warp::reply::with_status(json, code))
+        } else {
+            // Could be a NOT_FOUND, or any other internal error... here we just
+            // let warp use its default rendering.
+            Err(err)
+        }
+    };
+    async { err }
```
dbcfd: Same as above, use `futures::future::ready` so we're not paying the generator cost.

src/filter/map_err.rs
```diff
@@ -38,11 +42,11 @@ where
...
 {
-    type Item = T::Extract;
-    type Error = E;
+    type Output = Result<T::Extract, E>;
 
     #[inline]
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        self.extract.poll().map_err(|err| (self.callback)(err))
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let extract = &mut get_unchecked!(self).extract;
```
dbcfd: Possibly use pin project to get a pinned field to poll?

Cargo.toml
```diff
@@ -11,13 +11,18 @@ categories = ["web-programming::http-server"]
...
 
 [dependencies]
 bytes = "0.4.8"
-futures = "0.1"
+futures-preview = "0.3.0-alpha.18"
+futures-util-preview = "0.3.0-alpha.18"
 headers = "0.2"
 http = "0.1"
-hyper = "0.12.18"
+hyper = "=0.13.0-alpha.1"
```
zimond: hyper alpha 2 is released. I think here we can upgrade this dependency and just add some modifications in server code to adapt to the new version https://github.com/zimond/warp/commit/139a5abafdc23ff169af566f54147a38486a1fc8

examples/errors.rs
jxs: thanks! had missed that

Cargo.toml
jxs: thanks! had to also implement `Accept` due to `AddrIncoming` no longer implementing Stream

src/server.rs
```diff
@@ -217,10 +231,17 @@ where
...
+    ) -> Result<(SocketAddr, impl Future<Output = ()> + 'static), hyper::error::Error>
+    {
         let addr = addr.into();
         let (addr, srv) = try_bind!(self, &addr)?;
-        Ok((addr, srv.map_err(|e| error!("server error: {}", e))))
+        let srv = srv.map(|result| {
+            if let Err(err) = result {
+                log::error!("server error: {}", err)
+            }
+        });
```
udoprog: This hides the error on errors, while previously we used to return a `Result` with `Err = ()`. That's because old futures are equivalent to `Future<Output = Result<T, E>>`. With this, the server future is replaced with one that immediately yields when failing to bind which is probably not what we want.

src/server.rs
```diff
@@ -259,12 +280,16 @@ where
...
+    ) -> (SocketAddr, impl Future<Output = ()> + 'static) {
         let (addr, srv) = bind!(self, addr);
         let fut = srv
             .with_graceful_shutdown(signal)
-            .map_err(|e| error!("server error: {}", e));
+            .map(|result| {
+                if let Err(err) = result {
+                    log::error!("server error: {}", err)
+                }
+            });
```
udoprog: Same here.

src/server.rs
udoprog: I made a mistake. Binding is still synchronous through hyper and isn't affected by this changed!

Cargo.toml
```diff
@@ -26,10 +31,9 @@ scoped-tls = "1.0"
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.6"
-tokio = "0.1.11"
-tokio-io = "0.1"
-rustls = { version = "0.15", optional = true }
-tokio-threadpool = "0.1.7"
+tokio = "0.2.0-alpha.5"
+tokio-executor = "=0.2.0-alpha.5"
```
udoprog: time to bump tokio: `0.2.0-alpha.6` :)

udoprog:
I've rebased and bumped dependencies here because I need it:
https://github.com/udoprog/warp

I accidentally ran `cargo fmt` while working so I guess we have that now as well 🤷‍♂ . Would be nice if we could make that consistent in the codebase (@seanmonstar ?).

lpil:
Great stuff! Is there anything we can do to help get this over the line?

src/filters/sse.rs
```diff
@@ -40,24 +42,27 @@ use std::error::Error as StdError;
...
-use filter::One;
-use reply::Response;
-use {Filter, Rejection, Reply};
+use crate::filter::One;
+use crate::reply::Response;
+use crate::{Filter, Rejection, Reply};
 
 /// Server-sent event message
-pub trait ServerSentEvent: SseFormat + Sized + Send + 'static {
+pub trait ServerSentEvent: SseFormat + Sized + Send + Sync + 'static {
```
NeoLegends: Is there a reason for the `Sync` bound? Pre `std::Future` SSE replys did not have to be `Sync`.

seanmonstar:
This is super exciting, as I've said before. I've given @jxs merge rights, so when it's ready and green, we can merge :D

NeoLegends:
So the only requirement for now is waiting until async await hits stable?

udoprog:
We're two weeks out from a stable async/await. Building on beta is a very strong indicator that it will work by then, so I personally wouldn't mind merging to master.

examples/sse_chat.rs
```diff
@@ -111,19 +105,16 @@ fn user_connected(
 
     // When `drx` will dropped then `dtx` will be canceled.
     // We can track it to make sure when the user leaves chat.
-    warp::spawn(poll_fn(move || dtx.poll_cancel()).map(move |_| {
+    warp::spawn(async move {
+        dtx.closed().await;
+        drx.close();
```
benceszigeti: That usage hold the `oneshot::channel` open forever, so the `fn user_disconnected(...)` not reachable.

bstrie:
> The tokio-tungstenite branch I put above seems to be working. Would be great to pull it into this branch and do some more testing on it. I do need to add more tests to it.

@dbcfd Took a minute, but I see that https://github.com/snapview/tokio-tungstenite/pull/68 has finally landed. :) You were suggesting that we replace Warp's usage of tungstenite with tokio-tungstenite in order to fully complete the async transition, yes?

jxs:
I am going to open a separate issue, só it's easy to track

dbcfd:
@bstrie Yeah, I would recommend using https://github.com/snapview/tokio-tungstenite/pull/68. We encountered a few issues during the upgrade around context and waking, but they have been worked through, and it should be failure solid now. Hopefully easy to integrate as well, since the previous warp implementation was close to the tokio-tungstenite implementation.

PR #270, opened 2019-09-17
Adds example in docs with cors `allow_any_origin()`

FedericoPonzi:
Adds an example of usage of `allow_any_origin()` in the docs.


seanmonstar:
Thanks!

PR #272, opened 2019-09-21
Enable setting max_message_size on Ws2

emschwartz:


seanmonstar:
Thanks!

gakonst:
@seanmonstar could you perhaps do a point release for this feature?

Issue #32, opened 2018-08-07
Routing for many endpoints compiles very slowly


kiljacken:
When writing a router for an API with a lot of endpoints, compilation crawls to a halt.

The code in question can be found here: http://gist.github.com/kiljacken/866f6617b2998d672ac8f67cb6037186

When compiling with `cargo build` and only one route enabled the compilation takes around ~5s. With all routes enabled this increases to 1m 12s. This obviously makes iteration time quite high, and thus decreases productivity quite a bit.

It would be nice if there was more effective way to do this.

PR #273, opened 2019-09-21
Filter::debug_boxed()

emschwartz:
Compiling many composed filters can be very slow. Boxing speeds up compile times but slows down the program at runtime. This adds a method to box the Filter only for debug builds while leaving it unchanged for release builds.

Related to https://github.com/seanmonstar/warp/issues/32

### DIFF
--- a/examples/returning.rs
+++ b/examples/returning.rs
@@ -15,6 +15,13 @@ pub fn index_filter() -> impl Filter<Extract = (&'static str,), Error = Rejectio
     warp::path::end().map(|| "Index page")
 }
 
+// Option 3: box only for debug builds
+// In debug builds, this will return a BoxedFilter like Option 1
+// but in release builds it will return the filter as in Option 2.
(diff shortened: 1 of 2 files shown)


emschwartz:
It might be a bit strange to have a method whose return type depends on what type of build it is. Happy to reconsider PR this but I thought it shouldn't cause too many problems because most of the uses of warp filters seems to either deal with `impl Filter...` or `BoxedFilter`s.

seanmonstar:
I appreciate the goal, but I do think it could be confusing if a user were to write out the type `BoxedFilter` and then watch it break...

*Maybe* if this returned some `DebugBoxedFilter` that wasn't publicly nameable, and thus a user could only write `impl Filter`, this might be better? Though, it might just be a shortcut you keep in a personal project, also :)

emschwartz:
Happy to go either way. I'll close this for now but reopen it if you think including the `DebugBoxedFilter` is a good idea

PR #277, opened 2019-09-27
Update README.md

### LINKED ISSUE #276: Improve Readme
It can be hard to know what to put in a Readme, so I thought you might appreciate these things that I immediately wondered but it didn't answer:

1. Does it work on stable?
2. What is its concurrency model? Thread pool? Async/await? (**Edit**: Actually it does say "Asynchronous", but maybe you could expand on that..., e.g. what executor does it use?)
3. How do you deal with global shared state, e.g. an in-memory database (this always seems to be the trickiest part of Rust web frameworks)?

Also, but less critically:

1. Does it support LetsEncrypt easily (like Caddy)?

Timmmm:
Something like this? By the way, I'm still a Rust noob but why does the `Arc<Vec<u32>>` in the docs example not need a `Mutex<>`?

Fixes #276

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,10 @@ Since it builds on top of [hyper](https://hyper.rs), you automatically get:
 
 - HTTP/1
 - HTTP/2
-- Asynchronous
+- Asynchronous (using Tokio 0.1 by default)
 - One of the fastest HTTP implementations
 - Tested and **correct**
+- Works on stable Rust
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
The doc example wasn't trying to assume too much, maybe you prefill the Vec at startup and don't need to mutate. Or maybe the state in question already has synchronization (not a Vec), and so a mutex wouldn't be needed at all (maybe the thing has one internally, or uses lockfree structures).

Timmmm:
Ah ok fair enough!

seanmonstar:
Thanks for writing this up, but I'm going to leave the README as-is for now.

PR #285, opened 2019-10-08
Better document two foot-guns

emschwartz:
While switching Interledger.rs to `warp`, we made two subtle and hard-to-debug mistakes:
- failing to include the `warp::path::end()` filter and thus having a shorter route like `/math` accidentally match a request for a longer path like `/math/sum/:u32/:u32`
- putting auth checks before the path-related filters and winding up with auth-related errors when a request for a completely different route had an invalid body

This attempts to document those two issues in the rustdoc and examples, though there are probably some other places this type of gotcha could be documented.

This also includes a README for the `examples` directory to point newcomers to the examples in a specific order.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,37 @@
+# Examples
+
+Welcome to the examples! These show off `warp`'s functionality and explain how to use it.
+
+## Getting Started
+
+- [`hello.rs`](./hello.rs) - Just a basic "Hello World" API
+- [`routing.rs`](./routing.rs) - Builds up a more complex set of routes and shows how to combine filters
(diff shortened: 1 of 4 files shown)


emschwartz:
In case it's of interest, this commit was the one that fixed the 2nd issue so you can see the before and after: https://github.com/interledger-rs/interledger-rs/pull/324/commits/8400a254ba7f06a4b529dfff92361e32ca81f997

seanmonstar:
Thanks, these are great improvements! I've been torn about the situation of `path::end`. It's easy to miss, like you noticed. But I also didn't want it to be weird to slowly build up a path. Maybe the `path!` macro can make it better somehow?

Issue #255, opened 2019-07-19
custom error for path param


hwchen:
Hello,

When parsing the param where I have a custom type, e.g. `path!(api / MyParam)`, it seems like any error in parsing will result in `Rejection(NotFound)`. Is it possible for the parsing error to be passed down so that I can deal with it in `.recover()`?

Thanks!

seanmonstar:
There exists [`warp::path::param2`](https://docs.rs/warp/0.1.17/warp/filters/path/fn.param2.html) currently, but I think it needs some tweaks. But instead of boring you with those without asking, you could try `warp::path(api).and(warp::path::param2::<MyParam>())`. The error will be in the `Rejection::cause()`.

hwchen:
Awesome, worked perfectly.

I'm curious what you think the tweaks need to be.

seanmonstar:
- A better name, since I don't expect it to *replace* `param`.
- It probably shouldn't return a `NOT_FOUND` rejection with the error as a cause, but just the custom error as the direct rejection.

hwchen:
I’ve also had a hard time coming up with a good name :) . I guess it should somehow convey `param_with_explicit_err`, but shorter.

PR #293, opened 2019-10-28
Param filter with user-defined rejection

hwchen:
updates `filters::param2`:
- changes name to `param_with_err`. This still doesn't quite feel right to me, so let me know if there's other suggestions.
- returns `Rejection` directly from `FromStr` `Err`, instead converting the `Err` to a `Cause` and always returning a `not found` `Rejection`.
- created a fuller example in the docstring.

Let me know if this is the idea you had from #255.


hwchen:
Playing around with this a bit, I'm not sure it makes sense to have users implement their own `Err` -> `Rejection` if `Cause`s are going to be the main way of identifying errors in a `recover`. Also, from notes in the api, it looks like you're moving towards using `custom` rejections more generally.

So, perhaps it makes more sense to have `param_with_err` just return a `custom` rejection, instead of having the user define an extra impl that's not really that powerful.

seanmonstar:
> I'm not sure it makes sense to have users implement their own Err -> Rejection if Causes are going to be the main way of identifying errors in a recover.

I'm not quite sure what you mean by this. However, relatedly, master has just seen a refactor of the rejection system which would affect this (#311).

hwchen:
Ok, I'll look at the new rejection system.

hwchen:
Ok, I'll look at the new rejection system.

hwchen:
And it seems like you no longer need to implement `std::error::Error` to work with the custom rejection system, just `reject::Reject`.

hwchen:
It's been a while since this PR was submitted, I've updated it to better reflect the latest `Rejection` system.

Quick summary:

`path::param` will always return `reject::not_found()` if there's an error in parsing. `path::param_with_reject` will allow the user to specify a custom rejection in case of parsing failure.

I see that `param2`, which served this function previously (I think), was removed. Let me know if there's no longer interest and I can close this PR.

jxs:
Hi, and sorry for the delay! So if I understand correctly what you are trying to achieve is being able to replace a `Known` rejection with a custom one right? If so can't you do that with a `recover` handler and [`find`](https://docs.rs/warp/0.2.5/warp/reject/struct.Rejection.html#method.find) and return the response you want on that case

Issue #297, opened 2019-11-05
Expose `Never` in public API.


dylanede:
Currently functions returning `impl Filter<Extract = ..., Error = warp::never::Never>` cannot be written by the user, preventing them from returning irrefutable filters from functions.

jxs:
have you tried using `futures::never::Never`? there's an example which uses it :https://github.com/seanmonstar/warp/blob/master/examples/sse.rs#L3
I think the reason warp uses its own `Never` type is because Infalible doesn't implement `std::error::Error`. 
hmm but since warp implements `Reject` for `Never` `crate::never::Never`can maybe be replaced with `futures::never::Never`

taiki-e:
> I think the reason warp uses its own `Never` type is because Infalible doesn't implement `std::error::Error`.

Infalible implements `std::error::Error` via [`std::string::ParseError`](https://doc.rust-lang.org/nightly/std/string/type.ParseError.html).
https://github.com/rust-lang-nursery/futures-rs/pull/1681#discussion_r303720764

jxs:
ah nice then, thanks @taiki-e! i can submit a PR to address this

dylanede:
@jxs that PR looks like it should solve the problem. Regarding your first suggestion, I can't use a different `Never` type from warp in the return signature of my function because I am returning the result of `warp::any().map(...)`, which has the error type `warp::never::Never`.

jxs:
resolved with #298

PR #298, opened 2019-11-05
switch crate::never::Never with futures::never::Never

jxs:
address #297


seanmonstar:
I'd hoped we could replace the type with `!` when it stabilizes in 2037 ;_;

seanmonstar:
Oh, just realized `futures::never::Never` is a type alias for `std::convert::Infallible`, which itself is planned to be a type alias for `!`.

Ok, I'm fine with this, but let's just point at `std::convert::Infallible` directly.

jxs:
ok, updated it!

PR #306, opened 2019-11-07
Fix very slow concat implementation

karlri:
I love warp but when I tested multipart functionality, it turned out to be very slow for large files. I pinned it down to the body::concat() implementation and made the smallest possible change I could come up with to fix it. It is now magnitudes faster. Maybe this is not the best possible implementation, but it is certainly faster than the original.

Test with a 10mb file containing something useful to firefox shows 50x improvement: 5s vs 100ms.

➜  warp-test git:(master) ✗ time curl -F file=@/home/karlri/bin/firefox/omni.ja http://127.0.0.1:3030/file
0.02s user 0.00s system 0% cpu 4.990 total

➜  warp-test git:(master) ✗ time curl -F file=@/home/karlri/bin/firefox/omni.ja http://127.0.0.1:3030/file
0.00s user 0.01s system 16% cpu 0.097 total

Cheers,
Karl

### DIFF
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -105,8 +105,19 @@ pub fn stream() -> impl Filter<Extract = (BodyStream,), Error = Rejection> + Cop
 ///     });
 /// ```
 pub fn concat() -> impl Filter<Extract = (FullBody,), Error = Rejection> + Copy {
-    body().and_then(|body: ::hyper::Body| Concat {
-        fut: body.try_concat(),
+    stream().and_then(|mut body_stream: BodyStream| async move {
+        use bytes::{BytesMut, Bytes};
+        use futures::stream::*;
+        let mut bytes = BytesMut::new();
+        loop {
+            match body_stream.next().await {
+                Some(Ok(data)) => bytes.extend_from_slice(data.bytes()),
+                Some(Err(err)) => return Err(reject::known(BodyReadError(err))),
+                None => break,
+            }
+        }
+        let chunk = Chunk::from(Bytes::from(bytes));
+        Ok::<_, Rejection>(FullBody { chunk })
     })
 }
 
@@ -277,7 +288,7 @@ pub struct BodyStream {
 }
 
 impl Stream for BodyStream {
-    type Item = Result<StreamBuf, crate::Error>;
+    type Item = Result<StreamBuf, hyper::error::Error>;
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         let opt_item: Option<Result<Chunk, hyper::Error>> = ready!(Pin::new(&mut self.get_mut().body).poll_next(cx));
@@ -286,7 +297,6 @@ impl Stream for BodyStream {
             None =>  Poll::Ready(None),
             Some(item) => {
                 let stream_buf = item
-                    .map_err(|e| crate::Error::from(crate::error::Kind::Hyper(e)))
                     .map(|chunk| StreamBuf { chunk });
 
                 Poll::Ready(Some(stream_buf))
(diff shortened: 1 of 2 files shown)


karlri:
I am willing to clean up the implementation or do it differently if you could provide some guidance. I am also considering implementing multipart upload not with concat, but instead with this: https://github.com/abonander/multipart-async by the same author. That would allow us to support uploads streaming asynchronously directly to files.

src/filters/body.rs
```diff
@@ -105,8 +105,19 @@ pub fn stream() -> impl Filter<Extract = (BodyStream,), Error = Rejection> + Cop
...
 pub fn concat() -> impl Filter<Extract = (FullBody,), Error = Rejection> + Copy {
-    body().and_then(|body: ::hyper::Body| Concat {
-        fut: body.try_concat(),
+    stream().and_then(|mut body_stream: BodyStream| async move {
+        use bytes::{BytesMut, Bytes};
+        use futures::stream::*;
+        let mut bytes = BytesMut::new();
+        loop {
+            match body_stream.next().await {
+                Some(Ok(data)) => bytes.extend_from_slice(data.bytes()),
```
seanmonstar: Oh yea huh, I think `Bytes::extend(Bytes)` can't use specialization like `Vec` can, and so it's much slower. :cry:

src/filters/body.rs
```diff
@@ -277,7 +288,7 @@ pub struct BodyStream {
 }
 
 impl Stream for BodyStream {
-    type Item = Result<StreamBuf, crate::Error>;
+    type Item = Result<StreamBuf, hyper::error::Error>;
```
seanmonstar: Mind reverting this?

src/filters/fs.rs
```diff
@@ -63,6 +63,13 @@ pub fn file(path: impl Into<PathBuf>) -> impl FilterClone<Extract = One<File>, E
         })
 }
 
+
+/// useful if file path needs to be looked up in database etc.
+/// use conditionals() filter to get a Conditionals, then reply with this function
+pub async fn serve_file_reply(path: impl Into<PathBuf>, conditionals: Conditionals) -> Result<File, Rejection> {
```
seanmonstar: This could be useful, but I'd rather discuss the proper API in a new issue, yea?

src/filters/body.rs
karlri: True, Ideally, I would like to read the content length header, if any, and allocate a Vec::with_capacity, and then try to re-use that allocation somehow in the FullBody. What would be the best way to get the content length?

src/filters/body.rs
karlri: Absolutely, however, on line 115, to construct the concat error:
`Some(Err(err)) => return Err(reject::known(BodyReadError(err))),`
the BodyReadError variant needs to contain hyper::Error specifically.

Possible ways I can think of to resolve this:
1. match on Kind::Hyper and assert that it is that variant on line 115
2. make stream() return BodyStream with error type mapped from hyper::error::Error to warp::Error
3. rename BodyStream to BodyStreamInner make new BodyStream that delegates poll and maps error

Some other idea? :)

src/filters/fs.rs
karlri: Absolutely, since I'm used to gerrit I didn't realize that this would end up in the same PR :/

src/filters/body.rs
karlri: I realized I'm being stupid. There is the Body type which implements Stream and is exactly what I was looking for. I pushed a new branch to karlri:concat-performance with this fixed and the rest of the changes reverted, but I didn't figure out how to change the PR to merge that branch.

src/filters/body.rs
karlri: I figured out how to do it, ingenious filter chain architecture :) available on on karlri:concat-performance

src/filters/fs.rs
karlri: Done on karlri:concat-performance

seanmonstar:
Hm, since this PR is from the master branch, I don't think you can change it to another. You could close this and open a new PR from your new branch.

karlri:
Ok, will do!

PR #315, opened 2019-11-13
Make todos example use tokio::sync::Mutex instead of std::sync::Mutex.

Migi:
`std::sync::Mutex` is not supposed to be used inside async functions, because it blocks the executor when it locks. Both tokio and async-std have their own versions of `Mutex` that don't block but instead have a `lock()` function that returns a future.

I hope that this honest mistake isn't a foreshadowing of a fear I've had for quite some time now, that because tons of common functions in the rust standard library and ecosystem are blocking, and because this blocking is often quite invisible, both in the code and in tests and even in production under low traffic load, that this kind of mistake is going to happen all the time in many projects and libraries. With as a result that everything works well until you actually get a higher traffic load and then performance falls off a cliff. Debugging these issues in a complex application sounds hard (because it's only an issue under high load), detecting that this kind of blocking is happening in the first place sounds even harder.

This is why I think we should be extremely vigilant about avoiding all kinds of blocking in async functions, *before* they become a real issue. Not just in the library code but also in examples, in case people copy them and build on them without understanding the issue. Hence this PR.

### DIFF
--- a/examples/todos.rs
+++ b/examples/todos.rs
@@ -1,7 +1,8 @@
 #![deny(warnings)]
 
 use std::env;
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
+use tokio::sync::Mutex;
 use warp::{http::StatusCode, Filter};
 use serde_derive::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


jxs:
nice catch! 
thanks

PR #316, opened 2019-11-14
Fix very slow concat implementation 2

karlri:
Big multipart upload is now magnitudes faster

Change-Id: I5857cf07358ad89d7af16afabba5006565ee60f8

### DIFF
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -2,14 +2,15 @@
 //!
 //! Filters that extract a body for a route.
 
+use std::convert::TryInto;
 use std::error::Error as StdError;
 use std::fmt;
+use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
-use std::future::Future;
 
 use bytes::Buf;
-use futures::{future, ready, TryFuture, TryStreamExt, Stream};
+use futures::{future, ready, Stream, TryFuture};
 use headers::ContentLength;
 use http::header::CONTENT_TYPE;
 use hyper::{Body, Chunk};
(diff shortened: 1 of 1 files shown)


karlri:
New pull request due to me being inexperienced with github, sorry :)

Fixed:
1. reverted change of error type of stream()
2. reverted other unrelated commit
3. pre-allocate buffer length based on content-length header to avoid realloc. Could make it easy for someone to make server allocate all memory, but then, shouldn't a max content length filter be used?
4. use Vec instead of BytesMut as buffer
5. general cleanup

I hope it could be of use :)
Br
Karl

-

src/filters/body.rs
```diff
@@ -105,9 +106,30 @@ pub fn stream() -> impl Filter<Extract = (BodyStream,), Error = Rejection> + Cop
...
+    body().and(crate::filters::header::optional2()).and_then(
+        |mut chunk_stream: Body, cl_header: Option<ContentLength>| {
+            async move {
+                // allocate a buffer to hold the full body
+                let mut buf = Vec::with_capacity(match cl_header {
+                    Some(ContentLength(len)) => match len.try_into() {
+                        Ok(length) => length, // fits in usize
+                        Err(_) => return Err(reject::payload_too_large()),
+                    },
+                    _ => 1024, // reasonable?
```
seanmonstar: Hm, if there's no chunks, this would allocate a `Vec` that ws never needed. Perhaps assume `0` if we don't know otherwise.

karlri: I guess, if content-length is set, but length does not match, then it is a protocol violation and we should return 400 Bad Request and close the connection even if keep-alive. The real question is: is that handled elsewhere? I suppose that should be handled at a central point.

jbg: [RFC 7230 Section 3.3.3](https://tools.ietf.org/html/rfc7230#section-3.3.3) has guidance:

Regarding `Content-Length` header value greater than actual received bytes:

> If a valid Content-Length header field is present without Transfer-Encoding, its decimal value defines the expected message body length in octets.  If the sender closes the connection or the recipient times out before the indicated number of octets are received, the recipient MUST consider the message to be incomplete and close the connection.

Regarding actual received bytes greater than `Content-Length` header value:
> If the final response to the last request on a connection has been completely received and there remains additional data to read, a user agent MAY discard the remaining data or attempt to determine if that data belongs as part of the prior response body, which might be the case if the prior message's Content-Length value is incorrect.  A client MUST NOT process, cache, or forward such extra data as a separate response, since such behavior would be vulnerable to cache poisoning.

seanmonstar:
Master now makes use of `hyper::body::{aggregate, to_bytes}` instead of `try_concat`, which should have better performance.

PR #318, opened 2019-11-17
Add `warp::tls::peer_certificates()` filter.

### LINKED ISSUE #227: Add filter to be able to get peer's certificate on TLS server
It would be nice to have a filter (or add this feature by any other means), so that the we can check peer's certificate (i.e. [something like this](https://docs.rs/openssl/0.10.22/openssl/ssl/struct.SslRef.html#method.peer_certificate) when running the server in TLS mode. That would be very beneficial for the HTTPS servers analyze peer certificate and use for some sort of authentication and/or identity check.

jxs:
Returns the tls certificate chain from the connection, if the transport
type is using tls and has certificates
closes #227
there's an issue with this though, if we want to address #56 implementing `Service<T>` gets worse due to having  to pass `Request`, `remote_address`, and `tls_peer_certificates` to `Service::call()`

### DIFF
--- a/src/filters/tls.rs
+++ b/src/filters/tls.rs
@@ -1,0 +1,24 @@
+//! Tls filters
+
+use std::convert::Infallible;
+
+use crate::filter::{filter_fn_one, Filter};
+
+/// Creates a `Filter` to get the remote tls certificate chain of the connection.
+///
+/// If the underlying connection doesn't have certificates, this will yield
+/// `None`.
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+///
+/// let route = warp::::tls::peer_certificates()
+///     .map(|certs: Option<Vec<Vec<u8>>>| {
+///         println!("peer certificates = {:?}", certs);
+///     });
+/// ```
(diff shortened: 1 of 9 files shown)


src/filters/tls.rs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// ```
+/// use warp::Filter;
+///
+/// let route = warp::::tls::peer_certificates()
+///     .map(|certs: Option<Vec<Vec<u8>>>| {
+///         println!("peer certificates = {:?}", certs);
+///     });
+/// ```
+pub fn peer_certificates() -> impl Filter<Extract = (Option<Vec<Vec<u8>>>,), Error = Infallible> + Copy {
```
seanmonstar: I wonder if this should be a stronger type. Maybe `Vec<Certificate>`, with `struct Certificate(Vec<u8>)`?

jxs: yeah I recall now you mentioned it on the issue, Should it also have methods to return it's inner, like `get_ref` or `into_inner`?

jxs:
this approach would not work because of https://github.com/hyperium/hyper/issues/2175

PR #320, opened 2019-11-21
fix infinite recursion of path! macro, closes #221

### LINKED ISSUE #221: Fix infinite recursion of `path!` macro
Since the `path!` macro has a catch-all case of `$($piece:tt)*`, anything that doesn't match one of the other cases ends up infinitely expanding the catch-all. This can be fixed by using a separate `path_internal` macro that doesn't have the catch-all. However, using this macro internally can break people who were using `#[macro_use(path)]`, so slotted for 0.2.

jxs:


src/filters/path.rs
```diff
@@ -485,10 +485,19 @@ fn path_and_query(route: &Route) -> PathAndQuery {
 /// that the shorter path filter does not match the longer paths.
 #[macro_export]
 macro_rules! path {
-    (@start $first:tt $(/ $tail:tt)*) => ({
-        let __p = path!(@segment $first);
+        ($($pieces:tt)*) => ({
+            use $crate::internal_path;
```
seanmonstar: The use isn't needed, right?

src/filters/path.rs
```diff
@@ -485,10 +485,19 @@ fn path_and_query(route: &Route) -> PathAndQuery {
...
+            $crate::internal_path!(@start $($pieces)*)
+        });
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! internal_path {
```
seanmonstar: To follow the pattern, let's add a comment that it's not a public API, and prefix with double underscores.

src/filters/path.rs
jxs: yeah, we just need to prefix all the calls with `$crate::`

jxs:
ok, updated!

PR #321, opened 2019-11-22
add warp::temporary_redirect

hamptokr:
This just adds `warp::temporary_redirect` (307) alongside `warp::redirect` (301). Found this utility function helpful and figured others might need it too. 

In my case I was implementing oauth login and needed code to run in the handler before every redirect. A permanent redirect (301) gets cached in most browsers so `warp::redirect` wouldn't work because the server would never even get called on subsequent login attempts.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,7 +145,7 @@ pub use self::filters::{
 #[doc(hidden)]
 pub use self::filters::ws::ws;
 #[doc(hidden)]
-pub use self::redirect::{redirect, temporary_redirect};
+pub use self::redirect::redirect;
 #[doc(hidden)]
 #[allow(deprecated)]
 pub use self::reject::{reject, Rejection};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,16 +145,16 @@ pub use self::filters::{
 #[doc(hidden)]
 pub use self::filters::ws::ws;
 #[doc(hidden)]
-pub use self::redirect::redirect;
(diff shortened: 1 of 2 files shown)


src/redirect.rs
```diff
@@ -28,6 +28,27 @@ pub fn redirect(uri: impl AsLocation) -> impl Reply {
...
+/// ```
+/// use warp::{http::Uri, Filter};
+///
+/// let route = warp::path("v1")
+///     .map(|| {
+///         warp::temporary_redirect(Uri::from_static("/v2"))
+///     });
+/// ```
+///
+pub fn temporary_redirect(uri: impl AsLocation) -> impl Reply {
```
seanmonstar: How about just name this `temporary`, and then it's used like `warp::redirect::temporary(uri)`.

src/lib.rs
```diff
@@ -145,16 +145,16 @@ pub use self::filters::{
 #[doc(hidden)]
 pub use self::filters::ws::ws;
 #[doc(hidden)]
-pub use self::redirect::redirect;
+pub use self::redirect::{redirect, temporary_redirect};
```
seanmonstar: I'd just let the more specialized functions exist only in the `redirect` module.

src/redirect.rs
hamptokr: I like that! I will make the change.

seanmonstar:
Thanks!

Issue #222, opened 2019-05-09
Consider removal of `warp::ext::set`


seanmonstar:
It takes advantage of warp's usage of thread-local storage, but that means we can't change away from internally using TLS as long as this function requires it.

remexre:
What would the replacement API look like?

seanmonstar:
I'm not certain. Perhaps by adding a way to turn a `Filter` into a `Service`, extensions could be set before reaching the filters. Or a new `warp::ext::with_mut(impl Fn(&mut Extensions))` could be added.

carllerche:
Makes sense :+1:

itsfarseen:
How to set an extension in latest version of warp?

paulzhang5511:
@itsfarseen 
```rust
fn with_db(pool: Pool) -> impl Filter<Extract = (Pool,), Error = Infallible> + Clone {
    warp::any().map(move || pool.clone())
}

let user_create = warp::path!("api" / "user" / "create")
        .and(warp::post())
        .and(warp::body::json())
        .and(with_db(pool.clone()))
        .and_then(user::user_create);
```

PR #322, opened 2019-11-26
[Draft] tower_service::Service <-> warp::Filter interop

jxs:
I continued the work started on #70 to try and make `tower_service::Service`'s run on warp and the ability to run `warp::Filter`'s as a service on Hyper.
### Notes
 
- to be able to call `tower_service::Service::call(&mut self, req: Request)` I had to remove the `Arc` usage and `Clone` to the trait bounds, I benchmarked and results seemed similar benchmarking the hello world example with `--release`:

(with Arc)
```
$ wrk -t4 -c1000 -d30s --latency http://127.0.0.1:3030
Running 30s test @ http://127.0.0.1:3030
  4 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.34ms    3.79ms  82.56ms   87.93%
    Req/Sec    12.71k     3.47k   22.53k    65.86%
  Latency Distribution
     50%    3.19ms
     75%    5.16ms
     90%    8.69ms
     99%   18.67ms
  1519131 requests in 30.06s, 188.34MB read
  Socket errors: connect 751, read 114, write 2, timeout 0
Requests/sec:  50530.67
Transfer/sec:      6.26MB
```
(with Clone)
```
$ wrk -t4 -c1000 -d30s --latency http://127.0.0.1:3030
Running 30s test @ http://127.0.0.1:3030
  4 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.31ms    2.95ms  41.62ms   81.27%
    Req/Sec    12.68k     3.62k   22.61k    60.70%
  Latency Distribution
     50%    3.57ms
     75%    5.20ms
     90%    8.03ms
     99%   14.95ms
  1517745 requests in 30.08s, 188.17MB read
  Socket errors: connect 751, read 102, write 42, timeout 0
Requests/sec:  50463.45
Transfer/sec:      6.26MB
```

- I looked into #222 and since the `WarpService` trait already existed and is also passing the `remote_addr` extra argument, so I impled `WarpService` for T: `tower_service::Service `, when a `Service` is called via warp `remote_addr`(or other future arguments's, see https://github.com/seanmonstar/warp/pull/318) is discarded

- impl `WarpService` for T: `tower_service::Service` requires `tower_service::Service::Error : warp::Reject`  that seemed to be the simplest way to then integrate Service errors with warp's logging and error handling, should another error be used/should `Reject` be impled for T? 

- I think there still needs to be developed a way to compose `warp::Filter`'s with `tower_service::Service`'s, but maybe that could be done after closing the interop API?

### DIFF
--- a/examples/tower_service.rs
+++ b/examples/tower_service.rs
@@ -1,0 +1,40 @@
+extern crate warp;
+extern crate hyper;
+
+use std::task::{Context, Poll};
+use std::pin::Pin;
+use std::future::Future;
+use warp::{Response, Request};
+use warp::reject::Reject;
+use tower_service::Service;
+
+#[derive(Clone)]
+struct TowerService;
+
+#[derive(Debug)]
+struct ServiceError;
+impl Reject for ServiceError {}
+
+impl Service<Request> for TowerService {
+    type Response = Response;
+    type Error = ServiceError;
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, _req: Request) -> Self::Future {
+        // Create the HTTP response
+        let resp = Response::new("hello world".into());
+
+        // Return the response as an immediate future
+        Box::pin(futures::future::ok(resp))
+    }
+}
+
+#[tokio::main]
+async fn main() {
+
+    warp::serve_service(TowerService).run(([127, 0, 0, 1], 3030)).await;
+}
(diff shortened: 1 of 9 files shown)


examples/tower_service.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use std::task::{Context, Poll};
+use std::pin::Pin;
+use std::future::Future;
+use warp::{Response, Request};
+use warp::reject::Reject;
+use tower_service::Service;
+
+#[derive(Clone)]
+struct TowerService;
```
LucioFranco: You could use service_fn as an example here

seanmonstar:
Thanks for the draft, I used what you had here in #367.

jxs:
awesome! do you plan to impl Filter for Service on a following PR?

seanmonstar:
I'm not so sure there needs to be a conversion the other way around, one can `svc.call` inside a `filter.and_then` if need be.

PR #332, opened 2019-12-08
Adds example for Unix sockets. Resolves #282.

### LINKED ISSUE #282: How to use Unix sockets with warp? (documentation request)
How can I use Unix sockets with Warp?  From the [relevant section of the documentation](https://docs.rs/warp/0.1.20/warp/struct.Server.html), I see that I should be able to use the `use_incoming` method, which is described like this:
> Run this Server forever on the current thread with a specific stream of incoming connections.
> This can be used for Unix Domain Sockets, or TLS, etc.

But, for some reason, I cannot figure out how to do so.  I have attempted to pass `use_incoming` a `std::os::unix::net::UnixListener` or a `std::os::unix::net::UnixStream`; neither worked.  I also tried passing it a String with the address of the socket, which also didn't work.  Can anyone point me in the right direction?

If anyone can help me figure this out, I'd be happy to submit a documentation PR with an example so that others aren't similarly confused in the future.

remexre:
Once we're on Tokio 0.2.0, this example also needs the `uds` feature on Tokio.

### DIFF
--- a/examples/unix_socket.rs
+++ b/examples/unix_socket.rs
@@ -1,0 +1,13 @@
+#![deny(warnings)]
+
+use tokio::net::UnixListener;
+
+#[tokio::main]
+async fn main() {
+    pretty_env_logger::init();
+
+    let incoming = UnixListener::bind("/tmp/warp.sock").unwrap().incoming();
+    warp::serve(warp::fs::dir("examples/dir"))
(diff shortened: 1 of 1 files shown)


jxs:
nice! thanks :fireworks:

PR #336, opened 2019-12-11
Update to hyper 0.13/tokio 0.2

kornholi:
- [x] [headers](https://github.com/hyperium/headers) released with http 0.2

I got rid of a few `'static` bounds from server.rs to get the `unix_socket` example builduing, but it seems like most of the others are unnecessary as well?

### DIFF
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -107,7 +107,7 @@ fn path_from_tail(
             future::ready(sanitize_path(base.as_ref(), tail.as_str()))
                 .and_then(|mut buf| async {
                     let clone_buf = buf.clone();
-                    let is_dir = tokio_executor::blocking::run(move || clone_buf.is_dir()).await;
+                    let is_dir = tokio::task::spawn_blocking(move || clone_buf.is_dir()).await.expect("FIXME");
                     if is_dir {
                         log::debug!("dir: appending index.html to directory path");
                         buf.push("index.html");
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -106,8 +106,7 @@ fn path_from_tail(
         .and_then(move |tail: crate::path::Tail| {
             future::ready(sanitize_path(base.as_ref(), tail.as_str()))
                 .and_then(|mut buf| async {
-                    let clone_buf = buf.clone();
-                    let is_dir = tokio::task::spawn_blocking(move || clone_buf.is_dir()).await.expect("FIXME");
+                    let is_dir = tokio::task::block_in_place(|| buf.is_dir());
                     if is_dir {
                         log::debug!("dir: appending index.html to directory path");
                         buf.push("index.html");
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -85,12 +85,6 @@ pub fn file(path: impl Into<PathBuf>) -> impl FilterClone<Extract = One<File>, E
 /// // - `GET /static/app.js` would serve the file `/www/static/app.js`
 /// // - `GET /static/css/app.css` would serve the file `/www/static/css/app.css`
 /// ```
-///
-/// # Note
-///
-/// This filter uses `tokio-fs` to serve files, which requires the server
-/// to be run in the threadpool runtime. This is only important to remember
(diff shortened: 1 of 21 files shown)


src/filters/fs.rs
```diff
@@ -107,7 +107,7 @@ fn path_from_tail(
             future::ready(sanitize_path(base.as_ref(), tail.as_str()))
                 .and_then(|mut buf| async {
                     let clone_buf = buf.clone();
-                    let is_dir = tokio_executor::blocking::run(move || clone_buf.is_dir()).await;
+                    let is_dir = tokio::task::spawn_blocking(move || clone_buf.is_dir()).await.expect("FIXME");
```
kornholi: should this get its own rejection?

jxs: I think you can [block_in_place](https://docs.rs/tokio/0.2.0/tokio/task/fn.block_in_place.html) as tokio is doing the same within `tokio::fs::` :

src/test.rs
```diff
@@ -93,8 +93,8 @@ use futures::{future, FutureExt, TryFutureExt};
 #[cfg(feature = "websocket")]
 use futures::{SinkExt, StreamExt};
 #[cfg(feature = "websocket")]
-use tokio::{
-    sync::{mpsc, oneshot},
+use futures::{
+    channel::{mpsc, oneshot},
```
jxs: what was the rational for this change?

kornholi: tokio's `UnboundedSender` doesn't implement `Sink` anymore and I couldn't quickly figure out how to rewrite this: https://github.com/seanmonstar/warp/pull/336/files/1786ff96aeb1d5bbb58ac93eae9594c4e6cd1489..6c95083a2783050622fcefa12e1ef0801f8f3e51#diff-c5fc4beef0f03493c2e9a9c876073f0fR514-R523

I kept running into weird errors where `read` wasn't a stream/future anymore...?

jxs: ah, yeah you are right, it's simpler to just switch to std one

kornholi:
ptal @seanmonstar

src/filters/fs.rs
```diff
@@ -106,8 +106,7 @@ fn path_from_tail(
         .and_then(move |tail: crate::path::Tail| {
             future::ready(sanitize_path(base.as_ref(), tail.as_str()))
                 .and_then(|mut buf| async {
-                    let clone_buf = buf.clone();
-                    let is_dir = tokio_executor::blocking::run(move || clone_buf.is_dir()).await;
+                    let is_dir = tokio::task::block_in_place(|| buf.is_dir());
```
seanmonstar: I think this *requires* the multi-threaded runtime to be running, whereas it could be written like this:

```rust
let is_dir = tokio::fs::metadata(buf).await.map(|m| m.is_dir()).unwrap_or(false);
```

tests/fs.rs
```diff
@@ -20,52 +20,62 @@ async fn file() {
     assert_eq!(res.body(), &*contents);
 }
 
-#[tokio::test]
+#[tokio::test(threaded_scheduler)]
```
seanmonstar: With the change to `is_dir` above, these can use the regular scheduler, right?

src/filters/fs.rs
jxs: sorry, it was me who suggested `block_in_place` in https://github.com/seanmonstar/warp/pull/336#discussion_r356627991 as i saw fs filter already requires the multi threaded runtime

seanmonstar: It definitely *did* in 0.1, but tokio 0.2 can now allow fs operations even with the basic scheduler.

jxs: ah nice! fs filter doc can also be updated then, as it no longer needs to reference that: https://docs.rs/warp/0.1.10/warp/filters/fs/fn.dir.html#note

kornholi: cool! this is what I wanted but didn't know existed.

kornholi:
@seanmonstar is there anything left to do here?

seanmonstar:
Oh sorry, I didn't get any notification of the force push :D

PR #355, opened 2019-12-29
Remove panic branch in `Message::as_bytes`.

derekdreery:
Currently  using `as_bytes` panics in some cases. This patch handles all cases, by returning an option that is `Some` if the message contains a payload, and `None` if not.

Alternatively, this PR could be changed so that `Ping` and `Pong` messages also returned `None`, if preferred.

This requires changing the signature from `&[u8]` to `Option<&[u8]>`.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -439,12 +439,14 @@ impl Message {
         }
     }
 
-    /// Return the bytes of this message.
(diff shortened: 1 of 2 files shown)


seanmonstar:
I suppose an alternative is to just return an empty slice in those cases.

PR #360, opened 2020-01-03
Add filters::ext::get_optional

AtsukiTak:
I love the warp and its concept. Thanks much for creating this framework!

This PR adds `filters::ext::get_optional` function. That is same as

```rust
filters::ext::get().map(|t| Some(t)).or_else(|_| None)
```

but more convenient and straightforward :)

### DIFF
--- a/src/filters/ext.rs
+++ b/src/filters/ext.rs
@@ -1,7 +1,7 @@
 //! Request Extensions
 
 use futures::future;
-use std::error::Error as StdError;
+use std::{convert::Infallible, error::Error as StdError};
 
 use crate::filter::{filter_fn_one, Filter};
 use crate::reject::{self, Rejection};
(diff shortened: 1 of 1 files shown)


src/filters/ext.rs
```diff
@@ -21,6 +21,17 @@ pub fn get<T: Clone + Send + Sync + 'static>(
     })
 }
 
+/// Get a previously set extension of the current route.
+///
+/// If the extension doesn't exist, it yields `None`.
+pub fn get_optional<T: Clone + Send + Sync + 'static>(
+) -> impl Filter<Extract = (Option<T>,), Error = Rejection> + Copy {
```
seanmonstar: Since this never rejects, it can change to `Error = Infallible`.

AtsukiTak: Thanks. I'll fix it.

seanmonstar:
Thank you! I changed the name from `ext::get_optional` to just `ext::optional` to match the other optional filters, and merged into master with https://github.com/seanmonstar/warp/commit/c1235e37882a130256bf0191f6d1dea66c65a39b.

PR #371, opened 2020-01-10
Change `warp::path` to work with any `AsRef<str>` argument.

### LINKED ISSUE #31: warp::path requires &'static str
Hi,

I'm attempting to use this crate in a project where some of the paths to route aren't known until runtime, as they're acquired from the CLI via `clap`. Unfortunately, this appears to mean that I cannot use `warp` at all as `warp::path` requires `p` to be `&'static str`.

I'm wondering if `warp` will ever support the use case where it's simply given a `&str` to route.

Thanks.

seanmonstar:
Closes #31

### DIFF
--- a/src/filters/path.rs
+++ b/src/filters/path.rs
@@ -132,9 +132,10 @@ use std::str::FromStr;
 use futures::future;
 use http::uri::PathAndQuery;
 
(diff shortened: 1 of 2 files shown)


phyber:
Excellent, thanks!

Issue #307, opened 2019-11-11
Rejection and anyhow


Geobert:
Hi,

I'm trying to use `anyhow` but as warp::reject::Rejection doesn't impl std::error::Error, they doesn't compose well.

Am I missing something or it just can't be done (yet)?

jxs:
Hi @Geobert, it should be the other way around, wrapping your `anyhow`or any implementation of `std::error::Error` in [warp::reject::custom](https://github.com/seanmonstar/warp/blob/master/examples/errors.rs#L38)

Geobert:
I see, so we can't use `anyhow::Result` but have to use the standard `Result` to return a `Rejection`.

jxs:
yea, 
you can't use `anyhow::Result` because it is a [type alias](url) where the Error is `anyhow::Error` and you need to return `Result<T, warp::Rejection>`

udoprog:
Might want to re-open this. It doesn't seem to be possible since #311

`warp::reject::custom` only accept types which implements `Reject`, but `anyhow::Error` doesn't implement it:
```
89  | ...                   .map_err(warp::reject::custom)
    |                                ^^^^^^^^^^^^^^^^^^^^ the trait `warp::reject::Reject` is not implemented for `anyhow::Error`
```
This seems to mean it's only possible to use custom rejections with types defined locally in your crate (Lack of blanket impl?). This is what I use as a workaround:
```rust
#[derive(Debug)]
struct CustomReject(anyhow::Error);

impl warp::reject::Reject for CustomReject {}

pub(crate) fn custom_reject(error: impl Into<anyhow::Error>) -> warp::Rejection {
    warp::reject::custom(CustomReject(error.into()))
}
```

thomaseizinger:
We are using `HttpApiProblem` (RFC7807), and excellent standard for returning errors, in our appliction together with warp and anyhow. Maybe it helps in this case: https://github.com/comit-network/comit-rs/blob/master/cnd/src/http_api/problem.rs

aslamplr:
I have tried to follow what @thomaseizinger has suggested. It is a really clean approach.
Unfortunately, I am getting the following error.

```
the trait bound `http_api_problem::HttpApiProblem: warp::reject::Reject` is not satisfied
the trait `warp::reject::Reject` is not implemented for `http_api_problem::HttpApiProblem`
```

I have put together sample code to recreate the issue:

```rust
use warp::Filter;

#[tokio::main]
async fn main() {
[...]

PR #374, opened 2020-01-14
impl Reject for T: Display + Debug

jxs:
- addresses #307 
- allows types which `impl std::error::Error` to not need `impl Reject for T {}` 
- allow one to just return:
```Rust
warp::reject::custom("quick rejection, TODO: improve into a type")
```
so that anyhow! dependency becomes unnecessary

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -145,6 +145,8 @@ fn __reject_custom_compilefail() {}
 // would be double-boxing it, and the downcasting wouldn't work as expected.
 pub trait Reject: fmt::Debug + Sized + Send + Sync + 'static {}
 
+impl<T: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static> Reject for T {}
+
 trait Cause: fmt::Debug + Send + Sync + 'static {
     fn as_any(&self) -> &dyn Any;
 }


seanmonstar:
When I redesigned the rejection system, one of my goals was to make rejection explicit. To make someone decide if the route should be *rejected*, and thus other routes should be tried, or if their error should be converted into a reply.

jxs:
oh, yeah I just realized reading [reject.rs:L128](https://github.com/seanmonstar/warp/blob/master/src/reject.rs#L128) that it was assumed :+1:

udoprog:
@seanmonstar What initially raised this was #307, I'm curious what you'd recommend to support rejections then for types foreign to the crate (like with `anyhow::Error`). Would that be using newtypes (or some other crate-local rejection type) [as per my comment](https://github.com/seanmonstar/warp/issues/307#issuecomment-570833388)?

seanmonstar:
@udoprog if the filter should *reject* and try another filter, then sure. You can define some type that you'd expect to look for later in a `recover`.

thomaseizinger:
`HttpApiProblem` is getting support for `warp` as soon as 0.2 is released: https://github.com/chridou/http-api-problem/pull/17

udoprog:
> @udoprog if the filter should _reject_ and try another filter, then sure. You can define some type that you'd expect to look for later in a `recover`.

I'll be referencing the `errors` example below:
https://github.com/seanmonstar/warp/blob/master/examples/errors.rs#L42

_My_ goal with rejections have never been try another filter. The way I've implemented error handling in my project is to recover immediately at the end of each unique filter to limit the number of retries. The `errors` example above strictly speaking _does_ try another filter since `recover` is used after the inner filter. But error customization also seems to be its primary use since the paths do not match (`nope` != `oops`). Reading your response and other issues it seems like error customization is a secondary goal of rejecting?

Some dedicated mechanism for propagating errors down to an error handler without trying other filters might be worth considering (if one doesn't already exist?). To my experience that is currently the dominant use case for custom reject/recover. Two other people I've talked to about error handling in warp have been tripped by this so far.

udoprog:
@seanmonstar I'd be curious on your take of the above?

Issue #377, opened 2020-01-15
How to close ws with a custom status code?


NickeZ:
The [CloseEvent](https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent) of the websocket have a range of application specific close codes (4000-4999). Is it possible to close the connection with code from that range?

NickeZ:
Something like this works as a proof of concept. I verified that I can read out the `code` attribute of the close event on the client side.

```diff
diff --git a/src/filters/ws.rs b/src/filters/ws.rs
index 4b3554b..0d7e49c 100644
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -411,6 +411,18 @@ impl Message {
         }
     }
 
+    /// Construct a new Close `Message`.
+    pub fn close<V: Into<u16>, S: Into<std::borrow::Cow<'static, str>>>(
+        v: Option<(V, S)>,
+    ) -> Message {
+        Message {
+            inner: protocol::Message::Close(v.map(|(v, s)| protocol::CloseFrame {
+                code: protocol::frame::coding::CloseCode::from(v.into()),
+                reason: s.into(),
+            })),
+        }
+    }
+
     /// Returns true if this message is a Text message.
     pub fn is_text(&self) -> bool {
         self.inner.is_text()
```

seanmonstar:
I'd be fine with add a `Message::close` constructor.

Is there a need to have such a complex signature? What situations is it trying to provide for? (Sorry, I'm less familiar with websocket close messages).

NickeZ:
I wouldn't say that I'm an expert either. But you can provide a close code (u16) and a reason (str). I'm also not an expert in rust interfaces so maybe my suggestion is strange.

Is there a reason to not just expose the complete inner tungstenite `Message`? It seems to be a maintenance burden to reimplement the whole type in this crate.

seanmonstar:
The reason for not exposing the inner `tungstenite` is to allow for a) upgrading to new versions, or b) changing implementations, without it being a breaking change in the API.

How about this:

- `Message::close()` uses the default.
- `Message::close_with(code: impl Into<u16>, reason: impl Into<Cow<'static, str>>)` for adding details.

NickeZ:
OK, I'll make a PR

PR #382, opened 2020-01-16
Allow new Message type `Close` to be created

NickeZ:
Fixes https://github.com/seanmonstar/warp/issues/377


seanmonstar:
Thanks!

PR #390, opened 2020-01-19
Add protobuf support with PROST!

ParkMyCar:
This PR adds protobuf support to warp via the [PROST!](https://github.com/danburkert/prost) crate.

I implemented a protobuf Filter by adding `pub fn protobuf(...)` to `filters/body.rs`. Its usage is exactly the same as the json Filter, except it uses the `decode(...)` function provided by the `trait prost::Message`.

Similarly I implemented the `trait Reply` for a `struct Protobuf` so Filters can return a protobuf encoded response.

All of this is shown in a new `examples/protobuf.rs`, which includes an example cURL command to test encoding, sending, receiving, then decoding protobufs.

@seanmonstar thanks for building this awesome framework :)


Cargo.toml
```diff
@@ -43,11 +45,13 @@ pretty_env_logger = "0.3"
...
 handlebars = "1.0.0"
 tokio = { version = "0.2", features = ["macros"] }
+prost-build = "0.6.1"
 
 [features]
 default = ["multipart", "websocket"]
-websocket = ["tungstenite"]
+protobuf = ["prost"]
 tls = ["rustls"]
+websocket = ["tungstenite"]
```
ParkMyCar: This was just a reordering to make the features alphabetical

src/filters/body.rs
```diff
@@ -255,6 +259,43 @@ pub fn form<T: DeserializeOwned + Send>() -> impl Filter<Extract = (T,), Error =
...
+pub fn protobuf<T: Message + Send + Default>(
+) -> impl Filter<Extract = (T,), Error = Rejection> + Copy {
+    async fn from_reader<T: Message + Send + Default>(buf: impl Buf) -> Result<T, Rejection> {
+        T::decode(buf).map_err(|err| {
+            log::debug!("request protobuf body error: {}", err);
+            reject::known(BodyDeserializeError { cause: err.into() })
+        })
+    }
+
+    aggregate().and_then(from_reader)
```
ParkMyCar: noteably, this does not include a content-type check, like the `form(...)` or `json(...)`, that is because the mime crate does not have one. I plan to submit a PR to that crate to add the MIME type, but have yet to

seanmonstar:
Thanks for the PR!

I wonder, is using protobuf really that common in web servers, besides using with gRPC? I settled on providing form-urlencoded and JSON because they are very common. If not, perhaps protobuf support can just be a 3rd party library?

ParkMyCar:
I've seen blog posts about writing APIs in proto, but I don't think it is very common. Serialized proto messages are smaller than serialized json though, which to me goes along with the performance characteristics of Rust.

More than happy to move this into a `warp-protobuf` repro though if you don't think it belongs :)

jxs:
also, have you checked [tonic](https://github.com/hyperium/tonic),? It's a Rust implementation of Grpc on top of hyper

jeffparsons:
> I wonder, is using protobuf really that common in web servers, besides using with gRPC? I settled on providing form-urlencoded and JSON because they are very common. If not, perhaps protobuf support can just be a 3rd party library?

FWIW, I'm interested in Protobuf for HTTP requests/responses, without gRPC.

Context: a large sprawling set of services written in multiple languages that use Protobuf as their lingua franca, both for live communication and persistence. HTTP for direct RPC between services is great because it's universal. It's also pretty easy to find Protobuf support for most languages. So the pairing works well, and wherever we don't have good Protobuf support we can fall back to an internal standard JSON encoding for that last mile.

gRPC on the other hand comes with too many opinions for our liking (we're not about to restructure all our stuff to fit the gRPC model of the world) and when we experimented we've been sorely disappointed by the maturity and reliability of gRPC implementations across multiple languages anyway. It just doesn't offer us anything that we need that much, and yet requires a significant investment. It would be the proverbial tail wagging the dog.

As for having Protobuf support in Warp itself vs. in an external crate... I don't have any strong opinions. If anything, I'd have an ever so slight slight preference for something like that living out-of-tree because it _is_ exotic compared to the other things Warp is doing, and having it out-of-tree might actually serve well as an example of how to extend Warp from the outside — which is something I hope to see lots of!

/ my AU$0.02

🕊 + 💖

seanmonstar:
> I'm interested in Protobuf for HTTP requests/responses, without gRPC.

I mean, I'm sure there are instances. I just haven't seen it often at all, and for that reason, feel it probably would be better left outside of warp core.

ParkMyCar:
Sounds good! Closing in favor of moving to a separate repo

PR #396, opened 2020-01-19
Add From<Reject> for Rejection.

Koxiaet:
This allows the `?` operator to easily be used from within functions without `map_err`.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -127,6 +127,8 @@ fn __reject_custom_compilefail() {}
 
 /// A marker trait to ensure proper types are used for custom rejections.
 ///
+/// Can be converted into Rejection.
+///
 /// # Example
 ///
 /// ```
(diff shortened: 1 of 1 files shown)


seanmonstar:
Hm, this may be a good idea, actually!

It *may* be a breaking change, though. Anyone how implemented `Reject` and `Into<Rejection>` for their own type would suddenly no longer compile with this change...

Koxiaet:
Yes, I just checked and it is a breaking change. Save it for the next breaking update maybe?

thomaseizinger:
What about adding a feature flag (like `unstable`) that is off by default and guards this (and other ideas which people may like but it is not totally clear if we want to commit to their API?)

AlisCode:
Hi!

Any news on this ? This sounds like a major ergonomics improvement.

nui:
The orphan rule that forbid this impl has been relaxed since Rust 1.41. So you can implement `From<T> for warp::Rejection` if `T` is your local type. This allows what OP asked without breaking changes.

jxs (approved):

jxs:
thanks for the wait! :blue_heart:

Issue #389, opened 2020-01-18
Interest in a map_body filter


asonix:
Hello!

I'm working on a library that needs to inspect the body of a request to compute a hash, but it doesn't need to _parse_ the body. I was wondering if there was any way to implement, or plans to implement a filter that can map over the body of a request without consuming it. This kind of feature could be useful as well for implementing compression/decompression middlewares

PR #397, opened 2020-01-19
Body wrap filter

asonix:
I recognize that this code is probably not near polished enough to be included, but I wanted to submit an example of a solution to the issue I rose in #389 

The things I'm specifically worried about for this implementation are:
- using a `Pin<Box<async move {}>>` instead of finding a way to represent this as generics, or providing a type that directly implements `Future`
- Relying on errors being `Into<Rejection>`. This is fine if things can possibly reject, but for entirely infallible operations, this magically produces a `Rejection` error type.
- Silently ignoring possible rejections from the body future, specifically returning the `Option` rather than a `Result`
- Since the closure passed to `map_request_body` isn't really nameable, returning the result from a function becomes difficult

Despite these worries, I think this idea could be useful to future discussions

### DIFF
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -17,8 +17,170 @@ use serde::de::DeserializeOwned;
 use serde_json;
 use serde_urlencoded;
 
-use crate::filter::{filter_fn, filter_fn_one, Filter, FilterBase};
-use crate::reject::{self, Rejection};
+use crate::filter::{filter_fn, filter_fn_one, Filter, FilterBase, WrapSealed};
+use crate::reject::{self, IsReject, Rejection};
+use crate::reply::Reply;
+
+/// Create a wrapping filter to manipulate a request body before it is parsed
+///
+/// This provides access to the underlying hyper::Body
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+///
+/// let body = warp::body::map_request_body(move |body| {
+///     async move {
+///         // process the body
+///         Some(body)
+///     }
+/// });
+///
+/// let route = warp::any()
+///     .map(warp::reply)
+///     .with(body);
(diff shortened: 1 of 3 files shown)


src/reject.rs
```diff
@@ -553,6 +553,16 @@ mod sealed {
...
     }
 
+    impl IsReject for Box<dyn IsReject + 'static> {
+        fn status(&self) -> StatusCode {
+            self.as_ref().status()
+        }
+
+        fn into_response(&self) -> crate::reply::Response {
+            self.as_ref().into_response()
+        }
```
asonix: This is leftover from a previous iteration and no longer relevant

seanmonstar:
This is an interesting idea! I get what you're trying to do. I have a couple concerns:

- This may solidify the `hyper::Body` type as the one stored in `Route`.
- Mutating the `Route` can cause problems if a later filter rejects, since we usually need to "undo" the mutation. For instance, what if this mapping should only apply to some path, and it rejects, the mapped body will still be stored in the `Route` to be processed by an external filter.

asonix:
@seanmonstar so I went ahead and re-did this whole thing to allow access to each `&Bytes` in the body stream, rather than an owned `Body`. This allows the creation of filters that can inspect the body, and prevents "un-undoable" request modifications.

It also avoids boxing things, and in general is much simpler.

src/filters/body.rs
```diff
@@ -20,6 +21,73 @@ use serde_urlencoded;
...
+                    acc = (func.clone())(acc, &bytes).await;
+                    sender
+                        .send_data(bytes)
+                        .await
+                        .map_err(|_| reject::known(BodyReplaceError { _p: () }))?;
+                }
+
+                Ok(acc)
+            } else {
+                Ok(acc)
```
asonix: it's reasonable that rejecting on an absent body is "more correct"

Issue #16, opened 2018-08-02
Custom `with` filters


bbatha:
Under the current api, `Filter::with` requires that the type be `W: Wrap` however the `Wrap` trait is sealed making it difficult to write custom wrapping filters. For instance, I'm trying to write a filter that will forward a header from a request to the response. So far I've come up with the filter to generate the either grab the request id or generate a new one:

```rust
let req_id = warp::header::<String>("x-request-id").or_else(|_| Ok(("generated".to_owned(),)))
```
But I can't figure out using the combinators provided how to store the request-id in the filter and use it on the response.

It also seems that you cannot call `map` after returning a `warp::reply` in the combinator chain so I can't even make the above work with boiler plate something like

```rust
let my_route = req_id.map(warp::reply).map(|id: String, reply: warp::Reply| {
    reply.set_header("x-request-id", id)
})
```

seanmonstar:
Right, the `with` and `Wrap` thing is sealed since it is pretty new, and I wasn't sure yet if it could fit all needs.

In your example, I wouldn't expect `req_id.map(warp::reply)` to work, since the `req_id` filter has 1 value extracted, and `warp::reply()` takes 0 arguments...

Would this work for you?

```rust
let my_route = req_id.map(|id| {
    Response::builder()
        .header("x-request-id", id)
        .body("")
})
```

bbatha:
> Right, the with and Wrap thing is sealed since it is pretty new, and I wasn't sure yet if it could fit all needs.

Totally understood. The library is excellent so far and after handlers have a lot more complexity.

> Would this work for you?

Ya the composition is a little awkward since this is a filter that I'd like to be universally be a applied. But I landed on something like this:

```rust
    let req_id = warp::header::<String>("x-request-id").or_else(|_| {
        Ok(("deadbeef".to_owned(),))
    }).map(|id: String| {
        let mut res = Response::builder();
        res.header("x-request-id", id.as_str());

        res
    }).map(|mut res: response::Builder| {
        res.body("bar").unwrap()
    });
```

Which allows me to compose different body handlers after, and filters that don't pass along arguments before.

kamalmarhubi:
I'm learning that writing these is kinda non-trivial: you need to name the `Wrapped` filter type, which means you effectively have to write the `Filter` impl by hand, and avoid any closures. This in turn leads you to need to write a `Future` impl by hand. I hope some of this gets easier with existential types, but in the meantime it's kinda rough.

[...]

PR #408, opened 2020-01-23
Add a tracing filter which enables contextual, structured logging

### LINKED ISSUE #288: Add tracing support
I'd like to use [`tracing`](https://github.com/tokio-rs/tracing/) with a `warp` API. Do you have any recommendations about the best way to do this? Should there be a `Wrap` for tracing like the `Log` one, or should `Filter` implement `tracing-futures::Instrument`?

I'm happy to work on a PR for this if I can figure out the right direction (and if you think this should be added to warp).

Thank you!

barskern:
[Tracing](https://crates.io/crates/tracing) enables us to keep track of some contextual state throughout the lifetime of a request. This means that any logging done when processing a request will automatically also log the context. This is especially useful when processing requests in parallel because it means that one can separate intertwined logs.

The author of tracing made a good and informative [video](https://www.youtube.com/watch?v=JjItsfqFIdo) about the subject, and showing off some of the benefits.

This is based upon @emschwartz [issue](https://github.com/seanmonstar/warp/issues/288) and his previous work.

This is mainly a rough first implementation to see if you would like this feature. If you are interested in this feature I was planning to go through all uses of `log` in the crate to replace with appropriate `span`'s and `event`'s. What do you think @seanmonstar ?

Closes #288

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,12 +37,16 @@ rustls = { version = "0.16", optional = true }
 tungstenite = { default-features = false, version = "0.9", optional = true }
 urlencoding = "1.0.0"
 pin-project = "0.4.5"
+tracing = "0.1.12"
+tracing-futures = "0.2.1"
 
 [dev-dependencies]
 pretty_env_logger = "0.3"
 serde_derive = "1.0"
 handlebars = "1.0.0"
 tokio = { version = "0.2", features = ["macros"] }
+tracing-subscriber = "0.1.6"
+tracing-log = "0.1.1"
 
 [features]
 default = ["multipart", "websocket"]
--- a/src/filters/mod.rs
+++ b/src/filters/mod.rs
@@ -19,6 +19,7 @@ pub mod path;
 pub mod query;
 pub mod reply;
 pub mod sse;
+pub mod tracing;
 #[cfg(feature = "websocket")]
 pub mod ws;
 
--- a/src/filters/tracing.rs
+++ b/src/filters/tracing.rs
@@ -1,0 +1,208 @@
+//! Tracing Filters
+
+use tracing::Span;
+
+use std::fmt;
+use std::net::SocketAddr;
+
+use http::{self, header};
+
+use crate::filter::{Filter, WrapSealed};
+use crate::reject::IsReject;
+use crate::reply::Reply;
+use crate::route::Route;
+
+use self::internal::WithTrace;
+
+/// Create a wrapping filter which adds a span with request info
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+///
+/// let route = warp::any()
+///     .map(warp::reply)
+///     .with(warp::tracing());
+/// ```
+pub fn tracing() -> Trace<impl Fn(Info) -> Span + Clone> {
+    let func = move |info: Info| {
+        tracing::info_span!(
+            "request",
+            remote_addr = %OptFmt(info.route.remote_addr()),
+            method = %info.method(),
+            path = %info.path(),
+            version = ?info.route.version(),
+            // status = %info.status().as_u16(),
+            referer = %OptFmt(info.referer()),
+            user_agent = %OptFmt(info.user_agent()),
+        )
+    };
+    Trace { func }
+}
+
+/// Create a wrapping filter which adds a custom span with request info
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+///
+/// let tracing = warp::tracing::custom(|info| {
+///     // Create a span using tracing macros
+///     tracing::info_span!(
+///         "request",
+///         method = %info.method(),
+///         path = %info.path(),
+///     );
+/// });
+/// let route = warp::any()
+///     .map(warp::reply)
+///     .with(tracing);
+/// ```
+pub fn custom<F>(func: F) -> Trace<F>
+where
+    F: Fn(Info) -> Span + Clone,
+{
+    Trace { func }
+}
+
+/// Decorates a [`Filter`](::Filter) to log requests and responses.
+#[derive(Clone, Copy, Debug)]
+pub struct Trace<F> {
+    func: F,
+}
+
+/// Information about the request/response that can be used to prepare log lines.
+#[allow(missing_debug_implementations)]
+pub struct Info<'a> {
+    route: &'a Route,
+}
+
+impl<FN, F> WrapSealed<F> for Trace<FN>
+where
+    FN: Fn(Info) -> Span + Clone + Send,
+    F: Filter + Clone + Send,
+    F::Extract: Reply,
+    F::Error: IsReject,
+{
+    type Wrapped = WithTrace<FN, F>;
+
+    fn wrap(&self, filter: F) -> Self::Wrapped {
+        WithTrace {
+            filter,
+            trace: self.clone(),
+        }
+    }
+}
+
+impl<'a> Info<'a> {
+    /// View the remote `SocketAddr` of the request.
+    pub fn remote_addr(&self) -> Option<SocketAddr> {
+        self.route.remote_addr()
+    }
+
+    /// View the `http::Method` of the request.
+    pub fn method(&self) -> &http::Method {
+        self.route.method()
+    }
+
+    /// View the URI path of the request.
+    pub fn path(&self) -> &str {
+        self.route.full_path()
+    }
+
+    /// View the `http::Version` of the request.
+    pub fn version(&self) -> http::Version {
+        self.route.version()
+    }
+
+    /// View the referer of the request.
+    pub fn referer(&self) -> Option<&str> {
+        self.route
+            .headers()
+            .get(header::REFERER)
+            .and_then(|v| v.to_str().ok())
+    }
+
+    /// View the user agent of the request.
+    pub fn user_agent(&self) -> Option<&str> {
+        self.route
+            .headers()
+            .get(header::USER_AGENT)
+            .and_then(|v| v.to_str().ok())
+    }
+
+    /// View the host of the request
+    pub fn host(&self) -> Option<&str> {
+        self.route
+            .headers()
+            .get(header::HOST)
+            .and_then(|v| v.to_str().ok())
+    }
+}
+
+struct OptFmt<T>(Option<T>);
+
+impl<T: fmt::Display> fmt::Display for OptFmt<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if let Some(ref t) = self.0 {
+            fmt::Display::fmt(t, f)
+        } else {
+            f.write_str("-")
+        }
+    }
+}
+
+mod internal {
+    use futures::{future::Inspect, FutureExt};
+
+    use std::future::Future;
+
+    use super::{Info, Trace};
+    use crate::filter::{Filter, FilterBase, Internal};
+    use crate::reject::IsReject;
+    use crate::reply::Reply;
+    use crate::route;
+
+    #[allow(missing_debug_implementations)]
+    #[derive(Clone, Copy)]
+    pub struct WithTrace<FN, F> {
+        pub(super) filter: F,
+        pub(super) trace: Trace<FN>,
+    }
+
+    use tracing::Span;
+    use tracing_futures::{Instrument, Instrumented};
+
+    impl<FN, F> FilterBase for WithTrace<FN, F>
+    where
+        FN: Fn(Info) -> Span + Clone + Send,
+        F: Filter + Clone + Send,
+        F::Extract: Reply,
+        F::Error: IsReject,
+    {
+        type Extract = F::Extract;
+        type Error = F::Error;
+        type Future = Inspect<Instrumented<F::Future>, fn(&<F::Future as Future>::Output)>;
+
+        fn filter(&self, _: Internal) -> Self::Future {
+            let span = route::with(|route| (self.trace.func)(Info { route }));
+            let _guard = span.enter();
+
+            fn finished_logger<R: Reply, E: IsReject>(reply: &Result<R, E>) {
+                if let Err(e) = reply {
+                    tracing::warn!(target: "warp::filters::tracing", msg = ?e, "error during processing");
+                } else {
+                    tracing::info!(target: "warp::filters::tracing", "successfully finished processing");
+                }
+            }
+
+            tracing::info!(target: "warp::filters::tracing", "processing request");
+            self.filter
+                .filter(Internal)
+                .in_current_span()
+                .inspect(finished_logger)
+        }
+    }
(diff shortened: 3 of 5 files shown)


seanmonstar:
I'm certainly interested in this, I think `tracing` can add a lot of value. Part of the reason I hadn't (yet) was I figured that applications that wanted `tracing` could add it themselves. Do you find that's not the case? What would you say is key in this PR solving that problem?

barskern:
> Part of the reason I hadn't (yet) was I figured that applications that wanted `tracing` could add it themselves. Do you find that's not the case?

I am not sure this would be possible to add with the current public API due to the fact that I used `Filter::with`. This gives a very nice abstraction over a wrapping a `Filter` with a tracing logger. 

However in retrospect I think one could also get similar functionality by appending the span to the future returned by `Server::bind`. I will try this out. A limitation of this is that one cannot add different spans to different filters, which the implementation using `Filter::with` allows.

> What would you say is key in this PR solving that problem?

Instead of providing tracing in warp, one could also expose e.g. `With` so that other libraries can create structs which wrap a filter.

barskern:
> However in retrospect I think one could also get similar functionality by appending the span to the future returned by `Server::bind`. I will try this out. A limitation of this is that one cannot add different spans to different filters, which the implementation using `Filter::with` allows.

After some testing I haven't figured out a way to do this without access to `Wrap`. The reason is that we need to create a span for each request, which means that we need to have some way of modifying the resulting future of a `Filter` (`FilterBase` to be more exact). As this trait is private to `warp`, it is currently not possible to do that.

In conclusion I think the best way forward would be to consider making `Wrap` unsealed if you want this functionality in a separate crate. However personally I would recommend you to change from `log` to `tracing` for the entire library as it brings a lot of great benefits. Especially the fact that its structured logs means that one can easily set up automated tasks to inspect logs, or easily search through logs.

jxs:
hi, have you tried the new Filter to Service conversion? you can now wrap a warp filter:
```Rust
use std::convert::Infallible;
use warp::Filter;
use hyper::{Body, Request};
use tower_service::Service;

#[tokio::main]
async fn main() -> Result<(), hyper::error::Error> {
    let route = warp::any().map(|| "Hello From Warp!");
    let mut warp_svc = warp::service(route);
    let make_svc = hyper::service::make_service_fn(move |_| async move {
        let svc = hyper::service::service_fn(move |req: Request<Body>| async move {
            println!("before request");
            let resp = warp_svc.call(req).await;
            println!("after request");
            resp
        });
        Ok::<_, Infallible>(svc)
    });

    hyper::Server::bind(&([127, 0, 0, 1], 3030).into())
        .serve(make_svc)
        .await?;

    Ok(())
}
```

barskern:
> hi, have you tried the new Filter to Service conversion? you can now wrap a warp filter:
> 
> ```rust
> use std::convert::Infallible;
> use warp::Filter;
> use hyper::{Body, Request};
> use tower_service::Service;
> 
> #[tokio::main]
> async fn main() -> Result<(), hyper::error::Error> {
>     let route = warp::any().map(|| "Hello From Warp!");
>     let mut warp_svc = warp::service(route);
>     let make_svc = hyper::service::make_service_fn(move |_| async move {
>         let svc = hyper::service::service_fn(move |req: Request<Body>| async move {
>             println!("before request");
>             let resp = warp_svc.call(req).await;
>             println!("after request");
>             resp
>         });
>         Ok::<_, Infallible>(svc)
>     });
> 
>     hyper::Server::bind(&([127, 0, 0, 1], 3030).into())
>         .serve(make_svc)
>         .await?;
> 
>     Ok(())
> }
> ```

Oh no I haven't tried that. That looks very promising! Thank you for the tip. I will try it out a bit later today.

LucioFranco:
There is also an in progress `tracing-tower` crate https://github.com/tokio-rs/tracing/tree/master/tracing-tower

barskern:
@jxs Your suggestion works perfectly (despite being a bit more verbose). Hence it is probably smarter to use this solution rather than adding dependencies and modifying `warp`. 

If you are still interested in exploring using `tracing` instead of `log` in warp, I will leave this PR open for now. Feel free to close it or merge it when you find it applicable.

```rust
use hyper::{Body, Request};
use std::convert::Infallible;
use tower_service::Service;
use tracing_futures::Instrument;
use warp::Filter;

#[tokio::main]
async fn main() -> Result<(), hyper::error::Error> {
    // Setup a log subscriber (responsible to print to output)
    let subscriber = tracing_subscriber::fmt::Subscriber::builder()
        .with_env_filter("tracing_service=trace")
        .without_time()
        .finish();

    // Set the previously created subscriber as the global subscriber
    tracing::subscriber::set_global_default(subscriber).unwrap();
    // Redirect normal log messages to the tracing subscriber
    tracing_log::LogTracer::init().unwrap();


    let route = warp::any().map(|| "Hello From Warp!");
    let mut warp_svc = warp::service(route);

    let make_svc = hyper::service::make_service_fn(move |_| {
        async move {
            let svc = hyper::service::service_fn(move |req: Request<Body>| {
                async move {
                    // FIXME instead of printing the entire request, we probably want to filter out
                    // some fields?
                    let span = tracing::info_span!("request", req = ?req);
                    let _guard = span.enter();

                    tracing::info!("processing request");
                    let resp = warp_svc.call(req).in_current_span().await;

                    tracing::info!(status = ?resp.as_ref().unwrap().status().as_u16(), "processed request");

                    resp
                }
            });
            Ok::<_, Infallible>(svc)
        }
    });

    hyper::Server::bind(&([127, 0, 0, 1], 3030).into())
        .serve(make_svc)
        .await?;

    Ok(())
}
```

@LucioFranco I would probably rather use that if it was published, I will keep my eye on it. Thank you for the tip!

barskern:
As an aside, would you be interested in having the code above as an example?

barskern:
Using the service function solution above I met some issues when the `Filter` is only `Clone` (not `Copy`). I made a post on the subreddit because it seems to be a general borrow checker problem, but linking it here aswell as it might be relevant when people stumble across this PR. 

https://www.reddit.com/r/rust/comments/ettmfs/trouble_moving_clonable_hyperservice_into_async/

fhsgoncalves:
@barskern thank you!
This is exactly what I was needing!

Also, you can avoid this issue cloning the service on every move, like that:
https://github.com/seanmonstar/warp/issues/16#issuecomment-578376676

little-dude:
> @jxs Your suggestion works perfectly (despite being a bit more verbose). Hence it is probably smarter to use this solution rather than adding dependencies and modifying `warp`.
> 
> If you are still interested in exploring using `tracing` instead of `log` in warp, I will leave this PR open for now. Feel free to close it or merge it when you find it applicable.
> 
> ```rust
> use hyper::{Body, Request};
> use std::convert::Infallible;
> use tower_service::Service;
> use tracing_futures::Instrument;
> use warp::Filter;
> 
> #[tokio::main]
> async fn main() -> Result<(), hyper::error::Error> {
>     // Setup a log subscriber (responsible to print to output)
>     let subscriber = tracing_subscriber::fmt::Subscriber::builder()
>         .with_env_filter("tracing_service=trace")
>         .without_time()
>         .finish();
> 
>     // Set the previously created subscriber as the global subscriber
>     tracing::subscriber::set_global_default(subscriber).unwrap();
>     // Redirect normal log messages to the tracing subscriber
>     tracing_log::LogTracer::init().unwrap();
> 
> 
>     let route = warp::any().map(|| "Hello From Warp!");
>     let mut warp_svc = warp::service(route);
> 
>     let make_svc = hyper::service::make_service_fn(move |_| {
>         async move {
>             let svc = hyper::service::service_fn(move |req: Request<Body>| {
>                 async move {
>                     // FIXME instead of printing the entire request, we probably want to filter out
>                     // some fields?
>                     let span = tracing::info_span!("request", req = ?req);
>                     let _guard = span.enter();
> 
>                     tracing::info!("processing request");
>                     let resp = warp_svc.call(req).in_current_span().await;
>                     tracing::info!(status = ?resp.as_ref().unwrap().status().as_u16(), "processed request");
> 
>                     resp
>                 }
>             });
>             Ok::<_, Infallible>(svc)
>         }
>     });
> 
>     hyper::Server::bind(&([127, 0, 0, 1], 3030).into())
>         .serve(make_svc)
>         .await?;
> 
>     Ok(())
> }
> ```
> 
> @LucioFranco I would probably rather use that if it was published, I will keep my eye on it. Thank you for the tip!

Thanks for sharing this example. I have a question about this portion of the snippet:

```rust
let span = tracing::info_span!("request", req = ?req);
let _guard = span.enter();
```

In the `tracing` repo [there is a warning](https://github.com/tokio-rs/tracing#in-asynchronous-code) against using `span.enter()` in async code. Is it actually ok in this specific example?

barskern:
Actually, in the service I am using this I have some problems with nested spans, and I think you just figured it out! I didn't know that was an issue in async code, but I will read the docs about it and edit the example accordingly. Thank you! :smile:

hawkw:
Welp, I just went and wrote my own PR to add a `tracing` filter PR, and only noticed this PR after I finished. Whoops! 

Looks like @barskern and I have implemented more or less the same thing. My code is [here](https://github.com/seanmonstar/warp/compare/master...hawkw:eliza/tracing) if you want to compare notes.

The one main difference in my PR is that I also added a `warp::trace::context` filter. This is just a convenience API for using `tracing` spans to name different filter stacks, like
```rust

    let hello = warp::path("hello")
        .and(warp::get())
        .map(|| {
            tracing::info!("saying hello...");
            "Hello, World!"
        })
        .with(warp::trace::context("hello"));
    
    let goodbye = warp::path("goodbye")
        .and(warp::get())
        .map(|| {
            tracing::info!("saying goodbye...");
            "So long and thanks for all the fish!"
        })
        .with(warp::trace::context("goodbye"));
```
Using that, we can get nice nested traces like this:
warp_traces

---

As for the question of whether or not Warp _should_ provide `tracing` support, I have some thoughts. It's definitely possible to get traces from Warp filters using the Tower compatibility, as folks have discussed above. However, this seems pretty ungainly, and something that's always seemed appealing to me about Warp is that it has a remarkably nice, fluent API. Providing Tracing `Filter`s instead of going through the Tower compatibility layer lets users write more concise, nicer looking, "warp-ier" code. I also appreciate Warp's relative minimalism, but the Tracing filters could always be a default-off feature.

barskern:
That is why I didn't close this PR yet, because I sort of expected others to want this functionality aswell, and leaving it open makes it a bit easier to discover for others.

I really like the addition you made using `context`! That is a great way of adding useful information to the queries.

---

Further I think your argument regarding the "warp-ier" code by including this in `warp` is a good one. Currently using `tracing` with warp using my workaround (see earlier comment in this issue) is very verbose, quite "messy" and easy to get wrong (as I first had some borrowing issues and then a wrong usage of tracing).

I do however understand the maintenance burden of adding this feature, and that the nice API of warp is certainly due to a lack of non-essential features. 

I think that it would be better to try to stabilize and unseal a version of the `With` (with the functionality of `WithSealed`) trait to enable this to be created in a separate library.

little-dude:
With this implementation would the request's span be available in `.recover()` ? 

Right now this is my problem: I create a span for the request in `.and_then()`, but I don't have access to it in `.recover()`, so in the logs I don't have all the information I want when a request fails.

Here is a snippet from my code to illustrate the problem:

```rust
    let download_global_weights = warp::get()
        .and(warp::path::param::<ClientId>())
        .and(warp::path::param::<Token>())
        .and(handle.clone())
        // handler
        .and_then(move |id, token, handle| {
            // in this span, I have the `client_id`
            let span =
                trace_span!(parent: parent_span.clone(), "api_download_request", client_id = %id);
            handle_download_request(id, token, handle).instrument(span)
        })
        // here I won't have the `client_id` in the logs
        .recover(handle_download_rejection)
        .recover(handle_rejection)
```

barskern:
> With this implementation would the request's span be available in `.recover()` ?
> 
> Right now this is my problem: I create a span for the request in `.and_then()`, but I don't have access to it in `.recover()`, so in the logs I don't have all the information I want when a request fails.
> 
> Here is a snippet from my code to illustrate the problem:
> 
> ```rust
>     let download_global_weights = warp::get()
>         .and(warp::path::param::<ClientId>())
>         .and(warp::path::param::<Token>())
>         .and(handle.clone())
>         // handler
>         .and_then(move |id, token, handle| {
>             // in this span, I have the `client_id`
>             let span =
>                 trace_span!(parent: parent_span.clone(), "api_download_request", client_id = %id);
>             handle_download_request(id, token, handle).instrument(span)
>         })
>         // here I won't have the `client_id` in the logs
>         .recover(handle_download_rejection)
>         .recover(handle_rejection)
> ```

Yes, using this implementation it. Here you instrument the future in the `and_then` part

barskern:
> With this implementation would the request's span be available in `.recover()` ?
> 
> Right now this is my problem: I create a span for the request in `.and_then()`, but I don't have access to it in `.recover()`, so in the logs I don't have all the information I want when a request fails.
> 
> Here is a snippet from my code to illustrate the problem:
> 
> ```rust
>     let download_global_weights = warp::get()
>         .and(warp::path::param::<ClientId>())
>         .and(warp::path::param::<Token>())
>         .and(handle.clone())
>         // handler
>         .and_then(move |id, token, handle| {
>             // in this span, I have the `client_id`
>             let span =
>                 trace_span!(parent: parent_span.clone(), "api_download_request", client_id = %id);
>             handle_download_request(id, token, handle).instrument(span)
>         })
>         // here I won't have the `client_id` in the logs
>         .recover(handle_download_rejection)
>         .recover(handle_rejection)
> ```

I haven't tested, but I am quite certain that this implementation would make the span avaliable in the `recover` method, as long as the `with` call with the span is the last call in the chain.

hlb8122:
Super excited for this!

ufoscout:
Is there any progress on this issue? Or is there something we can do to help to fix it?

As tracing is fundamental in my company, this is becoming a blocking issue for us. I know that there is a workaround using the `tower::service`, but we are mostly interested in the solution provided in https://github.com/seanmonstar/warp/pull/408#issuecomment-606325196

We tried to reimplement it by ourselves but that implementation depends on private traits (BTW is there a reason not to make those traits/struct public?).

hgzimmerman:
I would really like tracing functionality added to Warp, and I don't think its practical to implement it outside of the library due to `warp::filter::wrap::Wrap` and `warp::route::Route` being private.
I personally like the implementation detailed in comment https://github.com/seanmonstar/warp/pull/408#issuecomment-606325196.
@hawkw would you be willing to open a PR with the patch listed in your comment and @jxs (or another collaborator) would you be willing to assist incorporating it into the project?

seanmonstar:
Thanks for exploring this, we've merged `tracing` support in #656.

PR #426, opened 2020-01-28
switch rustls for tokio-rustls

jxs:
[`tokio_rustls::server::TlsStream`](https://docs.rs/tokio-rustls/0.12.2/tokio_rustls/server/struct.TlsStream.html) doesn't expose constructor methods, the public entrypoint is TlsAcceptor which has [`accept`](https://docs.rs/tokio-rustls/0.12.2/tokio_rustls/struct.TlsAcceptor.html#method.accept) that returns [`Accept`](https://docs.rs/tokio-rustls/0.12.2/tokio_rustls/struct.Accept.html) which impls `Future`.
The way i found to get around it was implementing AsyncRead/AsyncWrite on a wrapper `TlsStream` that handshakes `Accept` first. 

I tried first handshaking on `warp::tls::TlsAcceptor` but on errors `Hyper` server would stop

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,11 +32,11 @@ serde_json = "1.0"
 serde_urlencoded = "0.6"
 tokio = { version = "0.2", features = ["blocking", "fs", "stream", "sync", "time"] }
 tower-service = "0.3"
-rustls = { version = "0.16", optional = true }
 # tls is enabled by default, we don't want that yet
 tokio-tungstenite = { version = "0.10", default-features = false, optional = true }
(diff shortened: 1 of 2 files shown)


quininer:
I think `poll_accept` should return a `Future` to complete tls handshake.

like
```rust
trait Accept {
    type Conn;
    type Error;
    type Future: Future<Output = Result<Self::Conn, Self::Error>>;

    fn poll_accept(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Future>>;
}

Issue #269, opened 2019-09-11
websocket `auto_ping` as parallel to sse `keep_alive`


codesections:
The SSE filter has a `keep_alive` method that sends periodic messages to clients to persist the connection.  However, the WS filter lacks similar functionality, even though it is sometimes needed and is often included (for example, in the `startAutoPing` method in [Cluster WS](https://github.com/ClusterWS/cWS), one of the more prominent WebSockets libraries in the JavaScript ecosystem).

Would it be possible to add that functionality?  If so, I'm happy to help—though I took a look at the code and am not quite sure where to start, so I'd appreciate a bit of guidance.

najamelan:
The underlying implementation (tungstenite) supports Ping/Pong messages, but warp has created their own API which doesn't. 

So either:
- the warp API is to be rewritten to have a better coverage of the websocket protocol
- you use hyper to upgrade http connections and switch to tungstenite yourself
- use a websocket library (tungstenite/ws/websocket) directly on another port

Hope it clarifies things

codesections:
Thanks, that was my understanding as well.  

> the warp API is to be rewritten to have a better coverage of the websocket protocol

That is exactly what I am proposing.  If no one objects, I can work on an initial PR that would lay the groundwork for an extension of the Warp API to allow for ping messages.  I think the first step would be to give the `Message` struct a `.ping` public method, paralleling the `.text` and `.binary` methods it already exposes.

najamelan:
You can have a look at prior thoughts here:  #257 and #258

emschwartz:
@codesections have you started on this already? I also need either the manual way to send / receive pings and pongs or the `auto_ping` functionality you describe. If you haven't started working on it, I can also take a stab at implementing it.

codesections:
> have you started on this already?

My apologies for the delay in getting back to you on this.  All I have done so far is the (very small) PR in #281, which would allow users to manually send Pings

najamelan:
> @codesections have you started on this already? I also need either the manual way to send / receive pings and pongs or the `auto_ping` functionality you describe. If you haven't started working on it, I can also take a stab at implementing it.

Just remember that tungstenite always responds to pings automatically as long as you keep polling the stream or sending things in the sink. You should never manually answer the pings.

codesections:
> Just remember that tungstenite always responds to pings automatically as long as you keep polling the stream or sending things in the sink. You should never manually answer the pings.

[...]

PR #429, opened 2020-01-30
Framework for ping-pong keep-alives

jq-rs:
Include full ping and pong message support for manual creation of WebSocket keep-alives as described in https://tools.ietf.org/html/rfc6455#section-5.5.2

Part of #269

### DIFF
--- a/src/filters/path.rs
+++ b/src/filters/path.rs
@@ -137,6 +137,7 @@ use std::str::FromStr;
 use http::uri::PathAndQuery;
 
 use filter::{filter_fn, one, Filter, One, Tuple};
+use never::Never;
 use reject::{self, Rejection};
 use route::Route;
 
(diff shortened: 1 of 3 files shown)


jq-rs:
Fails of CI seem to be non-related to changes.

seanmonstar:
(Sorry for the slow response, been out of the office.)

I believe all of this exists in 0.2. Is there a reason to backport instead of just upgrading?

jq-rs:
Thanks for your reply! It takes quite an effort for me to upgrade to 0.2, that is the primary reason for 0.1 PR. I am not familiar how you have planned to maintain 0.1 so decided to do a PR anyway.

Regarding 0.2 differences, I believe pong send support is missing from 0.2. I did not see quickly that is pong receive missing due to differences between 0.1 and 0.2, it was originally.

Path and query fixes are added here by mistake. Nevertheless, they fix a real crash seen on live and would be nice to have in both branches. Surprisingly http crate's `path_and_query` gets an error and expect branch is taken in Warp.

jq-rs:
I can drop the `path_and_query` fixes, if this is wanted to be merged. If 0.1.x is not maintained anymore, then no need and this can be closed.

seanmonstar:
I can merge and release a 0.1.x patch. To help me understand, can we separate the path_and_query piece?

jq-rs:
Really grand if 0.1.x is supported still. I reverted the `path_and_query` changes.

jq-rs:
Thanks for the merge! Any guidance what should I do with the `path_and_query` changes? One option is to try to do a version which does not change the API, but instead returns e.g. index-path/empty-path by default if http crate's path and query fails.

seanmonstar:
What is the change trying to fix? I don't yet understand what the problem was :D

jq-rs:
The problem is that `tail()` call crashes to `.expect("server URIs should always have path_and_query")`.

seanmonstar:
Hm, I guess the only time that's not true would be `CONNECT foo.bar`...

jq-rs:
Some bot is probably doing that to public IPs all the time.

Issue #366, opened 2020-01-08
update websocket filter to use tokio-tungstenite


jxs:
now that [tokio-tungstenite](https://github.com/snapview/tokio-tungstenite/) has been updated to `std::futures` we could update the websockets filter to use it instead, should we wait for it to be published though?

seanmonstar:
I can't remember why I removed the support before... What do we gain by adding it back? Less code to maintain?

jxs:
yeah, also less `unsafe` in the codebase  since their implementation no longer has unsafe, and tls support for websockets afaik

jxs:
fixed with #430

PR #430, opened 2020-01-30
Websocket filter with tokio-tungstenite

AlexCovizzi:
I tried to add support for `tokio-tungstenite` in the websocket filter. The only difference in the internal api is that `WebSocket::from_raw_websocket` now returns a Future.

The main problem right now is that the `tokio-tungstenite` dependency is from the github repository  because the version on crates.io uses futures v1.

Edit: Reference #366

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -3,20 +3,19 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::future::Future;
-use std::io::{self, Read, Write};
+use std::io::{self};
 use std::pin::Pin;
-use std::ptr::null_mut;
 use std::task::{Context, Poll};
 
 use super::{body, header};
 use crate::filter::{Filter, One};
 use crate::reject::Rejection;
 use crate::reply::{Reply, Response};
-use futures::{future, FutureExt, Sink, Stream, TryFutureExt};
+use futures::{future, FutureExt, Stream, Sink};
 use headers::{Connection, HeaderMapExt, SecWebsocketAccept, SecWebsocketKey, Upgrade};
 use http;
-use tokio::io::{AsyncRead, AsyncWrite};
 use tungstenite::protocol::{self, WebSocketConfig};
+use tokio_tungstenite::WebSocketStream;
 
 /// Creates a Websocket Filter.
 ///
@@ -130,24 +129,20 @@ where
             .ws
             .body
             .on_upgrade()
-            .and_then(move |upgraded| {
+            .then(move |upgraded| {
                 log::trace!("websocket upgrade complete");
-
-                let io = protocol::WebSocket::from_raw_socket(
-                    AllowStd {
-                        inner: upgraded,
-                        context: (true, null_mut()),
-                    },
-                    protocol::Role::Server,
-                    config,
-                );
-
-                on_upgrade(WebSocket { inner: io }).map(Ok)
+                let socket = WebSocketStream::from_raw_socket(upgraded.unwrap(), protocol::Role::Server, config);
+                socket
             })
-            .map(|result| {
+            .then(move |socket| {
+                on_upgrade(WebSocket { inner: socket })
+            })
+            .map(|_result| {
+                /*
                 if let Err(err) = result {
                     log::debug!("ws upgrade error: {}", err);
                 }
+                */
             });
         ::tokio::task::spawn(fut);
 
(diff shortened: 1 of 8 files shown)


kellytk:
@AlexCovizzi https://github.com/snapview/tokio-tungstenite/issues/64#issuecomment-580555861 has been published. (https://github.com/snapview/tokio-tungstenite/releases/tag/v0.10.0 and crates)

src/filters/ws.rs
```diff
@@ -242,36 +173,27 @@ fn cvt<T>(r: tungstenite::error::Result<T>, err_message: &str) -> Poll<Result<T,
...
-            config,
-        );
-
-        WebSocket { inner: ws }
+        WebSocketStream::from_raw_socket(upgraded, role, config)
+            .map(|inner| WebSocket { inner })
+            .await
     }
 
     fn with_context<F, R>(&mut self, ctx: Option<&mut Context<'_>>, f: F) -> R
```
jxs: I don't think we need `with_context` anymore, it was used to pass the `Context` to `Read` and `Write`, we can just call `Pin::new(&mut self)` or assign it to a variable if needed

src/filters/ws.rs
```diff
@@ -284,19 +206,26 @@ impl Stream for WebSocket {
     type Item = Result<Message, crate::Error>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
-        match (*self).with_context(Some(cx), |s| s.read_message()) {
-            Ok(item) => Poll::Ready(Some(Ok(Message { inner: item }))),
-            Err(::tungstenite::Error::Io(ref err)) if err.kind() == io::ErrorKind::WouldBlock => {
+        match (*self).with_context(Some(cx), |s, cx| s.poll_next(cx.unwrap())) {
+            Poll::Ready(Some(Ok(item))) => Poll::Ready(Some(Ok(Message { inner: item }))),
+            Poll::Ready(Some(Err(tungstenite::Error::Io(ref err))))
+                if err.kind() == io::ErrorKind::WouldBlock =>
```
jxs: same as below

src/filters/ws.rs
```diff
@@ -305,19 +234,24 @@ impl Sink<Message> for WebSocket {
...
     fn start_send(mut self: Pin<&mut Self>, item: Message) -> Result<(), Self::Error> {
-        match self.with_context(None, |s| s.write_message(item.inner)) {
+        match self.with_context(None, |s, _| s.start_send(item.inner)) {
             Ok(()) => Ok(()),
             // Err(::tungstenite::Error::SendQueueFull(inner)) => {
             //     log::debug!("websocket send queue full");
             //     Err(::tungstenite::Error::SendQueueFull(inner))
             // }
-            Err(::tungstenite::Error::Io(ref err)) if err.kind() == io::ErrorKind::WouldBlock => {
+            Err(tungstenite::Error::Io(ref err)) if err.kind() == io::ErrorKind::WouldBlock => {
```
jxs: do we need to address io::ErrorKind::WouldBlock ? it seems it's already being checked on [tokio-tungstenite]( https://github.com/snapview/tokio-tungstenite/blob/607b8bf3d801bb14898fe4c5505cc6292950cd7a/src/lib.rs#L291)

src/filters/ws.rs
AlexCovizzi: Ok, I removed `with_context` and replaced it with `Pin::new(&mut self)`

src/filters/ws.rs
AlexCovizzi: DIdn't notice, removed

src/filters/ws.rs
jxs: only this one's left

AlexCovizzi: this was also removed in the last commit

src/filters/ws.rs
```diff
@@ -284,19 +185,26 @@ impl Stream for WebSocket {
     type Item = Result<Message, crate::Error>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
-        match (*self).with_context(Some(cx), |s| s.read_message()) {
-            Ok(item) => Poll::Ready(Some(Ok(Message { inner: item }))),
-            Err(::tungstenite::Error::Io(ref err)) if err.kind() == io::ErrorKind::WouldBlock => {
+        match Pin::new(&mut self.inner).poll_next(cx) {
+            Poll::Ready(Some(Ok(item))) => Poll::Ready(Some(Ok(Message { inner: item }))),
+            Poll::Ready(Some(Err(tungstenite::Error::Io(ref err))))
+                if err.kind() == io::ErrorKind::WouldBlock =>
```
jxs: there were two, this one is still there

AlexCovizzi: Oh the `WouldBlock`, sorry, I thoght you meant the `with_context`.
It's not checked in the `tokio-tungstenite`'s [ `poll_next`](https://github.com/snapview/tokio-tungstenite/blob/607b8bf3d801bb14898fe4c5505cc6292950cd7a/src/lib.rs#L260), do I remove it anyway?

jxs: no worries :) 
yeah, from what I see [poll_next](https://github.com/snapview/tokio-tungstenite/blob/607b8bf3d801bb14898fe4c5505cc6292950cd7a/src/lib.rs#L260) calls [cvt](https://github.com/snapview/tokio-tungstenite/blob/master/src/lib.rs#L268) under `with_context` that [does that conversion](https://github.com/snapview/tokio-tungstenite/blob/master/src/compat.rs#L200)

AlexCovizzi: Looking at `poll_next` I noticed that `Error::ConnectionClosed` is also already handled.

AlexCovizzi: Fixed, I removed the both the `WouldBlock` and the `ConnectionClosed` check

jxs: thanks! had missed `Error::ConnectionClosed`.
 only some final minor nit that I also had missed, can you call [`ready!`](https://docs.rs/futures/0.3.1/futures/macro.ready.html) on the `Pin::new.poll` cases, so that Poll::Pending is handled, and the match brace becomes more readable, you have an example here: https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs#L399

AlexCovizzi: Done!

seanmonstar:
Sorry, I've been sick this week.

@jxs if it looks good to you, you can merge any internal changes ;) I'd just want to discuss API changes.

jxs:
thanks Sean! hope you are feeling better now :relieved:

PR #431, opened 2020-01-31
Upgrade automatically from http to https.

### LINKED ISSUE #417: Upgrade connection from HTTP to HTTPS
I have a case where I want to host a server on a non-standard port, like `8000` for example.

What I want is that users enter `asd.com:8000` and it would lead them directly to the server, which currently isn't possible in combination with `tls`, because most browsers automatically connect to `http`  by default, which will be denied.

So my question: is upgrading from `http` to `https` without having to provide your own `TlsAcceptor` in the scope of `warp`?

I'm already working on a PR.

daxpedda:
Fixes #417.

This gives the ability to handle upgrading HTTP to HTTPS on the same port.

-  by default disabled
-  on detecting a connection through HTTP, it will send a `301` status and a `Location: https://...` header
-  return `403` on failure to build `Location` header

This couldn't have been done without the help of @LucioFranco, @sfackler, @jxs, @davidbarsky, @darksonn (sorry for pulling you all in on this 😄) on discord, thanks everybody!

EDIT: relies on https://github.com/hyperium/hyper/pull/2127.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ tokio-tungstenite = { version = "0.10", default-features = false, optional = tru
 urlencoding = "1.0.0"
 pin-project = "0.4.5"
 tokio-rustls = { version = "0.12.2", optional = true }
+tower-util = "0.3"
 
 [dev-dependencies]
 pretty_env_logger = "0.4"
--- a/src/server.rs
+++ b/src/server.rs
@@ -8,10 +8,12 @@ use std::net::SocketAddr;
 use std::path::Path;
 
 use futures::{future, FutureExt, TryFuture, TryStream, TryStreamExt};
+use http::Response;
 use hyper::server::conn::AddrIncoming;
 use hyper::service::{make_service_fn, service_fn};
-use hyper::Server as HyperServer;
+use hyper::{Body, Server as HyperServer};
 use tokio::io::{AsyncRead, AsyncWrite};
+use tower_util::Either;
 
 use crate::filter::Filter;
 use crate::reject::IsReject;
@@ -53,11 +55,33 @@ macro_rules! into_service {
     ($into:expr) => {{
         let inner = crate::service($into);
         make_service_fn(move |transport| {
-            let inner = inner.clone();
-            let remote_addr = Transport::remote_addr(transport);
-            future::ok::<_, Infallible>(service_fn(move |req| {
-                inner.call_with_addr(req, remote_addr)
-            }))
+            if Transport::upgrade_to_https(transport) {
+                future::ok::<_, Infallible>(Either::B(service_fn(move |req| {
+                    let response = if let Some(host) = req
+                        .headers()
+                        .get("host")
+                        .and_then(|header| header.to_str().ok())
+                    {
+                        Response::builder()
+                            .status(301)
+                            .header("Location", format!("https://{}{}", host, req.uri()))
+                            .body(Body::default())
+                            .unwrap()
+                    } else {
+                        Response::builder()
+                            .status(403)
+                            .body(Body::default())
+                            .unwrap()
+                    };
+                    future::ok::<_, Infallible>(response)
+                })))
+            } else {
+                let inner = inner.clone();
+                let remote_addr = Transport::remote_addr(transport);
+                future::ok::<_, Infallible>(Either::A(service_fn(move |req| {
+                    inner.call_with_addr(req, remote_addr)
+                })))
+            }
         })
     }};
 }
@@ -84,8 +108,8 @@ macro_rules! bind_inner {
     (tls: $this:ident, $addr:expr) => {{
         let service = into_service!($this.server.filter);
         let (addr, incoming) = addr_incoming!($addr);
-        let tls = $this.tls.build()?;
-        let srv = HyperServer::builder(crate::tls::TlsAcceptor::new(tls, incoming))
+        let (tls, upgrade) = $this.tls.build()?;
+        let srv = HyperServer::builder(crate::tls::TlsAcceptor::new(tls, incoming, upgrade))
             .http1_pipeline_flush($this.server.pipeline)
             .serve(service);
         Ok::<_, Box<dyn std::error::Error + Send + Sync>>((addr, srv))
@@ -390,6 +414,11 @@ where
     /// Specify the in-memory contents of the certificate.
     pub fn cert(self, cert: impl AsRef<[u8]>) -> Self {
         self.with_tls(|tls| tls.cert(cert.as_ref()))
+    }
+
+    /// Specify if incoming HTTP connections should be upgraded to HTTPS.
+    pub fn upgrade(self, upgrade: bool) -> Self {
+        self.with_tls(|tls| tls.upgrade(upgrade))
     }
 
     fn with_tls<Func>(self, func: Func) -> Self
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -50,6 +50,7 @@ impl std::error::Error for TlsConfigError {}
 pub(crate) struct TlsConfigBuilder {
     cert: Box<dyn Read + Send + Sync>,
     key: Box<dyn Read + Send + Sync>,
+    upgrade: bool,
 }
 
 impl std::fmt::Debug for TlsConfigBuilder {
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -37,6 +37,7 @@ rustls = { version = "0.16", optional = true }
 tungstenite = { default-features = false, version = "0.9", optional = true }
 urlencoding = "1.0.0"
 pin-project = "0.4.5"
+tower-util = "0.3"
```
daxpedda: This is only to pull in `Either`, the dependency is very small.

src/server.rs
```diff
@@ -53,11 +55,33 @@ macro_rules! into_service {
...
+                        .and_then(|header| header.to_str().ok())
+                    {
+                        Response::builder()
+                            .status(301)
+                            .header("Location", format!("https://{}{}", host, req.uri()))
+                            .body(Body::default())
+                            .unwrap()
+                    } else {
+                        Response::builder()
+                            .status(403)
```
daxpedda: When requesting HTTP, it should always send a `Host` header.
Also currently when a HTTPS connection fails for other reasons, it will just send back nothing, I could make it send this instead? Same for HTTP actually.

src/server.rs
```diff
@@ -84,8 +108,8 @@ macro_rules! bind_inner {
     (tls: $this:ident, $addr:expr) => {{
         let service = into_service!($this.server.filter);
         let (addr, incoming) = addr_incoming!($addr);
-        let tls = $this.tls.build()?;
-        let srv = HyperServer::builder(crate::tls::TlsAcceptor::new(tls, incoming))
+        let (tls, upgrade) = $this.tls.build()?;
```
daxpedda: I could wrap `ServerConfig` in our own struct to include `upgrade`, but returning a tuple should be fine?

src/tls.rs
```diff
@@ -251,28 +259,30 @@ pub(crate) struct TlsStream<T> {
     io: AllowStd<T>,
     is_shutdown: bool,
     session: ServerSession,
+    is_tls: bool,
 }
 
 impl<T> TlsStream<T> {
```
daxpedda: I could make a `TlsUpgradeStream` if these `if` checks are disturbing performance.

src/tls.rs
```diff
@@ -352,14 +384,40 @@ impl Accept for TlsAcceptor {
         cx: &mut Context<'_>,
     ) -> Poll<Option<Result<Self::Conn, Self::Error>>> {
         let pin = self.get_mut();
-        match ready!(Pin::new(&mut pin.incoming).poll_accept(cx)) {
-            Some(Ok(sock)) => {
-                let session = ServerSession::new(&pin.config.clone());
-                // let tls = Arc::new($this.config);
-                return Poll::Ready(Some(Ok(TlsStream::new(sock, session))));
+        // fast path for not upgrading
+        if !pin.upgrade {
```
daxpedda: There is some duplicate code here to keep the default path as minimal as it was before.
I could refactor the duplicate code out into a function if thats better?

src/tls.rs
```diff
@@ -352,14 +384,40 @@ impl Accept for TlsAcceptor {
...
+                    let session = ServerSession::new(&pin.config.clone());
+                    // let tls = Arc::new($this.config);
+                    Poll::Ready(Some(Ok(TlsStream::new(sock, session, true))))
+                }
+                Some(Err(e)) => Poll::Ready(Some(Err(e))),
+                None => Poll::Ready(None),
+            }
+        } else if let Some(sock) = pin.socket.as_mut() {
+            let mut first_byte = [0];
+            match ready!(sock.poll_peek(cx, &mut first_byte)) {
```
daxpedda: This will keep peeking at the first byte until `Ready`, until then it will keep `Pending`.
It doesn't need a timeout or anything, I'm assuming this handled somewhere else?

src/tls.rs
daxpedda: I played around with hit a little, it's done.

src/tls.rs
```diff
@@ -327,20 +351,53 @@ impl<T: Transport + Unpin> Transport for TlsStream<T> {
...
+        // let tls = Arc::new($this.config);
+        Poll::Ready(Some(Ok(TlsStream::new(sock, session, is_tls))))
+    }
+}
+
+/// check for valid HTTPS before moving to `TlsStream`
+macro_rules! poll_check {
+    ($self:expr, $cx:expr, $sock:expr) => {{
+        let mut first_byte = [0];
+        match ready!($sock.poll_peek($cx, &mut first_byte)) {
```
daxpedda: This will keep peeking at the first byte until `Ready`, until then it will keep `Pending`.
It doesn't need a timeout or anything, I'm assuming this is handled somewhere else?

src/tls.rs
daxpedda: I moved this comment because it's location is outdated.

seanmonstar:
Thanks for the work so far! I've been sick this week, and next week I'll be away from the computer as well, so sorry it will take me so longer to take a look.

daxpedda:
That's no problem at all, I hope you are feeling better!

src/tls.rs
daxpedda: With the newest changes I suppose this is moot.

src/tls.rs
daxpedda: This is also now handled in `tokio-rustls`.

daxpedda:
I rebased this because of #426, which simplified this PR also quite a bit.
hyperium/hyper#2127 is merged now, but we still need a new version of `hyper` to be released before we merge this into master I suppose?

daxpedda:
So I found a bug that I didn't notice before.
Apparently when the user is redirect from HTTP to HTTPS, it will hang for a while.

I found out the problem is the following:
1. `TlsAcceptor` runs, using `poll_peek` it will detect an HTTP connection and spawn `TlsStream` accordingly
2. `TlsStream` will stream the HTTP to `HyperServer` and a service will correctly send the user the HTTPS redirection header
3. the user tries to connect again over HTTPS now
4. `TlsStream` which is still configured to decode HTTP, incorrectly tries to stream garbage to `HyperServer` and fails!
5. `TlsAcceptor` runs, but `poll_peek` will not find any data, because `TlsStream` already starved the `AddrStream`
6. user web page loading gets stuck until it resets and tries again

So basically:
- how can I shut down the `TlsStream` after I send the redirection header?
- or how can I tell `TlsStream` that this is a new connection and it should stop starving the `AddrStream` and just shut down on it's own

daxpedda:
So apparently, as far as I could figure it out at least,  the issue has nothing to do with `warp`, but somehow with `poll_peek`.
Additionally I can't reproduce this issue other then on my local machine and only with chrome. I tried plenty of machines with the same OS and chrome now and couldn't reproduce it.

daxpedda:
Rebased again.

I was still vehemently trying to fix that one bug that I can't reproduce anywhere else except on my local device, to no avail 😞.
See https://github.com/hyperium/hyper/issues/2200 for details.

Pierstoval:
Hello @daxpedda , just a friendly ping: how are you feeling this PR after a couple months, do you need some help? 😉

Pierstoval:
For the record @daxpedda, I tested this PR by fetching the changes and merging them right onto warp v0.2.5.

If I don't update my code at all, it upgrades successfully:

```
curl -i http://127.0.0.1:8000 
HTTP/1.1 301 Moved Permanently
location: https://127.0.0.1:8000/
content-length: 0
date: Sun, 29 Nov 2020 17:38:13 GMT
```

However, TLS doesn't work anymore:

```
$ curl -i https://127.0.0.1:8000
curl: (35) error:1408F10B:SSL routines:ssl3_get_record:wrong version number
```

Not sure what's wrong here, I'll try to investigate a bit, but I don't think I'm skilled enough :sweat_smile:

daxpedda:
@Pierstoval Sorry for my absence!

I don't remember the exact details, but I remember understanding that how I've done this isn't really working and that part of the solution was https://github.com/hyperium/hyper/issues/2321. I believe that the issue was that `Acceptor` isn't polling and therefore not woken up, so this code has to be moved earlier down the line, where the connection is still being polled. Also this feature should probably be combined with HSTS.

Feel free to re-use any code I produced here.

PR #439, opened 2020-02-11
Add specialized Result type for warp errors

joshleeb:
This PR adds a specialized `Result` type alias, similar to the
`io::Result` alias seen in the stdlib.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,6 +1,10 @@
 use std::convert::Infallible;
 use std::error::Error as StdError;
 use std::fmt;
+use std::result;
+
+/// A specialized `Result` type for warp errors.
+pub type Result<T> = result::Result<T, Error>;
 
 type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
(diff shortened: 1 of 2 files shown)


joshleeb:
> is it that common to deal with `warp::Errors`?

I wouldn't say common, but then again I'm still fairly new when it comes to working with warp.

The use case for me was with websockets and specifically wanting to keep track of all connections in a map of type `HashMap<Id, UnboundedSender<warp::Result<Message>>`.

> I'd think Rejections would probably be more common, and thus warp::Result<Foo> might be confusing as to which is the error type

Fair point :slightly_smiling_face: 

IMHO it's pretty clear from the code examples. And given that `warp::Result` is just an alias, with no extra logic, I wouldn't think it adds any more to the initial confusion of having the separate `Error` and `Rejection` types. Though when I looked into it and saw the docs on `warp::Error` (of it being used for "errors inside Warp") that cleared things up for me personally.

Happy for you to take it or leave it either way :slightly_smiling_face:  :+1:

seanmonstar:
For now, I'm going to close due to the reasons I mentioned before. Thanks though!

PR #443, opened 2020-02-12
remove (Clone + Send + Sync + 'static) from server::serve

s3bk:
It isn't needed.


s3bk:
Actually it is needed. I will see if I can remove the Sync.

Issue #440, opened 2020-02-11
Expose private fields on standard rejects


johnchildren:
When handling rejections in `recover` it would be quite handy to be able to access in internal fields of the standard rejection structs. For example, with a missing header it would be nice to be able to see which header was missing, such as when the "Authorization" header is missing and you would like to provide a "WWW-Authenticate" header in your rejection response.

I suspect there are other use cases as well for doing this when you want to include more information in error messages and it should be a reasonably trivial change.

Am willing to submit a PR if this would be appropriate.

jxs:
Hi, can't you get those fields by downcasting via [`Rejection::find`](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs#L69) to the `Reject` type? the [example shows how to](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs#L69)

johnchildren:
> Hi, can't you get those fields by downcasting via [`Rejection::find`](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs#L69) to the `Reject` type? the [example shows how to](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs#L69)

Unless I'm missing something that will give you the struct but the fields inside the struct are private, for example:

```rust
    if let Some(err) = rejection.find::<warp::reject::MissingHeader>() {
        let missing_header = err.name; // error[E0616]: field `name` of struct `warp::reject::MissingHeader` is private
    }
```

What I would like to be able to do is get access to the `name` field, sorry if my original explanation was poor.

jxs:
>What I would like to be able to do is get access to the name field, sorry if my original explanation was poor.

ah I see, no worries, I didn't understand that sorry.  Regarding `MissingHeaders`, it implements `Display` which makes it impl `ToString` but it will convert into "Missing request header {:?}" header.name but that is not what you are looking for yeah.

Known Rejections like `MissingHeader`, `InvalidHeader`, `MissingCookie` could offer a `name()` method to access it's inner fields
CC @seanmonstar

PR #447, opened 2020-02-17
Add method to expose missing/invalid reject names

johnchildren:
Add a name method to `MissingHeader`, `InvalidHeader` and
`MissingCookie` that allow access to their internal name field. This is
useful when recovering from rejections and down-casting to return
specific replies for cases where additional information should be sent
to the client.

Resolves https://github.com/seanmonstar/warp/issues/440

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -506,6 +506,13 @@ pub struct MissingHeader {
     name: &'static str,
 }
 
+impl MissingHeader {
+    /// Retrieve the name of the header that was missing
+    pub fn name(&self) -> &str {
+        self.name
+    }
+}
+
 impl ::std::fmt::Display for MissingHeader {
     fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
         write!(f, "Missing request header {:?}", self.name)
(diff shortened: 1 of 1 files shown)


src/reject.rs
```diff
@@ -506,6 +506,13 @@ pub struct MissingHeader {
     name: &'static str,
 }
 
+impl MissingHeader {
+    /// Retrieve the name of the header that was missing
+    pub fn name(self) -> &'static str {
```
seanmonstar: I'd suggest this should just return `&str`, in case we ever needed to store a dynamic `String` in some cases.

johnchildren: Thanks, I was a bit unsure about this lifetime, but I have changed it as suggested to live as long as the rejection itself (I think). Hope this works!

seanmonstar: Oh, I didn't notice that these methods take `self` by value instead of by reference (`&self`). That should be changed, otherwise you can't call these methods from a `&MissingHeader` etc.

With that change, you don't need to specify any specific lifetime. They'll just look like `pub fn name(&self) -> &str {`.

johnchildren: Oh sorry again, I actually didn't know you could do that for some reason. Anyway have hopefully fixed it now and have squashed my commits.

PR #450, opened 2020-02-18
Generalize Reply impl for Box<dyn Reply>

mikeyhew:
We had a case where we were boxing up replies, and needed them to impl `Send + Sync` to use in other code. `Box<dyn Reply + Send + Sync>` unfortunately doesn't implement `Reply`, because the impl is too specific. This generalizes the impl of `Reply` to include any `Box<T>` where `T: Reply + ?Sized`.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -289,7 +289,7 @@ pub trait Reply: BoxedReply + Send {
     */
 }
 
-impl Reply for Box<dyn Reply> {
+impl<T: Reply + ?Sized> Reply for Box<T> {
     fn into_response(self) -> Response {
         self.boxed_into_response(Internal)
     }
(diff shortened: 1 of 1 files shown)


mikeyhew:
Looks like it is, the example below compiles without the blanket impl. Probably because `Box` has the `#[fundamental]` attribute.

```rust
struct MyThingy;

impl Reply for Box<MyThingy> {
    fn into_response(self) -> warp::reply::Response {
        todo!()
    }
}
```

`Arc`, on the other hand, isn't `#[fundamental]`, and this impl fails to compile:
```rust
impl Reply for std::sync::Arc<MyThingy> {
    fn into_response(self) -> warp::reply::Response {
        todo!()
    }
}
```

with the error message:

```
error[E0117]: only traits defined in the current crate can be implemented for arbitrary types
 --> server/src/api/products.rs:6:1
  |
6 | impl Reply for std::sync::Arc<MyThingy> {
  | ^^^^^^^^^^^^^^^------------------------
  | |              |
  | |              `std::sync::Arc` is not defined in the current crate
  | impl doesn't use only types from inside the current crate
  |
  = note: define and implement a trait or new type instead
```

PR #457, opened 2020-02-23
add ext::with_mut

AtsukiTak:
This PR adds `ext::with_mut` function which enables to access `http::Extensions` before reaching wrapped `Filter`. 
https://github.com/seanmonstar/warp/issues/222#issuecomment-491106090

I referred a `log::log` implementation.

I'm not a native speaker, so if someone correct my English on documentation I'm very happy :)

Thanks.

### DIFF
--- a/src/filters/ext.rs
+++ b/src/filters/ext.rs
@@ -4,8 +4,10 @@ use std::convert::Infallible;
 
 use futures::future;
 
-use crate::filter::{filter_fn_one, Filter};
+use crate::filter::{filter_fn_one, Filter, WrapSealed};
 use crate::reject::{self, Rejection};
+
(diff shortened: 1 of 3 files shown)


AtsukiTak:
Thanks for comment!

> I just think this could be a regular `Filter` instead of a `Wrap`.

Yes! it could be.
My concern is that how can I restrict user from accessing `Extensions` after reaching the filters? And my idea was to wrap a root filter. I thought that by requiring wrapped `Filter` to extract `Reply`, I could force user to pass root `Filter`. But it turns out not working because user can easily pass non-root `Filter`.
So I don't find any solution for now. Maybe writing some document is enough.

Would you like to give me your thoughts? Thanks!

Issue #451, opened 2020-02-19
Unreasonably hard to recover internal errors from Rejection.


tobz:
At present, for any custom recovery method, a user is forced to manually call `Rejection::find` for every possible internal error if they want to be able to properly line up the error to a suitable HTTP status code.

It would be incredibly useful if we could simply expose the internal status code for a rejection that it would otherwise assign when handling rejections using the default error handler.  This unlocks more than enough (namely, being able to call `StatusCode::canonical_reason`) as the hard work has already been mapping the known errors to a suitable HTTP status code.

Given that you actually deprecated/removed the ability to do this, can you add some color as to the _why_ and could we talk about either bringing it back or some other solution that doesn't involve a ton of boilerplate?

seanmonstar:
> Given that you actually deprecated/removed the ability to do this, can you add some color as to the *why*?

Sure! I removed the `Rejection::status` because it felt wrong in isolation. The status code it returned was only true if used by the default recovery stage. Additionally, a `Rejection` may be made of many rejections, and the status returned was selected by an internal ordering. By having the user write the `find` code, they could express which rejection types were of higher priority.

----

I admit the `Rejection` system may have rough edges, it's not a very common design. The inspiration comes from `akka-http`, with minor differences.

What's the biggest pain point at the moment? You're trying to handle certain rejections and use the default for others?

tobz:
Exactly, yeah!

You've already done the good work of expressing the common errors -- including ones that are specific to Warp's filtering design -- and I want to be able to fall back to that when the rejection I'm being handed isn't one of my own.

seanmonstar:
But you wish to tweak the response somehow beyond the default? So just returning `Err(rejection)` isn't enough?

tobz:
Ah, sorry, yes. I'm specifically spitting out JSend-compatible messages,
which is just JSON with fields that basically emulate status code/body/etc.

seanmonstar:
Would having a function available that converts the `Rejection` into its default `http::Response` be something you're looking for? And then you can modify any headers and the body?

tobz:
I think conceptually that works?  I guess either a `http::Response` or `StatusCode` are suitable for handing back as the response, but I'm probably biased here in that I really only need the status code so I can get the canonical reason.

Just to sort of rubber duck it out here, the difference would be something like...

```
async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
    let mut code = None;
    let mut message = None;

[...]

PR #458, opened 2020-02-23
Implement Reply for Result<impl Reply, impl Reply>

### LINKED ISSUE #388: Error Handling Examples
It's possible I'm just not following the rejections example well enough, but what is the correct way of indicating that a handler (called with `and_then()`) is fallible, and returning the error from that method in a reply?

For an example of the behaviour I'm trying to replicate, in Actix or Rouille, handler functions are defined as returning something resembling a `Result<GoodReply, BadReply>`, both of which can be expected to produce an HTTP response.

e.g a simple index handler which should generate a 500 if the templating fails:
```rust
async fn index(tmpl: web::Data<tera::Tera>) -> Result<HttpResponse, Error> {
    let html = tmpl
        .render("index.html", &tera::Context::new())
        .map_err(|_| HttpResponse::InternalServerError("Failed to render template")))?;

    Ok(HttpResponse::Ok().body(html))
}
```

Reading the docs for Warp, it seems like a `Rejection` should be used to indicate that this filter could not or should not accept the request, is this the case? And if so, is there an idiomatic way to replicate the behaviour I've sketched above? I can go down the road of matching on results from fallible functions in my handlers, and doing early `return warp::reply(...)`, but then I lose out on the ergonomics offered by the `?` operator inside the function.

cjbassi:
This fixes the problems I've been having with error handling as described in #388. I don't think this is a breaking change since http::Error can be added to a result and you get the same effect as before. Let me know what you think of this. I also added an example demonstrating how this would be used.

Close #388

### DIFF
--- a/examples/error_handling.rs
+++ b/examples/error_handling.rs
@@ -11,7 +11,7 @@ use warp::{Filter, Reply};
 async fn main() {
     let math = warp::path!("math" / u16)
         .and(warp::header::<u16>("div-by"))
-        .map(div_by);
+        .map_async(div_by);
 
     let routes = warp::get().and(math);
 
@@ -20,7 +20,7 @@ async fn main() {
 
 // Reply is implemented for Result<impl Reply, impl Reply>.
 // This makes it easy to handle errors since bad requests are just mapped to Result::Err().
-fn div_by(num: u16, denom: u16) -> Result<impl Reply, DivideByZero> {
+async fn div_by(num: u16, denom: u16) -> Result<impl Reply, DivideByZero> {
     let denom = NonZeroU16::new(denom).ok_or(DivideByZero)?;
 
     Ok(warp::reply::json(&Math {
--- a/examples/error_handling.rs
+++ b/examples/error_handling.rs
@@ -1,0 +1,65 @@
+#![deny(warnings)]
+
+use std::num::NonZeroU16;
+
+use serde_derive::Serialize;
+use warp::http::StatusCode;
+use warp::reply::Response;
+use warp::{Filter, Reply};
+
+#[tokio::main]
+async fn main() {
+    let math = warp::path!("math" / u16)
+        .and(warp::header::<u16>("div-by"))
+        .map(div_by);
+
+    let routes = warp::get().and(math);
+
+    warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
+}
+
+// Reply is implemented for Result<impl Reply, impl Reply>.
+// This makes it easy to handle errors since bad requests are just mapped to Result::Err().
+fn div_by(num: u16, denom: u16) -> Result<impl Reply, DivideByZero> {
+    let denom = NonZeroU16::new(denom).ok_or(DivideByZero)?;
+
+    Ok(warp::reply::json(&Math {
+        op: format!("{} / {}", num, denom),
+        output: num / denom.get(),
+    }))
+}
+
+// Error
+
+#[derive(Debug)]
+struct DivideByZero;
+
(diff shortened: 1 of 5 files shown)


cjbassi:
The one issue with this is that in the example, rejections are turned into plain text error responses instead of json but this should be easy to fix once #451 is implemented.

cjbassi:
Actually it seems like this doesn't work with async functions since map doesn't work with async functions... :(

jxs:
Hi, there's a PR in the pipeline to [`impl<T: Reject> From<T> for Rejection`](https://github.com/seanmonstar/warp/pull/396/) which will then allow to use the `?` operator on `Result<T: Reply, E: Reject>` will that help your case?
warp Filters are  similar to the [`futures api`](https://docs.rs/futures/0.3.1/futures), that's why `map` composes the filter with a function that maps the extracted value, while `and_then` composes the filter with a function that returns a `TryFuture` which can be an `async fn` that returns `Result<impl Reply, Rejection>`

cjbassi:
I hadn't seen that pull request but it turns out that I implemented the same thing in my codebase and unfortunately it did not do the trick for me since I still had to add a whole bunch of `.map_error(MyError::from)` to convert external errors to my error type first before using `?` which was pretty unergonomic.

Yah, I ended up studying the Futures API quite a bit yesterday to see if there was a way to do an async map with the current API :) But it doesn't look possible so I had to add a new method.

cjbassi:
Another idea: what if instead of having to return a Rejection struct we returned the Reject trait? That might work instead. So return values would be Result<impl Reply, impl Reject>. The Reject trait would probably have to be reworked. Idk if that would work, just throwing stuff out there.

sidju:
I am very interested in this.

Implementing Reply for Result where both implement reply seems trivial and a reasonable thing to add in general. When also combined with some kind of async .map it would suddenly enable me to write handlers as I would like, since handlers should infallibly return a response when executed and should fail fast on errors.

Of all the solutions mentioned in [https://github.com/seanmonstar/warp/issues/388](https://github.com/seanmonstar/warp/issues/388) this one seems the best.

(It seems to me that the clash with the futures API here lies in the fact that warp has declared the error type for us and given it special meaning.

It works great in the context of routing and makes writing your own filters a breeze, but it also requires the user to interact with this error type even when it is irrelevant.

This causes frustration when the ? operator could normally be used but the error type is strictly set by warp (and not recommended for returning from a handler, since it was the right route). 

Due to this (and the odd nature of request handlers compared to other "normal" async operations) I would say there is a need for a .map_async or .handler)

kaj (approved):
This looks good!  The changes in `src/reply.rs` is pretty much identical to a PR i did before seeing this (and closed as duplicate on finding this).  I'm not familiar enough with warp internals to say if the contents of `filter/map_sync.rs` are correct, but i _am_ familiar enough with _using_ warp to say that the `FilterBase::map_async` function is dearly wanted.

Of course, sitting since February there is merge conflicts to resolve, and the `map_async` function does need documentation.  Is @cjbassi still around and interested on doing that?  Or should I clone the branch, do the rebase and doc and submit a new PR?

PPakalns:
Resolved merge conflicts with the current master branch for this PR in [PPakalns/warp](https://github.com/PPakalns/warp) as I needed an up to date version. Can be used as a reference to fix conflicts in this PR or to make a new PR.

jxs:
superseeded by #878

sidju:
On the contrary, #878 doesn't implement Reply for Result<Reply, Reply>. Whilst the .map_async is indeed covered the Reply implementation may still be of relevance.

PR #472, opened 2020-03-08
Add Compression Filters

### LINKED ISSUE #46: Add compression Filters

ParkMyCar:
This PR adds compression filters to Warp, using the [`async-compression`](https://docs.rs/async-compression/0.3.1/async_compression/) crate. Compression is implemented by utilizing the fact that [`Body`](https://docs.rs/hyper/0.13.3/hyper/struct.Body.html) implements [`Stream`](https://docs.rs/futures/0.3.4/futures/stream/trait.Stream.html), and with a little bit of massaging we wrap the Response body in an encoder, and then create a new Body from that.

This implementation is definitely an MVP, there are two pitfalls a developer could fall into with this implementation
1. We don't check the `accept-encoding` header before compressing the Body, this could lead to some responses being invalid.
2. A user could theoretically double compress a response, leading to invalid responses.

From here there are a few more PRs that could be made
1. Add a `warp::compression::auto()` filter, that parses the `accept-encoding` header and selects a compression algo.
2. If `content-encoding` is already set, don't wrap the body in a another encoder, pass it through
3. For the `warp::compression::gzip(), etc.` filters, validate the `accept-encoding` header includes the proper value.
4. Add filters that allow you to set the quality of compression e.g. `warp::compression::gzip_with_quality(...)`

Because this is only an MVP, but also because this adds an additional dependency which is decently large, I put it behind a `"compression"` feature gate. When the feature is more flushed out we could remove the gate.

Resolves #46

### DIFF
--- a/src/filters/compression.rs
+++ b/src/filters/compression.rs
@@ -1,0 +1,182 @@
+//! Compression Filters
+
+use async_compression::stream::GzipEncoder;
+use bytes::Bytes;
+use futures::Stream;
+use http::header::{HeaderName, HeaderValue};
+use http::response::{Parts, Response as HttpResponse};
+use hyper::Body;
+use pin_project::pin_project;
+
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use crate::filter::{Filter, WrapSealed};
+use crate::reject::IsReject;
+use crate::reply::{Reply, Response};
+
+use self::internal::WithCompression;
+
+/// Compression
+#[derive(Clone, Copy, Debug)]
+pub struct Compression<F> {
+    func: F,
+}
+
+/// blah blah blah
+pub fn gzip() -> Compression<impl Fn(CompressionProps) -> Response + Copy> {
+    let func = move |mut props: CompressionProps| {
+        let body = Body::wrap_stream(GzipEncoder::new(props.body));
+        props.head.headers.append(
+            HeaderName::from_static("content-encoding"),
+            HeaderValue::from_static("gzip"),
+        );
+        Response::from_parts(props.head, body)
+    };
+    Compression { func }
+}
+
+/// Compression Props
+#[derive(Debug)]
+pub struct CompressionProps {
+    body: CompressableBody,
+    head: Parts,
+}
+
+impl From<HttpResponse<Body>> for CompressionProps {
+    fn from(resp: HttpResponse<Body>) -> Self {
+        let (head, body) = resp.into_parts();
+        CompressionProps {
+            body: body.into(),
+            head,
+        }
+    }
+}
+
+/// A wrapper around [`Body`](hyper::Body) that implements [`Stream`](futures::Stream) to be
+/// compatible with async_compression's Stream based encoders
+#[pin_project]
+#[derive(Debug)]
+pub struct CompressableBody {
+    #[pin]
+    body: Body,
+}
+
+impl Stream for CompressableBody {
+    type Item = std::io::Result<Bytes>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        use std::io::{Error, ErrorKind};
+
+        let pin = self.project();
+        // TODO: Use `.map_err()` (https://github.com/rust-lang/rust/issues/63514) once it is stabilized
+        Body::poll_next(pin.body, cx)
+            .map(|e| e.map(|res| res.map_err(|_| Error::from(ErrorKind::InvalidData))))
+    }
+}
+
+impl From<Body> for CompressableBody {
+    fn from(body: Body) -> Self {
+        CompressableBody { body }
+    }
+}
+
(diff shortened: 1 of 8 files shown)


src/filters/compression.rs
```diff
@@ -0,0 +1,263 @@
...
+        F::Error: IsReject,
+    {
+        type Output = Result<(Compressed,), F::Error>;
+
+        fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+            let pin = self.as_mut().project();
+            let result = ready!(pin.future.try_poll(cx));
+            match result {
+                Ok(reply) => {
+                    let resp = route::with(|_r| (self.compress.func)(reply.into_response().into()));
```
seanmonstar: Does this even need to call `route::with()`? Doesn't look like the `Route` is used.

src/filters/compression.rs
```diff
@@ -0,0 +1,263 @@
...
+            .headers
+            .append(CONTENT_ENCODING, CompressionAlgo::BR.into());
+        Response::from_parts(props.head, body)
+    };
+    Compression { func }
+}
+
+/// Compression Props
+#[derive(Debug)]
+pub struct CompressionProps {
```
seanmonstar: Does this need to be public?

src/filters/compression.rs
```diff
@@ -0,0 +1,263 @@
...
+            head,
+        }
+    }
+}
+
+/// A wrapper around [`Body`](hyper::Body) that implements [`Stream`](futures::Stream) to be
+/// compatible with async_compression's Stream based encoders
+#[pin_project]
+#[derive(Debug)]
+pub struct CompressableBody {
```
seanmonstar: Does this need to public?

src/filters/compression.rs
ParkMyCar: Good catch, definitely does not need to be, updated

src/filters/compression.rs
ParkMyCar: It does, because CompressionProps is used in the internal module, but also is exposed externally via the return type of the filter functions, e.g. `gzip() -> Compression<impl Fn(CompressionProps) -> Response + Copy>`

src/filters/compression.rs
ParkMyCar: It does not, updated!

ParkMyCar:
Thanks!

I responded to your comments and updated where necessary. I also realized `CompressableBody` could be more generic since its single `body` field really only needed to impl `futures::Stream`, let me know what you think.

Definitely agree most users would just want `warp::compression::auto()` and they could add it after they `.or(...)` together all of their filters. I figured first adding these individual filters would be good building blocks for the `auto()` filter. That being said, for parsing the value of the `accept-encoding` header, do you know if [`hyper`](https://docs.rs/hyper/0.13.3/hyper/), [`http`](https://docs.rs/http/0.2.0/http/) or something similar provides a good way to do this? Normally I think you could split on `','` but that would break when trying to handle quality value syntax

btw love the framework, especially the Filter design, feels like such a natural way to handle requests. Thanks for building it 😊

src/filters/compression.rs
seanmonstar: Hm, what if this were moved to the `mod internal` then, since the argument to the closure is just an implementation detail...

ParkMyCar: I moved `CompressionProps` into `mod internal` (and `CompressableBody` because it seemed to make sense), but this gets into a part of Rust I don't know about, and thats module visibility.

`CompressionProps` still needed to be public because of the return type of the `gzip() -> ...` and such, but does `CompressionProps` being in a private module protect its visibility?
Also I totally could have missed the point here, and maybe what I did makes no difference?

src/lib.rs
```diff
@@ -103,6 +103,8 @@ mod transport;
 
 pub use self::error::Error;
 pub use self::filter::Filter;
+#[cfg(feature = "compression")]
```
seanmonstar: Didn't notice, can you mark this `doc(hidden)` and move it below the comments why?

seanmonstar:
> do you know if hyper, http or something similar provides a good way to do this? Normally I think you could split on ',' but that would break when trying to handle quality value syntax.

Ideally, the [headers](https://github.com/hyperium/headers) would grow that functionality.

ParkMyCar:
> > do you know if hyper, http or something similar provides a good way to do this? Normally I think you could split on ',' but that would break when trying to handle quality value syntax.
> 
> Ideally, the [headers](https://github.com/hyperium/headers) would grow that functionality.

Good looks, I'll open a PR soon against headers to add such functionality

src/lib.rs
ParkMyCar: Updated and moved

seanmonstar:
Woo, thanks!

PR #478, opened 2020-03-11
Add XML support for body & reply, as well as strict content-type checking options for easier filter composition

m-doughty:
This pull request implements `warp::reply::xml()` and `warp::body::xml()` to match JSON functionality, as well as introducing strict type checking on content-type to allow for easier composition.

I've also added tests for XML & XML/JSON/Form with strict content types.

I would consider this a WIP because I couldn't find an example of how `warp::reply` is currently being tested, so I'd like to add tests for that as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ scoped-tls = "1.0"
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.6"
-serde-xml-rs = { git = "https://github.com/RReverser/serde-xml-rs" }
+serde-xml-rs = "0.4.0"
 tokio = { version = "0.2", features = ["blocking", "fs", "stream", "sync", "time"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ scoped-tls = "1.0"
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.6"
+serde-xml-rs = { git = "https://github.com/RReverser/serde-xml-rs" }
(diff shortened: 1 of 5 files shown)


seanmonstar:
Thanks for the PR!

My first thoughts are that XML is becoming less and less common, so I'm not sure about adding it for everyone. Can this be an external thing?

m-doughty:
Completely agree that XML is becoming less and less common, though in the public & financial sectors it's very much alive and kicking due to dinosaur versions of JVM and general unwillingness to retire existing services.

It probably could be a separate thing, but a lot of the underlying convenience methods & structs are only crate-visible so it'd be a lot more implementation (and largely re-implementation). I suspect we will just have to maintain a fork if xml isn't wanted upstream (it certainly wont be the only library we've forked for legacy compatibility). :)

ParkMyCar:
@m-doughty just for reference, I implemented a similar filter for Protobuf, moved it to an external crate here: https://github.com/ParkMyCar/warp-protobuf 😃

gzsombor:
There are a couple of well established protocol which still uses XML - for example WebDav, DLNA, OPDS, RSS/Atom - definitely, their usage is declining. What about adding a feature flag for this?

m-doughty:
> There are a couple of well established protocol which still uses XML - for example WebDav, DLNA, OPDS, RSS/Atom - definitely, their usage is declining. What about adding a feature flag for this?

I think this is probably the best way to handle it tbh, I'll do some reading into how to create feature flags and get back to you.

seanmonstar:
My personal feeling continues to be that this could be an external thing.

>  a lot of the underlying convenience methods & structs are only crate-visible so it'd be a lot more implementation (and largely re-implementation)

I think we should try to make this stuff somehow available, so other formats can more easily be extra libraries and not have to re-implement the same core pieces!

PR #482, opened 2020-03-12
Allow the max frame size to be set for a websocket

rklaehn:
We had a problem in production with an overly large websocket message from chrome not being received in a warp based service. To solve this, we had to fork warp and allow the max_frame_size to be set.

We are using 0.1.x in this system, but if this is accepted we could also do an additional PR against master.


seanmonstar:
LGTM!

PR #485, opened 2020-03-13
Allow the max frame size to be set for a websocket

bantonsson:
Forward port of #482


seanmonstar:
Thanks! (Also just published v0.1.22 with #482).

Issue #432, opened 2020-02-02
Add filter to get the host or :authority


Restioson:
It appears that when using TLS with Warp, the HTTP Host header is stripped. Therefore, `warp::header<String>("host")` on a TLS server will fail.

Ironically, the Haskell framework with the same name had an identical bug: https://github.com/yesodweb/wai/issues/478

jxs:
I don't think this is a bug, when warp is serving using TLS, clients offer Http2 using ALPN, hyper then upgrades the connection to Http2, and per the Http2 [spec](https://http2.github.io/http2-spec/):
>Clients that generate HTTP/2 requests directly SHOULD use the :authority pseudo-header field instead of the Host header field.

accessing the host via [`Uri::authority()` ](https://docs.rs/hyper/0.13.2/hyper/struct.Uri.html#method.authority) doesn't work for you?
you can get it constructing an `Uri` with [`warp::path::full()`](https://docs.rs/warp/0.2.1/warp/filters/path/fn.full.html)

Restioson:
Thanks, I will try both when I get home.

Restioson:
I first tried extracting `:authority` and `authority` using the `warp::header` method, to be greeted by identical behaviour of requiring `host`. For reference, here are my Firefox raw HTTPS request headers:

```
Host: 127.0.0.1:8080
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:71.0) Gecko/20100101 Firefox/71.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8
Accept-Language: en-US,en;q=0.5
Accept-Encoding: gzip, deflate, br
Connection: keep-alive
Upgrade-Insecure-Requests: 1
Cache-Control: max-age=0
TE: Trailers
```

Other info:
```
Request URL:https://127.0.0.1:8080/invite/4XwS5ZEoVHU
Request Method:GET
Remote Address:127.0.0.1:8080
Status Code:
400
Version:HTTP/2.0
```

However, `warp::path::full()` does work! Thanks!

Restioson:
Oops, nevermind. It looks like `warp::path::full` just returns the path rather than the entire URI (makes sense, given its name).

jxs:
ah! you are right! `warp::path::full` doesn't return the `Authority` 
hum, this could solved by making [`FullPath`](https://github.com/seanmonstar/warp/blob/master/src/filters/path.rs#L420) be 
[...]

PR #489, opened 2020-03-19
Implement host filter for parsing the host from the authority (http/2) or parsing the host header (http/1.x)

dmeijboom:
Related: https://github.com/seanmonstar/warp/issues/432

This is the initial version of parsing the host properly in Warp. First, it tries to parse the hostname based on the `:authority` pseudo-header. If that fails it tries to parse the `Host` header for http/1.1 and http/1.

I'm aware that this is probably not the final version as it's using a `Option<String>` where in the issue there was a mention about a `Host` type. Since I'm not an expert in Rust please advice what's missing or how it could be improved.

Thanks!

### DIFF
--- a/src/filters/host.rs
+++ b/src/filters/host.rs
@@ -1,0 +1,25 @@
+//! Host filter
+//!
+use futures::future;
+use crate::reject::{self, Rejection};
+use crate::filter::{Filter, filter_fn, one};
+
+/// Get the host of the route
+pub fn host() -> impl Filter<Extract = (Option<String>,), Error = Rejection> + Copy {
+    let name = "Host";
+
+    filter_fn(move |route| future::ok(one(
+        route.uri()
+            .authority()
+            .and_then(|authority| Some(authority.host()))
+            .and_then(|host| Some(host.to_string()))
+            .or_else(|| route.headers()
+                .get(name)
+                .ok_or_else(|| reject::missing_header(name))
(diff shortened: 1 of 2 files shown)


dmeijboom:
> Excellent first steps, thank you!
> 
> > Probably return some newtype `Host`, wrapping a `http::uri::Authority`
> 
> Would you be able to create a newtype, `pub struct Host(Authority)`, that can be returned instead of a `String`?

Yes. I added the `Host` type but I'm still struggling with handling the error properly. Does this actually work or will it silently ignore the `.map_err(|_| reject::invalid_header(name))` stuff because it's being transformed to an `Option<Host>`?

seanmonstar:
Best way to find out is to write some tests. :D

jakajancar:
Do I understand correctly that right now, if you want to do "virtual hosting" / support multiple domains using Warp, you're out of luck:

  - There is no way to get the authority header
  - There is no way to disable HTTP/2

?

jakajancar:
Are all that is missing here tests, or are there other objections?
I can add them if needed.

//cc @jxs

dmeijboom:
> Are all that is missing here tests, or are there other objections?
> I can add them if needed.
> 
> //cc @jxs

Iirc the only thing left was making sure it actually works and writing some tests. I need this as well but unfortunately haven't had the time to finish this :sweat_smile:

PR #500, opened 2020-03-23
Improve the Handlebars example

Draphar:
The boilerplate is unnecessary.


seanmonstar:
Thanks for the PR!

While this is certainly a clever way to do it, it's probably better to keep "the boilerplate" in this example, as it's really just meant to explain how to do something. More advanced users can certainly make things shorter if they understand better.

PR #505, opened 2020-03-25
Add warp::header::value() filter to HeaderValue

seanmonstar:


seanmonstar:
What do you mean "parse a `HeaderValue`?

This PR is to add a sort of "raw" header filter, to get the value without any parsing.

ParkMyCar:
My bad, I forgot that `HeaderValue` is more or less a wrapper around `Bytes`. I was thinking if `HeaderValue` impl'd `FromStr` then `warp::header::value(...)` == `warp::header::<HeaderValue>(...)` but that's not the case 🙂

On a related, but separate note, what would you think about a filter like:
```
use bytes::BufMut;
use std::io::Write;

pub fn raw(
  name: &'static str,
) -> impl Filter<Extract = One<impl [BufMut | Write]>, Error = Rejection> + Copy {...}
```

PR #516, opened 2020-04-01
Add RequestBuilder::with_remote_addr.

skywhale:
I needed this to be able to write unit tests for filters that use `filters::addr::remote()`.


skywhale:
Deleting this in favor of https://github.com/seanmonstar/warp/pull/517

PR #520, opened 2020-04-03
add example showing how to serve a react-style SPA using warp

Melvillian:
I spent around a day trying to figure out how to serve a single page application via `warp` and once I figured it out I thought I'd add an example for it since i figured I'm not the only one looking to use warp for react-style apps.

This PR is not mergeable as is, because this example is still not loading the `index.js` javascript correct from the `index.html`. I need someone more familiar with `warp` to tell me what needs to change here before we can add this example.

Note: I was able to get this example to work by changing `main.js` so there is 1 route for the `index.html` and ` route for the `index.js`. However this is just a kludgy workaround because if I want to add another asset for the browser to fetch (e.g. `index.css`) then I have to go and add an additional route. I want to be able to just expose a single `public/` directory and then the browser can fetch any assets it contains.

### DIFF
--- a/examples/html-with-js/main.rs
+++ b/examples/html-with-js/main.rs
@@ -1,0 +1,18 @@
+#![deny(warnings)]
+use warp::Filter;
+
+#[tokio::main]
+async fn main() {
+    // expose all of the files in public/
+    let public = warp::path!("examples" / "html-with-js" / "public")
+        .and(warp::fs::dir("examples/html-with-js/public"));
+
+    // GET / -> index html
+    let index = warp::get()
+        .and(warp::path::end())
+        .and(warp::fs::file("examples/html-with-js/public/index.html"));
+
+    let routes = index.or(public);
+
+    warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
+}
--- a/examples/html-with-js/public/index.html
+++ b/examples/html-with-js/public/index.html
@@ -1,0 +1,18 @@
+<!doctype html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <title>html + javascript served at warp speeds</title>
+</head>
+
+<body>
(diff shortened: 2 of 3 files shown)


ParkMyCar:
I'm not positive, but it looks like the `path!` [macro eventually finishes](https://github.com/seanmonstar/warp/blob/master/src/filters/path.rs#L558) with a `warp::path::end()` which means your route `examples/html-with-js/public` won't match `examples/html-with-js/public/...` or more specifically `examples/html-with-js/public/index.js`

What you could do is
```rust
// Note the non-macro path
let public = warp::path("public").and(warp::fs::dir("examples/html-with-js/public"));
```

```html
<script src="public/index.js"></script>
```

I've also created a very similar example [here](https://github.com/ParkMyCar/warp-examples/tree/master/react) if you wanted to take a look 😄

jxs:
Hi, thanks for your interest, and sorry for the delay! We already have a quite significant number of examples so it might become confusing adding more, but I think this might make sense as an upgrade of the [handlebars templating example](https://github.com/seanmonstar/warp/blob/master/examples/handlebars_template.rs) what do you think?

hamza1311:
> I think this might make sense as an upgrade of the [handlebars templating example](https://github.com/seanmonstar/warp/blob/master/examples/handlebars_template.rs) what do you think?

Handlebars is templates which renders the markup on server so I don't think it makes sense to include a SPA in that example. Static files example seems like a better fit for this, imo.

jxs:
>Handlebars is templates which renders the markup on server so I don't think it makes sense to include a SPA in that example. Static files example seems like a better fit for this, imo.

Ah right! the problem I see with the file example and this one replacing it is that this is using `fs::file` redundantly cause `fs::dir` already returns `index.html`  so should get a way to also showcase `fs:file`

PR #527, opened 2020-04-08
Breaking Change: replace `Reject` trait with `Into<Rejection>`

yaahc:
I'm not sure this is the best approach yet but the goal rn it to make it so warp can support 3rd party error types like `anyhow` and `eyre` without those types having to have a conditional dependency on warp just to impl the Reject trait.

### DIFF
--- a/examples/rejections.rs
+++ b/examples/rejections.rs
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 
 use std::convert::Infallible;
+use std::fmt;
 use std::num::NonZeroU16;
 
 use serde_derive::Serialize;
@@ -38,8 +39,12 @@ fn div_by() -> impl Filter<Extract = (NonZeroU16,), Error = Rejection> + Copy {
 
 #[derive(Debug)]
 struct DivideByZero;
-
-impl reject::Reject for DivideByZero {}
+impl std::error::Error for DivideByZero {}
+impl fmt::Display for DivideByZero {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "cannot divide by zero")
+    }
+}
 
 // JSON replies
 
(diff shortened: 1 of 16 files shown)


yaahc:
## do not merge

This trait doesn't yet solve the problem I care about, in fact it makes it worse, I still haven't been able to figure out how to have `Rejection` support wrapping any type that other than itself that is `Debug + Sized + Send + Sync + 'static`

The best idea I have rn is to just drop the `Debug` impl on Rejection and expose its debug impl through a member function, which would allow it to impl `From<Debug + Sized + Send + Sync + 'static>` without overlapping with `From<T> for T`, and then the from impl could be used whenever we create a custom, which would prevent it from double wrapping itself

seanmonstar:
One thing to note is that the rejection system in warp is purposefully *not* meant to be `std::error::Error`. The extra trait is to make one think of meaningful rejections, that you then look for in `recover`.

yaahc:
well that works out because basing it on std error was not possible without specialization afaik. The "remove Debug on Rejection" solution ended up working, bonus points is now when type inferences lets you you can use `?` to go straight into a `Rejection`

yaahc:
Got it working and passing tests, not being able to `unwrap()` a `Rejection` because it no longer impls Debug sure is annoying in test code :/

seanmonstar:
Thanks for the PR! I'm inclined to keep the `Reject` trait separate for now, as that design was on purpose (more can be found in the related PRs and commits). If that purpose turns out to be flawed in practice, we could just resort to `std::error::Error`, but I'd suggest we have a discussion with the reasons why that's better in an issue first.

yaahc:
That's fair, I probably wont push on this because I was only using warp for a small side project that ended up getting discarded but if someone else ends up asking about it or said `Reject` trait bikeshed starts happening please let me know and ill pop in and explain/defend the changes in this PR and why I think they're helpful.

PR #529, opened 2020-04-09
Bump LICENSE year

coolreader18:


x448:
Why replace copyright year "2018" with "2020"?  

Isn't it safer to replace "2018" with "2018-2020" if there's any content remaining from before 2020?

coolreader18:
Oh, I didn't think of that, I see it more as "this work has been claimed as copyrighted since this year", but I can definitely make it a range.

x448:
> Oh, I didn't think of that, I see it more as "this work has been claimed as copyrighted since this year", but I can definitely make it a range.

I used to think the same.  Unfortunately, if only 1 year is specified then it's treated as the earliest year.

PR #537, opened 2020-04-15
Add explicit error to WebSocket Message `to_str` method

kellytk:
I used guidance from https://github.com/seanmonstar/warp/blob/497505a50508ec15dbfdc984c24026f957f28738/src/tls.rs#L165 and https://github.com/seanmonstar/warp/blob/fbbbc5b8377914463e4411ef6fc48bfd97d15a6d/src/filters/ws.rs#L196.  Please inform me if it can be improved.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -3,6 +3,7 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::future::Future;
+use std::io::{self};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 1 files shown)


kellytk:
Apparently the "compression" test failed.  I'm unsure of how to proceed as the relevant section at https://github.com/seanmonstar/warp/runs/590111167?check_suite_focus=true displays "We are currently unable to download the log. Please try again later."

jxs:
Hi, sorry for the delay and thanks for your interest!  
I would not consider this an `io::Error` since it's used for  I/O operations sucj as  `Read`, `Write`, `Seek`. Also since this might be a breaking change what is the use case you need it for?

PR #540, opened 2020-04-16
Add Log constructor

hlb8122:
**Motivation**
Without a constructor and with the `func` field private, it is impossible to use the `Log` struct in order to implement new logging schemes or diagnostics.

There is a large amount of useful implementation surrounding `Log` which users could harness. Both `Log` and `Info` are public structs.

### DIFF
--- a/src/filters/log.rs
+++ b/src/filters/log.rs
@@ -83,6 +83,13 @@ pub struct Log<F> {
     func: F,
 }
 
+impl<F> Log<F> {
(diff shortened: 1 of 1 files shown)


hlb8122:
Whoops! Missed this https://docs.rs/warp/0.2.2/warp/filters/log/fn.custom.html

Feel free to close.

seanmonstar:
No problem :)

PR #563, opened 2020-05-03
Add filter method flatten()

naerbnic:
At this point in time, there isn't anyway to compose filters dynamically. You can decide to append dynamic async code to the end of a filter chain through `.and_then()`, but all the implementation after that point must be known before the request is being processed.

This function allows you to compose a function on the end of a filter which itself yields a new filter. This composition has the extract/error types of the yielded filter, and can be further composed.

FWIW, I'm not 100% sure that "flatten" is the right name, given that in futures, this is applied to a Future that yields a Future. I can change the implementation to do that (and then just use `and_then()` to do the composition otherwise, but I think the demand for something in this category is there.

My use case came up when I was trying to implement a session middleware in pure warp. This wasn't possible, because I couldn't create a filter that could obtain the session data, and pass into the next stage of a filter.

### DIFF
--- a/src/filter/flatten.rs
+++ b/src/filter/flatten.rs
@@ -5,7 +5,7 @@ use std::task::{Context, Poll};
 use futures::{ready, TryFuture};
 use pin_project::{pin_project, project};
 
-use super::{Filter, FilterBase, Func, Internal, CombineRejection};
+use super::{CombineRejection, Filter, FilterBase, Func, Internal};
 
 #[derive(Clone, Copy, Debug)]
 pub struct Flatten<T, F> {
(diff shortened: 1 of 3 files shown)


M-J-Hooper:
This reminds me of `Iterator`s [`flat_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map) if that helps with naming.

jxs:
Hi, and sorry for the delay! In the meantime this can now be accomplished with [`wrap_fn`](https://docs.rs/warp/0.3.1/warp/fn.wrap_fn.html) right?

PR #588, opened 2020-05-26
[PoC] [RFC] Allow routes to return anyhow::Error

thomaseizinger:
Using anyhow allows for really ergonomic error handling in Rust code. Unfortunately, these ergonomics end at the integration layer with warp. Currently, we work around this by having separate handler functions that return an `anyhow::Result` which we then pack into a rejection (we also have an intermediary step for `HttpApiProblem` but that is not important at this stage):

https://github.com/comit-network/comit-rs/blob/301eb63afea6327df8eb7281bc8251090cc120a3/cnd/src/http_api/routes.rs#L30-L37

I am putting this PR up as a proof-of-concept to show that we could improve ergonomics here fairly easily :)


jxs:
Hi, this has already been discussed, see https://github.com/seanmonstar/warp/pull/527, https://github.com/seanmonstar/warp/pull/374

thomaseizinger:
> Hi, this has already been discussed, see #527, #374

I am aware of these discussions, I was even part of some of them :)
The reason I am opening this PR is because in my opinion this was never properly addressed.
Ironically, https://github.com/seanmonstar/warp/issues/307 was just brought to my attention again so I am gonna try and continue the discussion there!

PR #596, opened 2020-06-01
add tungstenite::Error::ConnectionClosed guard to WebSocket 

### LINKED ISSUE #593: "Connection closed normally" WebSocket error
When I use a WebSocket as a Stream, a regularly-closed-by-the-client connection will end with a ConnectionClosed "error" from tungstenite; this creates noise in my error logs suggesting that something has gone wrong when it hasn't.

It would be convenient if that error ended the Stream instead of being exposed as an opaque WarpError; it appears that [used to be the case](https://github.com/seanmonstar/warp/commit/9dacd7d9977e7894aada7ea4cf71754aa41e1033#diff-9119b6774ba273de868213a966affe87L292), but I couldn't determine if it was intentionally changed.

jxs:
Stream::poll_next match on WebSocketStream
fixes #593

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -13,10 +13,9 @@ use crate::reply::{Reply, Response};
 use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt};
 use headers::{Connection, HeaderMapExt, SecWebsocketAccept, SecWebsocketKey, Upgrade};
 use http;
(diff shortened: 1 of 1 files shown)


jxs:
[tokio-tungstenite already deals with this](https://github.com/snapview/tokio-tungstenite/blob/master/src/lib.rs#L282)

PR #597, opened 2020-06-02
Update to tokio-rustls 0.13.1

mzabaluev:
Warp is an outlier in our dependency tree with its older version of tokio-rustls.
This update was trivial and passes `cargo test --all-features`.


mzabaluev:
NB: [Dependabot] is an awesome tool to automate chores like this.

[Dependabot]: https://help.github.com/en/github/administering-a-repository/about-github-dependabot

jxs:
thanks! :+1:

PR #606, opened 2020-06-13
Clarify websockets chat comments + optimizations

### LINKED ISSUE #578: Fix comments in websockets_chat example
https://github.com/seanmonstar/warp/blob/a6fcf76d43e0b3a11f445aacf638094e7dcd008b/examples/websockets_chat.rs#L105-L117

This has a few problems:
* "We use `retain` instead of a for loop so that we can reap any user that appears to have disconnected." is the exact opposite of what actually happens in code and the comment just 3 lines later.
* As a result, `iter_mut` is unnecessary
  * We could then probably replace the `Mutex` in that demo with `RwLock` because you don't need exclusive access to send messages

lights0123:
Closes #578

* Switch to `tokio::RwLock`
* Remove outdated comments
* JS: use `getElementById`, remove behavior that ["will lead to brittle code"](https://html.spec.whatwg.org/#named-access-on-the-window-object)
* JS: show message when disconnected

### DIFF
--- a/examples/websockets_chat.rs
+++ b/examples/websockets_chat.rs
@@ -6,7 +6,7 @@ use std::sync::{
 };
 
 use futures::{FutureExt, StreamExt};
-use tokio::sync::{mpsc, Mutex};
+use tokio::sync::{mpsc, RwLock};
 use warp::ws::{Message, WebSocket};
 use warp::Filter;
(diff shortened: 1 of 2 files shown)


jxs:
thanks!

PR #608, opened 2020-06-14
Adds negation as an addition method of composition for `Filter`

M-J-Hooper:
Adds a `negate` method to the `Filter` trait.

A filter is kind of like a `Predicate` on a request (kind of, ignoring return and error types). In other languages predicates can also be composed with `and` and `or` but also often have a `not` or `negate` method too.

We need to care about extraction and rejection, so here `negate` somewhat feels like the opposite of `recover`. It accepts a function to map the extraction items to a rejection in the negated version.

This is my first PR and I'm fairly new to Rust and Warp so open to suggestions, feedback, etc. 

(I did originally have this method as `invert` which is very similar but has an `Extract` equal to the `Error` of the original filter. Thoughts?)


jxs:
Hi, and thanks for your interest! Can you describe a usecase for `negate`?

M-J-Hooper:
I did put a basic use-case in the method doc but probably should have put it in the description. That example is negating a header filter to ignore requests with a certain header. Not really sure how exactly is can be used but it felt in line with the general composability of filtering. Feel free to close this if you think its not useful enough to make the cut :)

apiraino:
@M-J-Hooper Hi, just a comment from a bystander. I like the use case of blacklisting some headers and rejecting the request (example: an old API unsupported API). Could that be achieved with something like this (pseudo-code, not tested)?
```
fn is_old_api_user (
) -> impl Filter<Extract = ((),), Error = warp::Rejection> + Clone {
    warp::header::optional::<String>("X-Api-v1")
        .and_then(move |header_name: Option<String>| async move {
            match header_name {
                Some(_) => {
                        // Old APIs not supported anymore
                        return Err(warp::reject());                
               }
                _ => {
                    // ok fine, go ahead
                    Ok(())
                }
            }
        })
}
```

M-J-Hooper:
@apiraino Yes it can be done fairly easily like that too, although with `negate` its much more concise: 
```rust
fn is_old_api_user() -> impl Filter<Extract = ((),), Error = warp::Rejection> + Clone {
    warp::header::optional::<String>("X-Api-v1").negate(|_| warp::reject())
}
```
But I see how it is quite niche. I just like the idea of having a `not` when there is `and` and `or` 😄

hamza1311:
Another possible use case would be to return `index.html` for non-API routes (necessary while serving Single Page Applications). For example:
```rust
warp::path!("api" / ..).negate(|_| warp::reject()).and(warp::fs::file("index.html"))
```

PR #618, opened 2020-06-21
use html heart character and fix compilation problem

sternix:
i think use &hearts; character may be too meaningful;
and the code in the example section of reply.rs can't compiled,


jxs:
thanks!

PR #623, opened 2020-06-24
document which websocket messages need handling

### LINKED ISSUE #620: Outdated documentation on websocket Message
[`warp::filters::ws::Message`](https://docs.rs/warp/0.2.3/warp/filters/ws/struct.Message.html) claims that it only represents `Text` or `Binary`, but the message can also be `Close`.

I thought this meant that `Close` from clients was handled internally, but this is not the case.

euclio:
Fixes #620.


jxs:
thanks!

PR #633, opened 2020-07-02
Query string example

ayngling:


examples/query_string.rs
```diff
@@ -0,0 +1,35 @@
...
+async fn main() {
+    pretty_env_logger::init();
+
+    // get /example1?key=value
+    let example1 = warp::get()
+        .and(warp::path("example1"))
+        .and(warp::query::<HashMap<String, String>>())
+        .map(|p: HashMap<String, String>| match p.get("key") {
+            Some(key) => Response::builder().body(format!("key = {}", key)),
+            None => Response::builder().body(String::from("No param in query.")),
```
jxs, line 21: can we set the status to `BAD_REQUEST`?

ayngling: No problem at all. But, are we sure we want to consider the absence of a query param an error here? Some params are optional, which is why I did not consider it a bad request in the example.

ayngling: As an alternative, maybe the text should say "key = <missing>" instead.

Just a thought. Let me know and I'll change it, I have no strong opinion either way.

jxs: > Some params are optional, which is why I did not consider it a bad request in the example.

ah i see! can we then leave a comment that its optional?
btw, can we also add a third example showcasing how to deal with absent/failed to decode query parms, like:
```Rust
    let opt_query = warp::query::<MyObject>()
        .map(Some)
        .or_else(|_| async {
            Ok::<(Option<MyObject>,), std::convert::Infallible>((None,))
        });

    // get /example3?key1=value,key2=42
    let example3 = warp::get()
        .and(warp::path("example3"))
        .and(opt_query)
        .map(|p: Option<MyObject>| {
            match p {
                Some(obj) => Response::builder().body(format!("key1 = {}, key2 = {}", obj.key1, obj.key2)),
                None => Response::builder()
                .status(StatusCode::BAD_REQUEST)
                .body(String::from("failed to decode query param")),
            }
        });
```

ayngling: Good plan. I pushed some changes, including your example3. I formatted it a bit so it resembles example two, to make it more obvious to readers that it builds on example2 with custom error handling. I also added comments.

I hope you agree that it makes it clearer. If not, just let me know :)

examples/query_string.rs
```diff
@@ -13,23 +16,42 @@ async fn main() {
...
             Response::builder().body(format!("key1 = {}, key2 = {}", p.key1, p.key2))
         });
 
-    warp::serve(example1.or(example2))
+    // get /example3?key1=value,key2=42
+    // builds on example2 but adds custom error handling
+    let example3 =
+        warp::get()
+            .and(warp::path("example3"))
+            .and(warp::query::<MyObject>().map(Some).or_else(|_| async {
```
jxs, line 42: can we isolate the optional query param, so that it's easier to read?

ayngling: Sure thing, I have submitted an update.

ayngling:
My pleasure!

Issue #638, opened 2020-07-06
Unseal Wrap trait


L-as:
I don't understand why the Wrap trait has to be sealed.

seanmonstar:
It was sealed in case the design of the trait needed to evolve. For now, since it doesn't seem like things need to really change, we can probably unseal it.

L-as:
Thanks! I also see a lot of other potentially useful but private code in the crate, like `filter_fn`. Are these other things always going to be private, or are they potentially going to be made public at some point?

jxs:
see https://github.com/seanmonstar/warp/issues/607#issuecomment-650448029

aujxn:
@jxs Should this be closed after #693 and #842 merge?

jxs:
yup, makes sense thanks!

PR #640, opened 2020-07-08
Unseal Wrap trait

hgzimmerman:
Fixes https://github.com/seanmonstar/warp/issues/638.

The doc comments for the `Wrap` trait are not suitable for merging at the moment. Any feedback related to them would be appreciated.

### DIFF
--- a/src/filter/wrap.rs
+++ b/src/filter/wrap.rs
@@ -1,14 +1,14 @@
 use super::Filter;
 
-pub trait WrapSealed<F: Filter> {
+pub trait Wrap<F: Filter> {
     type Wrapped: Filter;
 
     fn wrap(&self, filter: F) -> Self::Wrapped;
 }
 
-impl<'a, T, F> WrapSealed<F> for &'a T
+impl<'a, T, F> Wrap<F> for &'a T
 where
-    T: WrapSealed<F>,
+    T: Wrap<F>,
     F: Filter,
 {
     type Wrapped = T::Wrapped;
@@ -17,11 +17,11 @@ where
     }
 }
 
-pub trait Wrap<F: Filter>: WrapSealed<F> {}
-
-impl<T, F> Wrap<F> for T
-where
-    T: WrapSealed<F>,
-    F: Filter,
-{
-}
+// pub trait Wrap<F: Filter>: WrapSealed<F> {}
+//
+// impl<T, F> Wrap<F> for T
+// where
+//     T: WrapSealed<F>,
+//     F: Filter,
+// {
+// }
--- a/src/filter/wrap.rs
+++ b/src/filter/wrap.rs
@@ -19,4 +19,3 @@ where
         (*self).wrap(filter)
     }
 }
-
--- a/src/filter/wrap.rs
+++ b/src/filter/wrap.rs
@@ -1,8 +1,11 @@
 use super::Filter;
 
+/// Wraps filters.
 pub trait Wrap<F: Filter> {
+    /// The type of the Filter produced by wrapping another Filter.
     type Wrapped: Filter;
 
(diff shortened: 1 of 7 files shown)


src/filter/wrap.rs
```diff
@@ -1,27 +1,21 @@
 use super::Filter;
 
-pub trait WrapSealed<F: Filter> {
+/// Wraps filters.
```
jxs, line 3: can we give a detailed explanation of the trait, in the line of others like [reply](https://github.com/seanmonstar/warp/blob/master/src/reply.rs#L198) with a working example?

hgzimmerman: I'm struggling to come up with an example use case. Most existing uses of the `Wrap` trait use some sealed structure, so I'm not sure what would make sense to demonstrate here.

jxs: it's getting nice! can you show where you are stalled? I think you can use `BoxedFilter`

hgzimmerman: I'm struggling to come up with a terse, but plausible use for `Wrap`. I look to something like `warp::filters::log::Log` or `warp::filters::reply::WithHeader` and see that their `Wrap` implementation is relying on private items. I don't think `BoxedFilter`s will make this task easier for items that are like `Log` or `WithHeader`

The ultimate motivation behind me wanting to unseal this trait is to be able to implement something like what is detailed in this comment: https://github.com/seanmonstar/warp/pull/408#issuecomment-606325196, but I see that that also relies on other private items.

So its a problem of 1. I don't know what the example should demonstrate. and 2. I don't know if a plausible example could be created with the other "primitives" that are exposed currently.

hgzimmerman:
Ok. So I finally figured out what you meant by using `BoxedFilter`. 
I don't quite have it working yet due to some trait constraints, the errors of which should be present in the CI system, but I'll post the most notable ones here:
```rust
---- src/filter/wrap.rs - filter::wrap::Wrap (line 18) stdout ----
error[E0277]: the trait bound `<<F as warp::filter::FilterBase>::Extract as warp::generic::Tuple>::HList: warp::generic::Combine<_>` is not satisfied
##[error]  --> src/filter/wrap.rs:33:14
   |
18 |             .and(warp::any().map(|| log::info!("")))
   |              ^^^ the trait `warp::generic::Combine<_>` is not implemented for `<<F as warp::filter::FilterBase>::Extract as warp::generic::Tuple>::HList`

error[E0277]: the trait bound `std::convert::Infallible: warp::reject::sealed::CombineRejection<<F as warp::filter::FilterBase>::Error>` is not satisfied
##[error]  --> src/filter/wrap.rs:33:14
   |
18 |             .and(warp::any().map(|| log::info!("")))
   |              ^^^ the trait `warp::reject::sealed::CombineRejection<<F as warp::filter::FilterBase>::Error>` is not implemented for `std::convert::Infallible`
```
Any advice on where to go from here would be appreciated.

jxs:
sorry for the late reply, and thanks for your effort! yeah this solution seems it would always need to unseal other traits which would lead to more confusion, and in the end even if it worked one would always need to impl `Wrap` and probably `Future` which would not be the way one works with `Filter`s.
I have since posted an update on https://github.com/seanmonstar/warp/issues/16#issuecomment-658750842 can you give your opinion there?
Thanks

hgzimmerman:
Should this PR be closed in favor of your `wrap_fn` proposal?

hgzimmerman:
Is there anything I can do to assist in getting your `wrap_fn` included into Warp?

jxs:
@hgzimmerman yeah we can close this one in favor of the `wrap_fn` proposal, which i am awaiting for feedback on the issue.
Thanks

Issue #650, opened 2020-07-15
Custom handling for BodyDeserializeError


zupzup:
Hey!

Thanks for this fantastic project, I've been loving it so far.

I have a question regarding the `.body()` filter with `json`. If the deserialization fails, a Rejection with `BodyDeserializeError` is triggered, which can be handled with `.recover`, so far, so good.

However, if I would now like to create, say, a nice custom, helpful error based on the underlying `serde` error and ran into the problem that `BodyDeserializeError` hides the `cause` and only provides a hard-coded error message with the stringified cause inside.

I'm using

```rust
err.find::<warp::filters::body::BodyDeserializeError>()
```

to get to the error, but the only thing I can get is the error message created by

```rust
write!(f, "Request body deserialize error: {}", self.cause)
```

within the `BodyDeserializeError` implementation.

Am I missing something here? The cause is private, which is probably by design to limit the API surface, but is there another way to do this, except for writing a custom `.body()` filter (which seems cumbersome and a bit unintuitive for this use-case), or to parse the serde part of the error within the string?

Obviously, this is not urgent at all, I was just wondering if there is a "best-practice" way to handle this use-case, or if there isn't, what would be a nice way to make that happen.

seanmonstar:
Much in `warp` was as conservative as possible in revealing internals. That said, I think it'd be fair to reveal the cause here.

PR #651, opened 2020-07-16
Add Cause for BodyDeserializeError and update Rejections example

zupzup:
PR for the issue #650 

Adds a `.cause()` method for `BodyDeserializeError`, so the underlying cause can be used to handle the error properly.

Also adds a small update to the `rejections.rs` example to showcase it.

### DIFF
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -325,6 +325,13 @@ impl fmt::Display for BodyDeserializeError {
 
 impl StdError for BodyDeserializeError {}
 
+impl BodyDeserializeError {
+    /// Access the cause of the deserialization error
+    pub fn cause(&self) -> &BoxError {
+        &self.cause
+    }
+}
+
 #[derive(Debug)]
 pub(crate) struct BodyReadError(::hyper::Error);
 
--- a/src/filters/body.rs
+++ b/src/filters/body.rs
@@ -323,12 +323,9 @@ impl fmt::Display for BodyDeserializeError {
     }
 }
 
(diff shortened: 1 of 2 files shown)


jxs:
Hi, in the rejections example, is there a way that 
```Rust 
else if let Some(e) = err.find::<warp::filters::body::BodyDeserializeError>
```
might be true?

zupzup:
> Hi, in the rejections example, is there a way that
> 
> ```rust
> else if let Some(e) = err.find::<warp::filters::body::BodyDeserializeError>
> ```
> 
> might be true?

Hey!

No, that's a good point actually.

Should I add another route, which takes a body (e.g. the denominator in a JSON object), which can trigger this?

zupzup:
I added another route now, which takes the denominator via a JSON body and which triggers the error handling using e.g.:

```bash
curl -X POST  http://localhost:3030/math/10 -d '{"denomaaaa": 2}' -H "content-type: application/json"
```

src/filters/body.rs
```diff
@@ -325,6 +325,13 @@ impl fmt::Display for BodyDeserializeError {
 
 impl StdError for BodyDeserializeError {}
 
+impl BodyDeserializeError {
+    /// Access the cause of the deserialization error
+    pub fn cause(&self) -> &BoxError {
```
seanmonstar, line 330: I'd make this the `source()` method in the `impl StdError`.

zupzup: Yep, that makes more sense. fixed!

Issue #648, opened 2020-07-14
Custom Mime types for dir filter


TilBlechschmidt:
Long story short: When deploying an HLS stream by using the dir filter the mime type mismatches.

By default the mime_guess crate introduced in #44 uses `audio/x-mpegurl` instead of `application/vnd.apple.mpegURL` for `.m3u8` files. Media files with `.m4s` are not recognised at all and yield `application/octet-stream` instead of `video/mp4` [as per the standard](https://developer.apple.com/documentation/http_live_streaming/deploying_a_basic_http_live_stream).

The latter problem might be solved by adding it to the mime_guess crate but the former could cause some issues as other use-cases might need the current mime type.

To solve this problem I'd suggest a way to add a file extension to mime type mapping to the `dir` filter. This would allow very flexible changes depending on the use-case.

If there is a simpler way using filter composition that I haven't figured out yet I'd love to be enlightened :D

seanmonstar:
I agree with the sentiment! We should add some setters to the `File` reply to override some headers easily. It could probably also hold the path name to allow a user to check it to decide if and how to modify it.

TilBlechschmidt:
That is actually a pretty good idea to add it to the output instead of making it an input parameter! I might be able to allocate some spare time to get an initial PR going for this.

Am I correct in the assumption that we’d need to add the path to the File struct upon initialization to gain access for the setter or is there an even cleaner way?

Regarding the setting of headers, could we just use the `warp::reply::with_header` method to construct a new Response to replace the current one (or is that not possible because Reply -> Response only goes one way)? That raises the question whether or not it should be a setter (as in replace the response within the File struct) or a consuming function that sets a single header/header map and returns a completely new reply (maybe even `WithHeaders`)?

seanmonstar:
True, a user could just use those functions! Seems like a good first step would be adding a getter to return `&Path` from that reply.

TilBlechschmidt:
I've submitted a PR for the change on the `File` struct. It solves the issue for my use case as the path can be used to extract the extension.

You might want to take a look at the doc-example though, there might be a better way to convert the return type of the `.map`

PR #652, opened 2020-07-16
Add path getter for File Reply

TilBlechschmidt:
#648 requires a way to retrieve the file path from a `File` reply. This PR implements such a getter by storing the `ArcPath` in the `File` struct and providing a reference to it. The documentation shows an example use-case where the `Content-Type` header is changed based on the filename.

### DIFF
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -212,6 +212,32 @@ fn conditionals() -> impl Filter<Extract = One<Conditionals>, Error = Infallible
 #[derive(Debug)]
 pub struct File {
     resp: Response,
+    path: ArcPath,
+}
+
+impl File {
+    /// Extract the `&Path` of the file this `Response` delivers.
+    ///
+    /// # Example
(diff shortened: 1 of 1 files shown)


src/filters/fs.rs
```diff
@@ -212,6 +212,32 @@ fn conditionals() -> impl Filter<Extract = One<Conditionals>, Error = Infallible
...
+    ///     .and(warp::fs::dir("/www/static"))
+    ///     .map(|reply: warp::filters::fs::File| {
+    ///         if reply.path().ends_with("video.mp4") {
+    ///             warp::reply::with_header(reply, "Content-Type", "video/mp4").into_response()
+    ///         } else {
+    ///             reply.into_response()
+    ///         }
+    ///     });
+    /// ```
+    pub fn path<'a>(&'a self) -> &'a Path {
```
seanmonstar, line 238: You can leave the explicit lifetimes off here.

Issue #653, opened 2020-07-16
Should MissingConnectionUpgrade be public?


wmain:
Hi there! I'm working with `warp` as an introduction to building web services with Rust and am building an application that uses Warp's fantastic websocket support. 

In order to handle a custom error I'm appending a `recover` handler to my chain of filters, much like in the rejections example, and am handling my custom error there. If a user hits the `ws` endpoint without the proper headers, however, they'll get thrown into the fall-through "Unhandled rejection" case even though Warp is aware that this is a `MissingConnectionUpgrade` `Rejection`. Is there a better way to handle this or should `MissingConnectionUpgrade` be public?

jxs:
Hi, yeah it seems it should be public as other `Known` rejections are public, want to submit a PR making it public?

wmain:
Will do, thanks!

PR #654, opened 2020-07-16
Make MissingConnectionUpgrade public 

wmain:
Addresses #653. Given that other `Known` rejections are public, `MissingConnectionUpgrade` should be as well.


jxs:
thanks!

PR #656, opened 2020-07-16
Add filters for emitting Tracing diagnostics

### LINKED ISSUE #288: Add tracing support

hawkw:
The `tracing` crate and related libraries provide an interface for Rust
libraries and applications to emit and consume structured, contextual,
and async-aware diagnostic information. `tracing` is maintained by the
Tokio project and is becoming widely adopted by other libraries in the
stack Warp uses, such as [`hyper`], [`h2`], and [`tonic`] and in [other]
[parts] of the broader Rust ecosystem.

[`tracing` crate]: https://github.com/tokio-rs/tracing
[`hyper`]: https://github.com/hyperium/hyper/pull/2204
[`h2`]: https://github.com/hyperium/h2/pull/475
[`tonic`]: https://github.com/hyperium/tonic/blob/570c606397e47406ec148fe1763586e87a8f5298/tonic/Cargo.toml#L48
[other]: https://github.com/rust-lang/chalk/pull/525
[parts]: https://github.com/rust-lang/compiler-team/issues/331

This PR provides a new `warp::trace` module with filters to be used
when instrumenting Warp applications with `tracing`. This has to be
implemented inside of `warp`, because correctly instrumenting futures
with `tracing` spans requires the ability to wrap the future returned
by the inner filter. This branch is based on my PR #655, which migrates
Warp's internal diagnostics to Tracing.

This branch builds on the work done by @barskern in #408. I've updated
his branch to be based on the latest Warp master, added the convenience
filter for naming routes, example, and improved documentation from [my
initial attempt][1] at adding Tracing support to Warp, and made some
tweaks to the formatting of the generated traces and the user-facing
API. But, an equal amount of the credit goes to @barskern! <3

Here's some sample output from the tracing example:
Screenshot_20200716_162141

Closes #288
Depends on #655

Co-authored-by: Ole Martin Ruud <barskern@outlook.com>
Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/src/filters/trace.rs
+++ b/src/filters/trace.rs
@@ -32,14 +32,14 @@ use self::internal::WithTrace;
 ///
 /// let route = warp::any()
 ///     .map(warp::reply)
-///     .with(warp::tracing());
+///     .with(warp::trace::request);
 /// ```
 ///
 /// [`Span`]: https://docs.rs/tracing/latest/tracing/#spans
 /// [`INFO`]: https://docs.rs/tracing/0.1.16/tracing/struct.Level.html#associatedconstant.INFO
-pub fn tracing() -> Trace<impl Fn(Info) -> Span + Clone> {
+pub fn request() -> Trace<impl Fn(Info) -> Span + Clone> {
     use tracing::field::{display, Empty};
-    let func = move |info: Info| {
+    trace(|info: Info| {
         let span = tracing::info_span!(
             "request",
             remote_addr = Empty,
@@ -64,8 +64,7 @@ pub fn tracing() -> Trace<impl Fn(Info) -> Span + Clone> {
         }
 
         span
-    };
-    Trace { func }
+    })
 }
 
 /// Create a wrapping filter that instruments every request with a custom
@@ -77,21 +76,20 @@ pub fn tracing() -> Trace<impl Fn(Info) -> Span + Clone> {
 /// ```
 /// use warp::Filter;
 ///
-/// let tracing = warp::tracing::custom(|info| {
-///     // Create a span using tracing macros
-///     tracing::info_span!(
-///         "request",
-///         method = %info.method(),
-///         path = %info.path(),
-///     )
-/// });
 /// let route = warp::any()
 ///     .map(warp::reply)
-///     .with(tracing);
+///     .with(warp::trace(|info| {
+///         // Create a span using tracing macros
+///         tracing::info_span!(
+///             "request",
+///             method = %info.method(),
+///             path = %info.path(),
+///         )
+///     });
 /// ```
 ///
 /// [`Span`]: https://docs.rs/tracing/latest/tracing/#spans
-pub fn custom<F>(func: F) -> Trace<F>
+pub fn trace<F>(func: F) -> Trace<F>
 where
     F: Fn(Info) -> Span + Clone,
 {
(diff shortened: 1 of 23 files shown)


src/filters/trace.rs
```diff
@@ -0,0 +1,319 @@
...
+/// [`Span`] at the [`INFO`] level, containing a summary of the request.
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+///
+/// let route = warp::any()
+///     .map(warp::reply)
+///     .with(warp::trace::request);
```
jxs, line 35:
```suggestion
///     .with(warp::trace::request());
```

src/filters/trace.rs
```diff
@@ -0,0 +1,319 @@
+//! [`tracing`] filters.
+//!
+//! [`tracing`] is a framework for instrumenting Rust programs to
+//! collect scoped, structured, and async-aware diagnostics. This module
+//! provides a set of filters for instrumenting Warp applications with `tracing`
+//! spans. [Spans] can be used to associate individual events  with a request,
```
jxs, line 6:
```suggestion
//! spans. [`Spans`] can be used to associate individual events  with a request,
```

src/filters/trace.rs
```diff
@@ -0,0 +1,319 @@
...
+//! and track contexts through the application.
+//!
+//! [`tracing`]: https://crates.io/tracing
+/// [Spans]: https://docs.rs/tracing/latest/tracing/#spans
```
jxs, line 10:
```suggestion
//! [`Spans`]: https://docs.rs/tracing/latest/tracing/#spans
```

src/filters/trace.rs
hawkw: whoops, thanks for catching that!

seanmonstar:
I merged #655, which looks like this was based on. Mind rebasing master with this?

hawkw:
@seanmonstar should be good now, I think?

src/filters/trace.rs
```diff
@@ -0,0 +1,319 @@
...
+        pub(super) trace: Trace<FN>,
+    }
+
+    use tracing::Span;
+    use tracing_futures::{Instrument, Instrumented};
+
+    fn finished_logger<E: IsReject>(reply: &Result<(Traced,), E>) {
+        match reply {
+            Ok((Traced(resp),)) => {
+                tracing::info!(target: "warp::filters::tracing", status = resp.status().as_u16(), "finished processing with success");
```
seanmonstar, line 257: The `target` I assume is to hide the `internal` module that this is in, yea? Should it match the name of the file, though? `warp::filters::trace`? I'm not saying it should, just checking that it's a conscious decision.

hawkw: Yeah, the intention is to hide the internal module.

It was supposed to match the name of the module --- I just renamed the module from `tracing` to `trace`, but forgot to change the targets. Will fix that!

hawkw:
@seanmonstar okay, I fixed the mismatched targets, and also made a couple more tweaks to hopefully get more readable logs from the default `warp::trace::request()` filter. The user-agent string and request headers are now in a `debug` event inside the request span, rather than added to the span context, as they make every log line incredibly long:
Screenshot_20200717_103137

Hopefully you folks agree that this is more usable?

seanmonstar:
Yes, that's a nice improvement! Looking again, perhaps the user-agent shouldn't be included separately if the full headers are also included, because then it appears twice.

hawkw:
> Yes, that's a nice improvement! Looking again, perhaps the user-agent shouldn't be included separately if the full headers are also included, because then it appears twice.

Good call --- fixed it!

PR #663, opened 2020-07-21
rework sse filter

### LINKED ISSUE #645: Change SSE module to be more like WS
The `sse` module uses `trait SseEvent` and helper functions that return `impl SseEvent`, and while clever, these cause confusion (see #296) and make it difficult to match types up (see #643). This issue describes how we should change `sse` to be more like `ws`, which just uses plain structs (`ws::Messasge`).

- Create a `struct Event`.
- This should probably have private fields for `name`, `id`, `data`, and `retry`. Some should probably all be optional? Some research there will help us to know which can be optional.

jxs:
closes #645

### DIFF
--- a/src/filters/sse.rs
+++ b/src/filters/sse.rs
@@ -83,8 +83,8 @@ pub struct Event {
 impl Event {
     /// Set Server-sent event data
     /// data field(s) ("data:<content>")
-    pub fn data<T: Display>(mut self, data: Option<T>) -> Event {
-        self.data = data.map(|d| DataType::Text(d.to_string()));
+    pub fn data<T: Into<String>>(mut self, data: Option<T>) -> Event {
+        self.data = data.map(|d| DataType::Text(d.into()));
         self
     }
 
@@ -100,15 +100,15 @@ impl Event {
 
     /// Set Server-sent event comment
     /// Comment field (":<comment-text>")
-    pub fn comment<T: Display>(mut self, comment: Option<T>) -> Event {
-        self.comment = comment.map(|c| c.to_string());
+    pub fn comment<T: Into<String>>(mut self, comment: Option<T>) -> Event {
+        self.comment = comment.map(Into::into);
         self
     }
 
     /// Set Server-sent event event
     /// Event name field ("event:<event-name>")
-    pub fn event<T: Display>(mut self, event: Option<T>) -> Event {
-        self.event = event.map(|e| e.to_string());
+    pub fn event<T: Into<String>>(mut self, event: Option<T>) -> Event {
+        self.event = event.map(Into::into);
         self
     }
 
@@ -121,8 +121,8 @@ impl Event {
 
     /// Set Server-sent event id
     /// Identifier field ("id:<identifier>")
-    pub fn id<T: Display>(mut self, id: Option<T>) -> Event {
-        self.id = id.map(|e| e.to_string());
+    pub fn id<T: Into<String>>(mut self, id: Option<T>) -> Event {
+        self.id = id.map(Into::into);
         self
     }
 }
--- a/src/filters/sse.rs
+++ b/src/filters/sse.rs
@@ -11,16 +11,16 @@
 //!
 //! fn sse_events() -> impl Stream<Item = Result<Event, Infallible>> {
 //!     iter(vec![
-//!         Ok(Event::default().data(Some("unnamed event"))),
+//!         Ok(Event::default().data("unnamed event")),
 //!         Ok(
-//!             Event::default().event(Some("chat"))
-//!             .data(Some("chat message"))
+//!             Event::default().event("chat")
+//!             .data("chat message")
 //!         ),
 //!         Ok(
-//!             Event::default().id(Some(13))
-//!             .event(Some("chat"))
-//!             .data(Some("other chat message\nwith next line"))
-//!             .retry(Some(Duration::from_millis(5000)))
+//!             Event::default().id(13.to_string())
+//!             .event("chat")
+//!             .data("other chat message\nwith next line")
+//!             .retry(Duration::from_millis(5000))
 //!         )
 //!     ])
 //! }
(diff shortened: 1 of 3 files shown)


src/filters/sse.rs
```diff
@@ -52,119 +53,126 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
+        };
+        self
     }
 
-    /// Convert to boxed
-    fn boxed(self) -> BoxedServerSentEvent {
-        BoxedServerSentEvent(Box::new(self))
+    /// Set Server-sent event comment
+    /// Comment field (":<comment-text>")
+    pub fn set_comment<T: Display>(mut self, comment: Option<T>) -> Event {
```
seanmonstar, line 103: For method naming, I'd say that builders usually can skip the `set_` prefix. Most often, `set_foo` operates on `&mut self`.

What's the likelihood that we'd also want to add getters to `Event`? If `id()` should return a reference, then I'd say the builder methods should be prefixed `with_`. If we probably don't want getters, we can skip prefixes altogether.

I don't yet know which is more useful.

src/filters/sse.rs
```diff
@@ -52,119 +53,126 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
+        self.data = data.map(|d| DataType::Text(d.to_string()));
+        self
     }
 
-    /// Convert to either B
-    fn into_b<A>(self) -> EitherServerSentEvent<A, Self> {
-        EitherServerSentEvent::B(self)
+    /// Set Server-sent event data
+    /// data field(s) ("data:<content>")
+    pub fn set_json_data<T: Serialize>(mut self, data: Option<T>) -> Event {
```
seanmonstar, line 93: I'm not sure about swallowing the error in here. What if this returned a Result instead?

src/filters/sse.rs
jxs: yeah the previous api did not not have getters, i'd say we stick with only setters

src/filters/sse.rs
jxs: yeah, i was not sure either, opted for swallowing because of the previous API and the issue with `self` being consumed on case of Error. updated!

xla:
Could this get merged? It would improve the usability of the module immensely.

src/filters/sse.rs
```diff
@@ -52,119 +53,121 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
-/// Server-sent event message
-pub trait ServerSentEvent: SseFormat + Sized + Send + Sync + 'static {
-    /// Convert to either A
-    fn into_a<B>(self) -> EitherServerSentEvent<Self, B> {
-        EitherServerSentEvent::A(self)
+// Server-sent event data type
+#[derive(Debug)]
+enum DataType {
+    Text(String),
+    Json(String),
```
xla, line 69: Could this variant carry something more strongly typed like a `Serialize`?

src/filters/sse.rs
```diff
@@ -52,119 +53,121 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
+}
+
+/// Server-sent event
+#[derive(Default, Debug)]
+pub struct Event {
```
xla, line 74: All these `Option`s leave quite a big combinatorial matrix to think about when handling events. Is there a way to rework this type? For example will a comment ever exist alongside data? Expressing the mutual exclusivity (if any exist - my knowledge of the standard is a bit dusty) could help ergonomics.

xla (commented):
Generally this looks very good!

src/filters/sse.rs
```diff
@@ -52,119 +53,121 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
+    pub fn comment<T: Display>(mut self, comment: Option<T>) -> Event {
+        self.comment = comment.map(|c| c.to_string());
+        self
     }
-}
 
-impl<T: SseFormat + Send + Sync + 'static> ServerSentEvent for T {}
+    /// Set Server-sent event event
+    /// Event name field ("event:<event-name>")
+    pub fn event<T: Display>(mut self, event: Option<T>) -> Event {
```
seanmonstar, line 110: I might consider changing the bounds to `Into<String>`, since with `T: Display`, passing a `String` will result it creating a 2nd string by calling `to_string()`.

src/filters/sse.rs
```diff
@@ -52,119 +53,121 @@ use futures::{future, Stream, TryStream, TryStreamExt};
...
+        self.event = event.map(|e| e.to_string());
+        self
+    }
 
-#[allow(missing_debug_implementations)]
-struct SseComment<T>(T);
+    /// Set Server-sent event retry
+    /// Retry timeout field ("retry:<timeout>")
+    pub fn retry(mut self, duration: Option<Duration>) -> Event {
```
seanmonstar, line 117: Is there a strong reason to keep these arguments as `Option`? With a builder, will people likely (ever) need to set a property, and later on the same builder, unset it?

src/filters/sse.rs
jxs: changing all the bounds to `Into<String>` breaks cases like the [sse example](https://github.com/seanmonstar/warp/blob/master/examples/sse.rs#L9) should we still do it and update the example, or change just for specific methods?

src/filters/sse.rs
jxs: following your question on https://github.com/seanmonstar/warp/issues/645#issue-655253240 I have the idea we spoke about making them all `Option` but I am not sure now aha :sweat_smile: , should we define mandatory fields, and/or implement  mutual exclusivity as refered on https://github.com/seanmonstar/warp/pull/663#discussion_r541655071?

seanmonstar: Well, they probably should be an option internally. My comment just now was that the builder function arguments probably don't need to take an `Option`.

src/filters/sse.rs
seanmonstar: I think it's fine that the example be changed to `data(counter.to_string())` or similar.

src/filters/sse.rs
jxs: ah right! yeah I think I the reason was allowing to unset, but i doubt people will need to set a property and later unset it.
Updated it!

seanmonstar (approved):

src/filters/sse.rs
jhpratt: Question regarding this! Is the string cloned internally, or is the allocation reused for every connection? If it's being cloned, that seems like a _major_ performance problem. Not so much for `fn event`, but for `fn data` or `fn json_data`.

PR #667, opened 2020-07-21
Relax some lifetime bounds

### LINKED ISSUE #666: Relax `'static` requirement of `bind`'s `addr: impl Into<SocketAddr>` argument
I might be misunderstanding, but I'm struggling to work out why a `'static` lifetime is required for `bind` and `bind_ephemeral`'s `addr` argument. I've removed it and warp still compiles fine with all the tests passing.

My particular use case is that I'm configuring the server's port externally (so it can be changed without recompilation), so my address doesn't have a `'static` lifetime. I could use `Box::leak` to work around this if there's a good reason to keep the `'static` lifetime.

JamesHinshelwood:
Closes #666 

This PR started by just removing some `'static` bounds on the `addr` arguments of some functions in `server.rs`.

While I was there I also relaxed some of the other arguments' lifetimes by adding an explicit lifetime to the returned type. I'm not yet sure if this was a wise idea. I don't believe its a breaking change as presumably all the current callers are providing `'static` arguments so their returned lifetime will still be `'static`. I might be wrong here though.

This is also my first PR, so let me know if I've missed anything obvious.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -128,7 +128,7 @@ where
     <F::Future as TryFuture>::Error: IsReject,
 {
     /// Run this `Server` forever on the current thread.
-    pub async fn run(self, addr: impl Into<SocketAddr> + 'static) {
+    pub async fn run(self, addr: impl Into<SocketAddr>) {
         let (addr, fut) = self.bind_ephemeral(addr);
         let span = tracing::info_span!("Server::run", ?addr);
         tracing::info!(parent: &span, "listening on http://{}", addr);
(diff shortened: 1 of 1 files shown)


JamesHinshelwood:
Looks like I might have been a bit overzealous. I will investigate the test failures.

JamesHinshelwood:
@seanmonstar Apologies for the delay in getting back to this. I've reduced the scope of the PR to just the cases which only affect the the argument lifetime.

PR #674, opened 2020-07-23
Add .with_var filter

sidju:
Adds a generic filter for binding in any Send + Clone variable into any filter chain. Useful for binding in database connections, password hashing settings etc. into handlers.

Not necessarily relevant to have in the framework, but nice to have in case you want it.

(If it is of interest I can write up an example with it as well)

### DIFF
--- a/src/filter/inject.rs
+++ b/src/filter/inject.rs
@@ -1,0 +1,59 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+use futures::{ready, TryFuture};
+use pin_project::pin_project;
+
+use super::{Filter, FilterBase, Func, Internal};
+
+#[derive(Clone, Copy, Debug)]
+pub struct Inject<T, V> {
+    pub(super) filter: T,
+    pub(super) variable: V,
+}
+
+impl<T, V> FilterBase for Inject<T, V>
+where
+    T: Filter,
+    V: Clone + Send,
+{
+    type Extract = (V, T::Extract,);
+    type Error = T::Error;
+    type Future = InjectFuture<T, V>;
+    #[inline]
+    fn filter(&self, _: Internal) -> Self::Future {
+        InjectFuture {
+            extract: self.filter.filter(Internal),
+            variable: self.variable.clone(),
+        }
+    }
+}
+
+#[allow(missing_debug_implementations)]
+#[pin_project]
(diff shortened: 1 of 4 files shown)


sternix:
Hi, thank you @sidju, i want a feature like this too,

first 
adding "var and var::with_var" to the "pub use self::filters::{ ... }" block in the 
https://github.com/seanmonstar/warp/blob/master/src/lib.rs file
provide us warp::with_var() usage instead of warp::filters::with_var()

but what you think about

1-) name will be "inject" ?
ex:
     warp::any()
          .and(warp::inject(db.clone()))
          .and_then(...)

2-) can be part of warp::Filter
ex: 
    warp::any()
         .inject(db.clone())
         .and_then(...)

Thanks.

sidju:
I like all three propositions, thanks for the feedback.
I will look into adding it to Filter in a day or two. If that doesn't work out I'll do the other two.

-------- Original Message --------

sternix:
Hi @sidju these are my suggestions, but @seanmonstar 's and @jxs 's ideas important for me,

jxs:
Hi, what's the advantage of this filter over using the pattern [`warp::any().map(move || db.clone`](https://github.com/seanmonstar/warp/blob/master/examples/todos.rs#L99) ?

sternix:
i think, 

less code,
common pattern

sidju:
I have made a test of inject as a part of filter now (I think).
My main frustration is that it currently clones the variable twice (once when creating the future and once out of the Pin), which really bothers me... Can I get away from that somehow?

sidju:
Some choice on this would be nice. I'd just like to be done with it, if you will.

@seanmonstar @jxs ; Is this a feature you wish to add, or is it sufficiently easy to do that you'd rather recommend using
```
warp::any().map(move || x.clone())"
```
?

jxs:
Hi, and sorry for the delay! This is such a simple case that I would rather only have
```Rust
warp::any().map(move || x.clone())"
```
yeah

hamza1311:
This also adds the problem of stack allocating the variable on every filter build. The solution that @jxs described is the best way to go about it.

sidju:
Well then, just thought I'd offer since I found it somewhat useful. But needless features are not in-line with the idea of the project, so I always knew it to be a long shot.
Thanks for your time reviewing!

PR #675, opened 2020-07-24
Remove static lifetime for serve_incoming

fanatid:
I'm not sure that I correct but for looks like `serve_incoming` / `serve_incoming_with_graceful_shutdown` no need static lifetime for incoming.

With PR I added example which not compile without changes:
```bash
error[E0597]: `listener` does not live long enough
  --> examples/serve_incoming.rs:14:25
   |
14 |         .serve_incoming(listener.incoming())
   |                         ^^^^^^^^-----------
   |                         |
   |                         borrowed value does not live long enough
   |                         argument requires that `listener` is borrowed for `'static`
15 |         .await;
16 | }
   | - `listener` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.
error: could not compile `warp`.
```


src/server.rs
```diff
@@ -314,9 +314,9 @@ where
     /// This can be used for Unix Domain Sockets, or TLS, etc.
     ///
     /// Returns a `Future` that can be executed on any runtime.
-    pub fn serve_incoming<I>(self, incoming: I) -> impl Future<Output = ()> + 'static
```
seanmonstar, line 317: I *think* by removing this `+ 'static` bound on the *return type*, it might break existing usage. It might break this:

```rust
fn inc() -> impl Stream<Item = Result<TcpStream, io::Error>> + Send + 'static {
    unimplemented!()
}

let srv = warp::serve(routes).serve_incoming(inc());
tokio::spawn(srv);
```

fanatid: I do not think that this will break something, `'static` lifetime on returning type only limiting the lifetime, after change `'static` still will be allowed (with other lifetimes). Just tried this code, compiled and executed without problem:
```rust
#![deny(warnings)]

use futures::{stream, Stream};
use tokio::{io, net::TcpStream};
use warp::Filter;

fn inc() -> impl Stream<Item = Result<TcpStream, io::Error>> + Send + 'static {
    stream::pending()
}

#[tokio::main]
async fn main() {
    pretty_env_logger::init();

    let routes = warp::get().map(warp::reply);
    warp::serve(routes).serve_incoming(inc()).await;
}
```

seanmonstar: Awesome, thanks for checking!

examples/serve_incoming.rs
```diff
@@ -0,0 +1,16 @@
+#![deny(warnings)]
```
seanmonstar, line 1: I'd suggest removing the example, it doesn't show too much, and having too many can make it harder to know what examples to look at to learn.

fanatid: Removed.

fanatid:
Thank you! Can you make a release?

PR #676, opened 2020-07-25
Add host filters

jakajancar:
Resolves issue https://github.com/seanmonstar/warp/issues/432.
Obsoletes PR https://github.com/seanmonstar/warp/pull/489 (better logic, added tests and documentation).

cc @dmeijboom

@jxs, I'd appreciate a review.

### DIFF
--- a/src/filters/authority.rs
+++ b/src/filters/authority.rs
@@ -1,0 +1,96 @@
+//! Authority (aka `Host` header) filter
+//!
+use std::str::FromStr;
+use crate::filter::{Filter, filter_fn_one, One};
+use crate::reject::{self, Rejection};
+use futures::future;
+pub use http::uri::Authority;
+
+/// Creates a `Filter` that requires a specific authority (target server's
+/// host and port) in the request.
+///
+/// # Example
+///
+/// ```
+/// use warp::{Filter, authority::Authority};
+///
+/// let multihost =
+///     warp::authority("foo.com").map(|| "you've reached foo.com")
+///     .or(warp::authority("bar.com").map(|| "you've reached bar.com"));
+/// ```
+pub fn authority(expected: &str) -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    let expected = Authority::from_str(expected).expect("invalid authority");
+    optional()
+        .and_then(move |option: Option<Authority>| {
+            match option {
+                Some(authority) if authority == expected => future::ok(()),
+                _ => future::err(reject::not_found())
+            }
+        })
+        .untuple_one()
+}
+
+/// Creates a `Filter` that looks for an authority (target server's host
+/// and port) in the request.
+///
+/// If found, extracts the `Authority`, otherwise continues the request,
+/// extracting `None`.
+///
+/// Rejects with `400 Bad Request` if the `Host` header is malformed or if there
+/// is a mismatch between the `Host` header and the target URI.
+///
+/// # Example
+///
+/// ```
+/// use warp::{Filter, authority::Authority};
+///
+/// let authority = warp::authority::optional()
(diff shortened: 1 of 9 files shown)


dmeijboom:
Nice work 👍🏽 I closed my PR in favor of this one.

src/filters/authority.rs
```diff
@@ -0,0 +1,92 @@
...
+/// # Example
+///
+/// ```
+/// use warp::{Filter, authority::Authority};
+///
+/// let multihost =
+///     warp::authority("foo.com").map(|| "you've reached foo.com")
+///     .or(warp::authority("bar.com").map(|| "you've reached bar.com"));
+/// ```
+pub fn authority(expected: &str) -> impl Filter<Extract = (), Error = Rejection> + Clone {
```
seanmonstar, line 21: Since this filter requires a match, perhaps the name should indicate that. What these options: `required`, `guard`, or `exact` (also used in `warp::header::exact`)?

src/filters/authority.rs
```diff
@@ -0,0 +1,92 @@
...
+        let name = "host";
+        let from_header = route.headers()
+            .get(name)
+            .map(|value|
+                // Header present, parse it
+                value.to_str().map_err(|_| reject::invalid_header(name))
+                    .and_then(|value| Authority::from_str(value).map_err(|_| reject::invalid_header(name)))
+            );
+
+        future::ready(match (from_uri, from_header) {
```
seanmonstar, line 76: Since checking the headers requires a hashmap lookup, I'd check `from_uri` first, and only if `None`, *then* try to get the value from the headers.

jakajancar:
> I initially read "authority" as "authorization". I wonder if others would think similarly. What about using the term "host" instead?

I was thinking about that. In the end, I went with Authority because:

 1. hyper and the http crate call it authority (and the struct you get is `http::uri::Authority`),
 2. HTTP2 and 3 normally refer to it as (server) authority, and the `Host` header is more of a historic thing.

But absolutely agree re. familiarity and happy to change.  I would keep using `http::uri::Authority` though, but let me know.

src/filters/authority.rs
jakajancar: Ok, I'll rename it to `exact` to match `header::exact` and remove the `warp::authority` altogether (so you either have to use `::exact` or `::optional`)

src/filters/authority.rs
jakajancar: I kinda like some strictness and see it fail in case there is a discrepancy between URL and Host, rather than just use one of them.

I would also imagine that in the overall inefficient design of warp where you need to do `::exact` multiple times in subtrees (as opposed to `::optional` once and route to a sub-filter dynamically based on that), a hash lookup is a comparatively tiny inefficiency.

So I would leave it as it is, unless you insist.

jakajancar:
@seanmonstar Can you please provide answers/acknowledge that my suggestions are OK on my 2 questions?

src/filters/authority.rs
seanmonstar: What you say makes sense, I didn't notice the subtlety in the implementation.

seanmonstar:
I agree that the header host is phased out for `:authority`... I suppose I'd wonder what other frameworks expose. If everything at the framework level calls it host, does that mean we should also? I don't think we have to. But maybe it's less confusing to users?

jakajancar:
- Renamed filter from `authority` to `host`
  - Removed "top-level" filter in favor of explicit `warp::host::exact()` and `warp::host::optional()`

"Host" does seem much more familiar. The returned object is still `Authority` (from `http::uri::Authority`), but hopefully the documentation bridges between the familiar & easy-to-find and the technically accurate.

PR #683, opened 2020-08-06
Remove `Sync` restriction on stream bassed into `warp::sse::reply`

wngr:
As the `Sync` requirement on `Body::wrap_stream` was removed in https://github.com/hyperium/hyper/pull/2187, it can now also be removed on the stream passed to `warp::sse::reply`.


jxs:
thanks! lgtm, I have also relaxed the `Sync` need on https://github.com/seanmonstar/warp/pull/663

PR #687, opened 2020-08-08
Document that filters::query::query uses serde_urlencoded, and add example for postprocessing filters::query::raw

jcaesar:
`serde_urlencoded` is used for to deserialize query strings, and hence not all structs that are `Deserialize` can be deserialized to. I think it would be neat to document that fact and also added an example that shows how to use a different deserialization function instead.

(I didn't want to add serde_qs as a dev-dependency just for this, so I fudged it.)

What do you think?

### DIFF
--- a/src/filters/query.rs
+++ b/src/filters/query.rs
@@ -59,6 +59,10 @@ use crate::reject::{self, Rejection};
 ///     });
 /// ```
 ///
+/// Not all structs are supported for use as query objects, e.g. structs containing vectors will
+/// result in an error at runtime. If you need more flexibility, consider using [raw] to supply
+/// your own deserialization mechanism.
+///
 /// For more examples, please take a look at [examples/query_string.rs](https://github.com/seanmonstar/warp/blob/master/examples/query_string.rs).
 ///
 /// [Serde]: https://docs.rs/serde
--- a/src/filters/query.rs
+++ b/src/filters/query.rs
@@ -79,6 +79,32 @@ pub fn query<T: DeserializeOwned + Send + 'static>(
 }
 
 /// Creates a `Filter` that returns the raw query string as type String.
+///
+/// Can be used to implement a filter with a custom deserializer, for example the `serde_qs` crate.
+///
+/// ```
+/// # fn main() {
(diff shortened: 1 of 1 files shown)


seanmonstar:
Yes this is a great idea, thanks! I wonder, what if instead of "promising" to use `serde_urlencoded`, we just say that currently the decoder strictly follows the URL spec, and so cannot decode lists etc.

jcaesar:
> currently the decoder strictly follows the URL spec

Hm, I'll rephrase it like that. But I do wonder whether it's accurate, or at least helpful.
On the one hand side, the URL spec allows duplicated keys (okay, it'd be odd to expect that to work with rust structs structs), on the other hand side, `serde_urlencoded` [supports tagged](https://github.com/nox/serde_urlencoded/issues/26#issuecomment-487720975)/[untagged](https://github.com/nox/serde_urlencoded/issues/66#issue-576914133) enums, which I had not expected to work at all (but only if all values are `String`).
Then again, mentioning `serde_urlencoded` does not help that much, since these things aren't documented.

jcaesar:
Maybe explicitly stating what can be decoded into is a good idea?

jxs:
sorry! missed this one and this has since been addressed by #843

jcaesar:
I don't see how #843 documents what limitations the limitations of query parameter deserialization. I wouldn't consider this addressed. But I do feel like I'm being pedantic. In the off-case that you want me to renovate this PR, please do tell.

jxs:
Oh sorry! And no, not pedantic at all, if you want to rebase your PR against master and add the limitations of parameter deserialization that would be great!

jcaesar:
I've rebased on master and reworded it so the documentation doesn't leak any implementation details.

jxs (approved):
LGTM thanks! Out of curiosity, what does `Pars` mean in this context?

jcaesar:
Ah, if you're asking, I should probably spell it out. `Parameters`. But I'll change it to `FooQuery` to match the other example.

jcaesar:
Please notify me if you're still interested in this. I'll renovate it if so.

PR #691, opened 2020-08-14
Update README.md

mooreniemi:
I think it is a bit confusing to advise `macros` rather than `full`, because only with `full` do we get total benefit of using multiple cores. Maybe I am missing something. :)


jxs:
Hi, yeah `full` seems like a better default, thanks!

PR #693, opened 2020-08-20
Add wrap_fn

jxs:


apiraino:
hey @jxs this PR looks really interesting!

I've loosely followed the referenced issue, I'm especially interested in the use case underlined by [this comment](https://github.com/seanmonstar/warp/issues/16#issuecomment-629765704), i.e. having the capability of making multi-step filters easier to write and follow.

How could this patch help in that regard? I can try hacking something when this PR gets merged, but I would love a more explicit (perhaps real world) example of usage.

Thanks!

jxs:
@apiraino i am going to merge this so as it becomes available and the issue can be closed, but the example can always be improved after as we discussed in discord

apiraino:
yes please! :+1: I didn't mean to block the merge, I'll try figuring that out as my time allows (couldnt make a good example quickly)

cutsoy:
Is there a good way to represent these `Wrap`s as return types for functions. For example:

```rust
pub fn my_custom_wrapper<F>() -> impl warp::Wrap<F>
where
    F: Filter,
    F::Extract: ...,
{
    warp::wrap_fn(|filter| ...)
}
```

To be used with:

```rust
warp::path(...).with(my_custom_wrapper())
```

This currently isn't possible yet due to the fact that ```Wrap``` is still sealed. Is there any other opaque type that can be returned?

jxs:
@cutsoy it should be
```Rust
fn my_custom_wrapper<F, T>(
    filter: F,
) -> impl Filter<Extract = (&'static str,)> + Clone + Send + Sync + 'static
where
    F: Filter<Extract = (T,), Error = std::convert::Infallible> + Clone + Send + Sync + 'static,
    F::Extract: warp::Reply,
{
...
}
```
take a look at the [wrapping example](https://github.com/seanmonstar/warp/blob/master/examples/wrapping.rs) for more details

PR #695, opened 2020-08-22
Upgrade dependence

### LINKED ISSUE #694: Upgrade tokio-tungstenite to latest(v0.11) version?
[tokio-tungstenite](https://crates.io/crates/tokio-tungstenite) released 0.11 version, we should be upgrade the dependence.

valbendan:
Fix #694 

It works well in my environment.

But when other user mixes use tokio-tungstenite v0.10 and v0.11, there will be compatibility issues.

ps: upgrade to the same version can solve this problem


jxs (approved):

jxs:
thanks!

PR #696, opened 2020-08-23
allow warp::path!() call with no arguments

utkarshkukreti:
This is a feature request + implementation for it. Let me know if you think this is a bad idea.

---

It is equivalent to `warp::path::end()`. This is useful to have some symmetry
when defining multiple routes with `path!`:

```
warp::path!().map(|| 0)
.or(warp::path!("add" / u32 / u32).map(|a, b| a + b))
.or(warp::path!("mul" / u32 / u32).map(|a, b| a * b))
```


utkarshkukreti:
Rebased!

masklinn:
FWIW I had this exact same need / request recently. I replaced the not-working `path!()` call by an `end()` but it feels unclear and icky.

jxs (approved):

jxs:
thanks!

Issue #460, opened 2020-02-26
Tail() call crashes to expect


jq-rs:
When configuring redirect with `tail()`, it crashes from time to time to `.expect("server URIs should always have path_and_query")` with Warp 0.1.x. It could be because someone accesses it with `CONNECT foo.bar`. See #429 for discussion.

jq-rs:
This can be worked around by not using tail() when doing redirect.

seanmonstar:
It looks like this occurs in 0.2.x as well.

Simple test case:

```rust
let filter = warp::path::tail();

let _tail = warp::test::request()
    .path("localhost")
    .filter(&filter)
    .await
    .unwrap();
```

PR #697, opened 2020-08-25
Fix #568

### LINKED ISSUE #568: Panic on invalid path
Hello,
I'm building a service and I'm using warp to for the http stack and at some point I had a malformed request and warp just panicked instead of returning an error for me to handle somehow...

The malformed request was
```
GET foo.txt HTTP/1.1


```
as you can see there's a missing slash at the start of the path, if I add it everything works fine...

This is the warp server code:
```
let warp = warp::get().and(warp::fs::dir(path)).with(warp::log::log("static webserver");
warp::serve(webserver).run(ADDR).await;
```

Backtrace [here](http://ix.io/2ll6).
This is the line just before the panic:
```
15: warp::filters::path::path_and_query
             at /home/karrq/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.2.2/src/filters/path.rs:465
```

cr0ax:
This is difficult to reproduce, the provided code snippet in the issue did not run (mismatched parens etc)

The full code snippet that I believe the issue author was trying to provide would look more like:

```
use warp::Filter;

#[tokio::main]
async fn main() {
    let static_server = warp::get()
        .and(warp::fs::dir("files"))
        .with(warp::log::log("static webserver"));
    warp::serve(static_server).run(([0, 0, 0, 0], 3030)).await;
}
```

However, it's unclear how to send a malformed request like the one they were referencing, precisely _because_ it is malformed.

Instead I added a breaking test that I think mimics that malformed request, and then added code to fix it.

### DIFF
--- a/src/route.rs
+++ b/src/route.rs
@@ -95,14 +95,13 @@ impl Route {
 
     pub(crate) fn set_unmatched_path(&mut self, index: usize) {
         let index = self.segments_index + index;
-
         let path = self.req.uri().path();
-
-        if path.len() == index {
+        if path.len() < index {
+            return;
+        } else if path.len() == index {
             self.segments_index = index;
         } else {
             debug_assert_eq!(path.as_bytes()[index], b'/');
-
             self.segments_index = index + 1;
         }
     }
--- a/src/route.rs
+++ b/src/route.rs
@@ -97,8 +97,7 @@ impl Route {
         let index = self.segments_index + index;
         let path = self.req.uri().path();
         if path.is_empty() {
-            // note: empty path is malformed due to
-            // mismatch in output of path_and_query
+            // malformed path
             return;
         } else if path.len() == index {
             self.segments_index = index;
--- a/src/route.rs
+++ b/src/route.rs
@@ -96,7 +96,9 @@ impl Route {
     pub(crate) fn set_unmatched_path(&mut self, index: usize) {
         let index = self.segments_index + index;
         let path = self.req.uri().path();
-        if path.len() == 0 {
+        if path.is_empty() {
+            // note: empty path is malformed due to
+            // mismatch in output of path_and_query
             return;
         } else if path.len() == index {
             self.segments_index = index;
--- a/src/route.rs
+++ b/src/route.rs
@@ -96,7 +96,7 @@ impl Route {
     pub(crate) fn set_unmatched_path(&mut self, index: usize) {
         let index = self.segments_index + index;
         let path = self.req.uri().path();
-        if path.len() < index {
+        if path.len() == 0 {
             return;
         } else if path.len() == index {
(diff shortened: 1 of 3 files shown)


cr0ax:
Trying to rationalize this further.

The following snippet:

```
    let some_uri = "todos.rs".parse::<http::Uri>()
        .expect("will not fail");
    println!("path and query: {:?}", some_uri.path_and_query());
    println!("static path_and_query: {}", http::uri::PathAndQuery::from_static(""));
```

produces the output:

```
path and query: None
static path_and_query: /
```

The call to `path_and_query` from [tail](https://github.com/seanmonstar/warp/blob/master/src/filters/path.rs#L294) returns `/`, but [Route#new](https://github.com/seanmonstar/warp/blob/ea42ba5690c75fabf847871788a79cd58177749b/src/route.rs#L46) sets `segments_index` to `0` because the actual `path` does not contain a `/` character.

Which will produce an out of bounds error on in the `debug_assert_eq!(path.as_bytes()[index], b'/');` check, and presumably will cause more issues down the line.

After investigating this I feel like perhaps the fix should be located in `tail`?

cr0ax:
For the 'localhost' (#460) case, inside `tail`, `route` is `''`, and tail gets `/` for `path_and_query(&route)`*, resulting in a call to `set_unmatched_path` with `index = 1` but `let path = self.req.uri().path() = ''`. Attempting to index the route will panic on the debug assertion.

For the 'todos.rs' (#568) case, a similar thing happens.

In both cases, not advancing the `segments_index` should be the right thing to do, considering the path is 0-length anyway, and returning `/` and `0` to the next filter should result in correct behavior.

*(because of the previously mentioned default for `from_static`)

src/route.rs
```diff
@@ -95,14 +95,13 @@ impl Route {
 
     pub(crate) fn set_unmatched_path(&mut self, index: usize) {
         let index = self.segments_index + index;
-
         let path = self.req.uri().path();
```
jxs, line 98: can we use `path.is_empty()` instead , and add a comment noting path malformed for the empty path?

jxs (changes requested):
Hi, and thanks for looking into this! only some minor addressing

src/route.rs
cr0ax: how's that ?

src/route.rs
```diff
@@ -95,14 +95,15 @@ impl Route {
...
-
-        if path.len() == index {
+        if path.is_empty() {
+            // note: empty path is malformed due to
+            // mismatch in output of path_and_query
```
jxs, line 101:
```suggestion
            // malformed path
```

jxs (approved):

jxs:
thanks! :balloon:

PR #701, opened 2020-08-30
Make redirect::AsLocation public

haileysome:
This enables documentation while still keeping the trait sealed outside of the `redirect` mod. Here's what the docs now look like:

screen_shot_2020_08_30_at_16 07 01-8922e96c0e046c

screen_shot_2020_08_30_at_16 07 08-263caea19cfdda


jxs (approved):

jxs:
Hi and thanks for the PR! Even the trait keeps sealed for implementation it's now open to use on type's and function signatures, thou I think it's ok in this case :+1:

PR #702, opened 2020-08-30
Implement fold function.

tmpfs:
So that dynamically generated routes can easily be chained.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -38,7 +38,7 @@ pub fn fold<F, I>(filters: I) -> Option<Or<F, F>>
 where
     F: Filter<Error = Rejection> + Sized,
     F::Error: CombineRejection<F::Error>,
-    I: IntoIterator<Item = F>
+    I: IntoIterator<Item = F>,
 {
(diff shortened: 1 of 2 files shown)


tmpfs:
Hey folks,

So this is testing the water to see if you are interested in including something like this; if so I will add docs, test cases and an example.

The use case is combining routes that should be dynamically generated but also helps for conditional routes too (just don't add them to the iterator). I ran into the TLS/Host issue and noticed the new `host` filters and wanted to define *virtual hosts* however these virtual host definitions should come from configuration data loaded at runtime therefore I needed some way to chain the filters.

I would prefer to pass a `acc: Or<F, F>` as the initial argument to `fold()` but couldn't get the types working correctly - doing this would allow us to remove the `Option` from the return type; but this has the downside that the first argument must be an `Or` as opposed to an `F` which makes it slightly less flexible.

tmpfs:
This doesn't work as expected and I can't figure out how to get the types working for this sort of functionality.

Would really appreciate some help with this feature, but as this doesn't work I will close this PR.

PR #703, opened 2020-08-31
Backport of tail crash fix #568 (#697) to 0.1.x

### LINKED ISSUE #568: Panic on invalid path

jq-rs:
return from set_unmatched_path on malformed path, closes #568 and #460, backport to 0.1.x

### DIFF
--- a/examples/handlebars_template.rs
+++ b/examples/handlebars_template.rs
@@ -6,7 +6,6 @@ extern crate warp;
 extern crate serde_json;
 extern crate serde;
 
-use std::error::Error;
 use std::sync::Arc;
 
 use handlebars::Handlebars;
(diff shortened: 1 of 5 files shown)


seanmonstar:
Published in v0.1.23.

Issue #242, opened 2019-06-23
Document whether path! and `param` (and param2) de-urlencode strings before conversion 


droundy:

droundy:
Here is how I figured out to work around this issue:
```
let list_of_lists = path!(String).map(|name: String| {
       let name = percent_encoding::percent_decode(name.as_bytes()).decode_utf8().unwrap();
       ...
});
```
It feels, however, like this should be done in the `path!` processing code.

seanmonstar:
I agree on both counts, it should be documented for sure, and probably should be decided automatically. First one is easier, the latter needs some thought to not affect performance too much.

droundy:
I wonder for the next breaking change whether you'd want to switch to a new `FromPercentEncoding` trait, which for the standard types could be equivalent to `FromStr` for types for which no valid representation contains invalid HTML characters and does percent decoding for types where it could matter.

On the other hand, maybe the performance cost to percent decode is minimal and doesn't matter? It's hard to imagine a path becoming large...

giacomocariello:
Bump

LucianoBestia:
Is this a good approach?

I crate a new type PercentDecoded that guarantees it has been decoded.
It has traits FromStr and ToString.
The standard  String cannot guarantee if it has been decoded or not.

in warp I get the parameter from url directly in PercentDecoded type:
```rust
warp::path!("name" / PercentDecoded)
	.and_then(|name: PercentDecoded| async move {
		search(name);
		...
```
the function accepts PercentDecoded to be sure it has been decoded.
It cannot accept a String.
```rust
pub async fn search(name: PercentDecoded) {
	let name = name.to_string();
	dbg!(name);
	...
```


```rust
//! encode_decode_mod

// cargo.toml: 
// percent-encoding = "2.1.0"
// anyhow = "1.0.31"

use core::str::FromStr;
use anyhow::{Error};
use std::string::ToString;
use percent_encoding::{percent_decode_str, AsciiSet, CONTROLS};

/// https://url.spec.whatwg.org/#fragment-percent-encode-set
const FRAGMENT: &AsciiSet = &CONTROLS.add(b' ').add(b'"').add(b'<').add(b'>').add(b'`');

pub fn utf8_percent_encode(s:&str)->String{
    percent_encoding::utf8_percent_encode(s,FRAGMENT).to_string()
}

// region: type with guarantee that it has been decoded

#[derive(Clone, Debug)]
pub struct PercentDecoded {
[...]

PR #704, opened 2020-09-02
Add param_decode() to URL decoding segments.

luke-clifton:
If you know your encoding might contain odd characters, you may need to
URL decode it first.

Also extends the `path!()` macro to allow for

```
path!("hello" / (decode String))
```
which will use `param_decode()` in the expansion instead of `param()`. Am happy to accept alternate syntax suggestions.

Somewhat addresses #242


luke-clifton:
Personally I think URL decoding should probably be the default, and if you want to bypass the encoding that should be opt-in. I am happy to flip the naming around to make it so, but it's a breaking change at that point.

hpux735:
given that this is a year-old MR that's still open, am I to gather that warp currently doesn't handle url-encoded path components?

PR #705, opened 2020-09-05
Reduce number of dependencies: Switch to percent-encoding & update serde_urlencoded

nickelc:
`percent-encoding` is already pulled in with the `url` & `serde_urlencoded` crates.
And `serde_urlencoded` v0.7 switched to `ryu` like `serde_json` did in the past.


jxs (approved):
LGTM, thanks!

PR #706, opened 2020-09-10
Permit stapling an OCSP response when using TLS

thombles:
OCSP isn't widely supported yet in rust TLS land (e.g. creating requests/responses) but rustls provides the stapling mechanism—you can supply a chunk of bytes, which is presumed to be a valid DER response, which is passed verbatim to the client.

In an ideal world this would become more fully featured:
* automatic fetching and caching of the OCSP response
* updating the response without having to restart the server, since any token is usually only valid for 7 days.

However the barebones/static version of this is useful to me and I would like to simply expose the rustls config.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -50,6 +50,7 @@ impl std::error::Error for TlsConfigError {}
 pub(crate) struct TlsConfigBuilder {
     cert: Box<dyn Read + Send + Sync>,
     key: Box<dyn Read + Send + Sync>,
+    ocsp_resp: Vec<u8>,
 }
 
 impl std::fmt::Debug for TlsConfigBuilder {
@@ -64,6 +65,7 @@ impl TlsConfigBuilder {
         TlsConfigBuilder {
             key: Box::new(io::empty()),
             cert: Box::new(io::empty()),
+            ocsp_resp: Vec::new(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


jxs (approved):
Hi, looks good to me!

src/tls.rs
```diff
@@ -136,7 +144,7 @@ impl TlsConfigBuilder {
 
         let mut config = ServerConfig::new(NoClientAuth::new());
         config
-            .set_single_cert(cert, key)
+            .set_single_cert_with_ocsp_and_sct(cert, key, self.ocsp_resp, Vec::new())
```
seanmonstar, line 147: If these vecs are empty, is it the same as just calling `set_single_cert`?

jxs: yeah, by [`ServerConfig::set_single_cert_with_ocsp_and_sct`](https://docs.rs/rustls/0.18.1/rustls/struct.ServerConfig.html#method.set_single_cert_with_ocsp_and_sct) doc: 
`ocsp is a DER-encoded OCSP response. Ignored if zero length. scts is an SignedCertificateTimestampList encoding (see RFC6962) and is ignored if empty.`

thombles:
For the record this wasn't as useful as I thought. My next challenge was `serve_incoming` missing from `TlsServer`, and the moment I finished [implementing it](https://github.com/seanmonstar/warp/compare/master...thombles:tls-server-incoming) I realised it was all rather foolish—the smarter thing is to accept TLS streams myself and pass them into a regular warp `Server`. In that configuration I can of course set whatever I want for the OCSP response without getting warp involved.

PR #710, opened 2020-09-18
Allow passing query parameters to the test websocket

caspg:
Right now query string is ignored when using `warp::test::ws()`. This change will allow to pass query parameters to the test websocket.


seanmonstar (approved):

seanmonstar:
Thanks!

PR #717, opened 2020-10-02
Provide a try_bind_with_graceful_shutdown method to TlsServer.

### LINKED ISSUE #680: Missing function: warp::TlsServer::try_bind_with_graceful_shutdown
warp library has the functions:

  - [`warp::Server::bind_with_graceful_shutdown`](https://docs.rs/warp/0.2.4/warp/struct.Server.html#method.bind_with_graceful_shutdown)
  - [`warp::Server::try_bind_with_graceful_shutdown`](https://docs.rs/warp/0.2.4/warp/struct.Server.html#method.try_bind_with_graceful_shutdown)
  - [`warp::TlsServer::bind_with_graceful_shutdown`](https://docs.rs/warp/0.2.4/warp/struct.TlsServer.html#method.bind_with_graceful_shutdown)

but is missing the function:

  - `warp::TlsServer::try_bind_with_graceful_shutdown`

jnicholls:
- Provide a try_bind_with_graceful_shutdown method to TlsServer in order to have fallible semantics rather than panics when a TLS configuration fails.
- Add panic warnings to the docs of each method on Server and TlsServer that could panic, where it was missing.

Fixes #680 

Side notes:

- I'm not entirely sure under what circumstances a lot of these methods need to take in an `Into<SocketAddr> + 'static` rather than just an `Into<SocketAddr>`, as well as returning a `Future<Output = ()> + 'static` rather than just a `Future<Output = ()>`. There are no tests to show a difference in behavior in terms of what contexts a static lifetime is warranted. Removing all forms of `+ 'static` and we compile all tests and they all pass. Likewise with the `signal` parameters being `+ Send + 'static`...this doesn't seem necessary because hyper's [with_graceful_shutdown](https://docs.rs/hyper/0.13.8/hyper/server/struct.Server.html#method.with_graceful_shutdown) does not require `Send + 'static` semantics. Can we just remove all of these extra trait & lifetime bounds?
- Some of the API interface is inconsistent imho.
  - `try_bind` doesn't return a `Result` but instead returns nothing and just does a tracing event. It should behave like `bind` and call `try_bind_ephemeral` and return a `Result<impl Future<Output = ()>, crate::Error>`
  - Some docs, like in `Server::bind` say they return a Future but they don't, they wait on the future, but are an `async fn` and so technically they do return a future, but one that awaits on the inner server future. It's kind of a confusing interface to passersby that do not understand the distinction even though they are technically the same. Should we re-write these methods to be regular functions and return an explicit `Future` just to be consistent?


jnicholls:
@seanmonstar Any thoughts on this one? I'm gonna close it if it's ignored any longer. Thanks.

seanmonstar (approved):
Sounds useful, thanks!

Issue #614, opened 2020-06-18
Is there any purpose to allow_any_origin()?


Newbytee:
Greetings. 

In my efforts to learn warp, I'm experimenting `warp::cors()` and I'm seemingly finding that it implicitly allows any origins **without** specifying `allow_any_origin()`. Curious, I gave removing `allow_any_origin()` on line 59 in the cors.rs test a go, and all tests passed still.

Is this intended, and if so, what is the purpose of `allow_any_origin()`?

ernestas-poskus:
> allow_any_origin()

Method simulates https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin and allows all origins, though does explicitly insert ACAO header with `*` asterisk value. 

> A response that tells the browser to allow code from any origin to access a resource will include the following:
> Access-Control-Allow-Origin: *

Newbytee:
> Method simulates https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin and allows all origins, though does explicitly insert ACAO header with `*` asterisk value.

Yes, that's how I imagined it would work, but it's actually not necessary for this. Here is its implementation: https://github.com/seanmonstar/warp/blob/master/src/filters/cors.rs#L193

Notice that it sets `self.origins` to None. Now, what is the default value of `self.origins`? Line 47 tells us: https://github.com/seanmonstar/warp/blob/master/src/filters/cors.rs#L47

It's also None. Therefore, I'm asking what the purpose of this is, and when it might be useful. The docs do not mention this behaviour and rather, by how I read it, imply that `allow_any_origin()` is mandatory: https://docs.rs/warp/0.2.3/warp/filters/cors/fn.cors.html

PR #726, opened 2020-10-19
Add support for CORS wildcard origin

paulhauner:
## What

This PR permits the following for `Cors::Builder`:

- `builder.allow_origin("*")`
- `builder.allow_origins(["*"])` (and any other combination including `*`)

## Why

The [MDN Access-Control-Allow-Origin](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) docs state that there is a `*` directive which:

> For requests without credentials, the literal value "*" can be specified, as a wildcard; the value tells browsers to allow requesting code from any origin to access the resource. Attempting to use the wildcard with credentials will result in an error.

### DIFF
--- a/src/filters/cors.rs
+++ b/src/filters/cors.rs
@@ -455,7 +455,6 @@ mod internal {
     use std::task::{Context, Poll};
 
     use futures::{future, ready, TryFuture};
-    use headers::Origin;
     use http::header;
     use pin_project::pin_project;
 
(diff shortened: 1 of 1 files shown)


paulhauner:
Hey, I'm wondering if there's interest in getting this merged? It's still something we desire on https://github.com/sigp/lighthouse, which is now in production.

jxs:
Hi, and sorry for the delay. 
Isn't the wildcard already possible via [`allow_any_origin`](https://docs.rs/warp/0.2.5/warp/filters/cors/struct.Builder.html#method.allow_any_origin)? and by default, as stated on https://github.com/seanmonstar/warp/issues/614#issuecomment-646521685

paulhauner:
> Hi, and sorry for the delay.
> Isn't the wildcard already possible via [`allow_any_origin`](https://docs.rs/warp/0.2.5/warp/filters/cors/struct.Builder.html#method.allow_any_origin)? and by default, as stated on [#614 (comment)](https://github.com/seanmonstar/warp/issues/614#issuecomment-646521685)

No worries, thanks for the reply. I'm pretty sure that we don't explicitly provide the `*` ACAO which has some negative effect. It's been a while and I'd need to produce a PoC to prove this. My hands are full at the moment, but I intend to do this soon. Thanks for the patience.

PR #727, opened 2020-10-19
Add a http-kv example

liufuyang:
As I had some fun trying to use warp to create a simple http server as a simple key/value storage, I think it might be nice to have this as an example?

It can illustrate how to use `async move {}` closures. And I feel it is a bit simpler than the Todo example.

A few questions though:

1. It seems that on line 44 I cannot use `.and(warp::any().map(|| map_clone))`, and clone it twice seems a bit weird but that's how I get it compiled otherwise I have error saying I am providing a FnOnce not Fn or something.
2. I haven't figure out a way to conveniently return a Json with a different HTTP code than 200.
3. I tried to use `Err(reject::not_found())` for the key not found path but while curling I saw `405 Method Not Allowed` instead of a `404`, not really sure why 🤔 

Hopefully you think this could be a valuable example? And it would be nice if you could help answer some of the questions above so I can remove some of the TODOs in the code. 

Thank you.

### DIFF
--- a/examples/http_kv.rs
+++ b/examples/http_kv.rs
@@ -1,0 +1,79 @@
+use serde_derive::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::env;
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+use warp::reply::Json;
+use warp::{reject, Filter, Rejection};
+
+#[derive(Deserialize, Serialize, Clone)]
+struct Value {
+    value: String,
+}
+
+/// Provides a RESTful web server as a simple key value store
+///
+/// API will be:
+///
+/// - `GET /:key`: get the value with key
+/// - `POST /:key "{"value": "some stuff"}"`: create/update a new key with value
+/// - `DELETE /:key`: delete a specific key
+///
+/// Testing curl command could be:
+/// - `curl -i -X POST -H "Content-Type:application/json" localhost:3030/k1 -d '{"value": "v1"}'`
+/// - `curl -i -X GET -H "Content-Type:application/json" localhost:3030/k1`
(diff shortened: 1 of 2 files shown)


liufuyang:
I tried to do something mentioned here but unsuccessful...
https://github.com/helgoboss/realearn/commit/b88b4606f85dee814f4c0053a4f9461dde1ca12f

examples/http_kv.rs
```diff
@@ -0,0 +1,88 @@
...
+        .and(warp::any().map(move || map_clone.clone()))
+        .and_then(|key: String, map: Map| get_handler(key, map));
+
+    let map_clone = map.clone();
+    let delete = warp::delete()
+        .and(warp::path::param::<String>())
+        .and(warp::any().map(move || map_clone.clone()))
+        .and_then(|key: String, map: Map| delete_handler(key, map));
+
+    // View access logs by setting `RUST_LOG=todos`.
```
mlh758, line 58: http_kv instead?

liufuyang: Thank you. I need to find some time recently to update this PR.

jxs:
Hi, thanks for the interest!  A [warp key value example](https://github.com/tower-rs/tower-http/blob/master/examples/warp-key-value-store/src/main.rs) was added to the `tower` but you seem to have added more methods such as `DELETE` , what do you think of PR your example there?
thanks

liufuyang:
@jxs Aha, nice to know, so you mean close this and then basically use the examples at tower-http? :)

jxs:
yeah :slightly_smiling_face:

liufuyang:
Sure, it sounds good. I will close this now :)

Issue #712, opened 2020-09-21
Improved rejections


mankinskin:
I get the impression that the rejection system in this crate ought to be severely refactored. The [rejections example](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs) does not show a simple way of handling functions which return `Results`:
```rust
async fn login(credentials: Credentials) -> Result<UserSession, LoginError> {
    ...
}
```
which is the standard way of handling Errors in Rust. The suggested way of handling this seems to be to
- implement `Reject for LoginError` (which requires me to own the `LoginError` type)
- call `recover` after the handling filter with a "rejection handler" function, which converts rejections into a `Reply` type which can be sent by warp

This causes a lot of cognitive and syntactic overhead, and I think the crate would benefit a lot from a refactor like this.

mankinskin:
For context, this is the way that I found to handle this case:
```rust
    let login = warp::post()
        .and(warp::path!("api"/"login"))
        .and(warp::body::json())
        .and_then(|credentials: Credentials| async {

            Ok(match login(credentials).await {
                Ok(session) => warp::reply::json(&session).into_response(),
                Err(status) => warp::reply::with_status("", status).into_response(),
            }) as Result<warp::reply::Response, core::convert::Infallible>

        });
```
There is a lot of boilerplate here that ought to be handled by `Filter::and_then` (and friends) or the `Reject` or `Reply` trait.
The problem I see currently is that [`Filter::and_then`](https://docs.rs/warp/0.2.5/warp/trait.Filter.html#method.and_then) requires a function, returning a `TryFuture`, resolving to a `Result`, where the `Error` implements [`CombinedRejection`](https://github.com/seanmonstar/warp/blob/master/src/reject.rs#L592), which is a sealed trait of this crate, and is only implemented for [`Rejection`](https://docs.rs/warp/0.2.5/warp/reject/struct.Rejection.html) and `Infallible`.

[...]

PR #740, opened 2020-11-05
WIP A Result can be a Reply

kaj:
Provide `impl Reply for Result<T, E>` where both `T` and `E` implements Reply.  This impl makes no difference between an Ok and an Err, but calls `into_response` for the contained data of either.  If  an error should be logged, that is the responsibility of the `Reply` impl on the `E` type.

This is testing my suggestion https://github.com/seanmonstar/warp/issues/712#issuecomment-722389194 .

To be really useful, it needs a way to use `async fn foo(...) -> impl Reply` as an handler.  Filter `.map` does not do async while `.and_then` needs a `Resule<impl Reply, Rejecttion>`.  Is there a middle version?

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -411,19 +411,24 @@ impl Reply for ::http::StatusCode {
     }
 }
 
-impl<T> Reply for Result<T, ::http::Error>
+impl<T, E> Reply for Result<T, E>
 where
     T: Reply + Send,
+    E: Reply + Send + std::error::Error,
 {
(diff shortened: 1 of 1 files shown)


kaj:
Hm.  Turns out that a) this is a duplicate of #458 and b) that PR also includes the `FilterBase::map_async` functions that makes it really useful and convenient.

Only problem with that PR is that it is from February and have merge conflicts.  But they should be solvable.

PR #739, opened 2020-11-04
Allow the activation on client authentication

Geobert:
Fixes #737

On behalf of Isode Ltd.

Issue #737, opened 2020-11-03
A way to activate Client Auth


Geobert:
**Is your feature request related to a problem? Please describe.**

I'm writing an OAuth2 server and need client authentication via Mutual TLS.

This line makes me think it's not possible to activate client auth: 
https://github.com/seanmonstar/warp/blob/2ceef6935ee67004027c4585dc661e4eee48f5ae/src/tls.rs#L145

**Describe the solution you'd like**

Add  `TlsConfigBuilder::client_cert_verifier(verifier: Arc<dyn ClientCertVerifier>)`

Issue #504, opened 2020-03-24
Allow providing custom rustls ServerConfig


dbrgn:
I'd like to configure the accepted TLS versions. I could also imagine wanting to configure the suggested ciphersuites in the future.

I don't think it makes sense to expose separate functions for all those potential TLS configuration options. Maybe warp could allow providing a custom [`ServerConfig`](https://docs.rs/rustls/0.17.0/rustls/struct.ServerConfig.html) instead?

seanmonstar:
I didn't want to expose the version of rustls in the public API. But I do think we can add those methods to the TlsBuilder.

seanmonstar:
Warp's TLS support is mostly for convenience. If one wants to customize all the finer details, you can convert the filters into a `Service` and provide your own acceptor.

dbrgn:
> Warp's TLS support is mostly for convenience. If one wants to customize all the finer details, you can convert the filters into a Service and provide your own acceptor.

Ok, good to know. Maybe that's worth a note in the docs?

In any case, an option to set accepted TLS versions and ciphersuites might be nice to have.

PR #743, opened 2020-11-06
Tls client authentication

janpetschexain:
This extends the `TlsServer` for optional client authentication by providing `client_auth_optional()` and `client_auth_required()` method (and their `_path()` counterparts) to expose `rustls` functionality, partly addressing #504.

It seems that others are having the same request/issue as well (#737), but this pr implements tls client authentication along the lines of the existing tls server authentication (`cert()` and `key()` methods) in contrast to #739.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -429,20 +429,24 @@ where
         self.with_tls(|tls| tls.cert_path(path))
     }
 
-    /// Specify the file path to read the trusted anchors for client authentication.
+    /// Specify the file path to read the trust anchor for optional client authentication.
     ///
-    /// If client authentication against the trusted anchors is `required`, then allow
-    /// only authenticated clients, otherwise allow anonymous or authenticated clients.
-    ///
-    /// If no trusted anchors are given via [`root_path`] or [`root`], then client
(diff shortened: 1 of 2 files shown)


jxs (approved):
LGTM!

seanmonstar (commented):
Thanks for the PR! I think adding the feature is a great idea!

I have a few suggestions around the exposed API:

- Perhaps the method names should indicate client authentication, `root` by itself made me think that's where it would auto search for the server key and cert...
- Boolean secondary arguments can easily become unclear, so I would suggest a separate method, maybe `client_auth_required()`.

janpetschexain:
> Thanks for the PR! I think adding the feature is a great idea!
> 
> I have a few suggestions around the exposed API:
> 
> * Perhaps the method names should indicate client authentication, `root` by itself made me think that's where it would auto search for the server key and cert...
> * Boolean secondary arguments can easily become unclear, so I would suggest a separate method, maybe `client_auth_required()`.

sure, i'll update this accordingly :)

jxs (approved):

jxs:
Thanks! :100:

Issue #62, opened 2018-08-19
Add Authorization header filters


algermissen:
I took a shot at adding support for Authorization header filters. I would like to know whether you like the approach I took: https://github.com/algermissen/warp/blob/add-auth-headers/src/filters/authorization.rs

If so, I'll work on making the code better and submit a PR after that - given I am relatively new to Rust, there is a lot of stuff that needs improvement (eg reduce String usage), so please do not judge that yet.

Some questions:
- I think an addition to Rejections is needed to support the 401 challenge response, yes?
- Since the 401 response includes parameters that must be settable by the code using warp, Rejections must somehow support additional information that can be set (eg the Basic auth filter would need to return a rejection with realm information):
```    
    let text3 = warp::path("protected").
        and(warp::authorization::basic("myRealm"))
        .map(|b: warp::authorization::BasicCredentials| {
            format!("User: {} Passwd: {}", b.user,b.password)
        });
```    
- Or would you rather return an error response instead of a rejection since from the point of view of the HTTP authentication protocol, the thing to do is to return 401 + challenge. Allowing for further rejection handling does not really make sense.

algermissen:
Playing around with this a bit more it seems one way to go would be to make rejections an enum so that those status codes that imply certain headers (eg 401 Unauthorized) can carry the data needed for these headers (eg scheme and realm).

I am also thinking of how to return [HTTP problems](https://tools.ietf.org/html/rfc7807) which is the same use case but with the additional need to be even more flexible.

I am happy to code up a suggestion, but would appreciate a hint on your preferences what path to take.

seanmonstar:
I believe something like this would be awesome! I've been spending time on [typed headers](https://github.com/hyperium/headers), as I expect them to be very useful in warp, and believe they can help with these filters.

algermissen:
Working on a new approach based on typed headers. Comments welcome:

https://github.com/algermissen/warp/blob/9e3e8bc1245800623fd1c69fbf0738f033cbd125/examples/basic_auth.rs#L10

Next step is to allow for a callback in basic() that validates credentials and returns realm.

algermissen:
Here is an updated version of an authorization filter 
https://github.com/algermissen/warp/blob/85a91d5ead7d52865db38afc0557f5ddee5a119f/examples/basic_auth.rs#L25

Very much WIP, though.

@seanmonstar Comments very much welcome, specifically:
- I have the feeling your plans how to approach this are quite different, yes?
[...]

PR #751, opened 2020-11-18
working auth filter rejection with authentication header and realm name

mike-kfed:
This is for issue #62 based on pr #716 but mine actually works. However I understand it is a crazy hack how I pass the realm name along to create the additional header. If some maintainer could guide me how to make that clean please? I'll fix it as needed.

For cleaner implementation: I'd assume we need a new rejection type. Because a failed authentication (i.e. user/pass wrong) will require the same headers returned as this authentication header missing filter does, only now needed by another filter: the one checking the credentials.

### DIFF
--- a/examples/basic_auth.rs
+++ b/examples/basic_auth.rs
@@ -1,0 +1,13 @@
+//#![deny(warnings)]
+use headers::{authorization::Basic, Authorization};
+use warp::Filter;
+
+#[tokio::main]
+async fn main() {
+    let routes = warp::auth::basic("Realm name").map(|auth_header: Authorization<Basic>| {
+        // TODO: some password lookups done in here
+        println!("authorization header = {:?}", auth_header);
+        format!("Hello, {} you're authenticated", auth_header.0.username())
+    });
+    warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
+}
--- a/src/filters/auth.rs
+++ b/src/filters/auth.rs
@@ -1,0 +1,30 @@
+//! HTTP Basic Authentication Filters
+
+use futures::future;
+use headers::{authorization::Basic, Authorization};
+
+use crate::filter::Filter;
+use crate::reject::Rejection;
+
+use super::header;
+
+/// Creates a `Filter` to the HTTP Basic Authentication header. If none was sent by the client, this filter will reject
+/// any request with a 401 Unauthorized.
+///
+/// # Example
+///
+/// ```
+/// use std::net::SocketAddr;
+/// use warp::Filter;
+/// use headers::{authorization::Basic, Authorization};
+///
+/// let route = warp::auth::basic("Realm name")
+///     .map(|auth_header: Authorization<Basic>| {
+///         println!("authorization header = {:?}", auth_header);
+///     });
+/// ```
+pub fn basic(
+    realm: &'static str,
+) -> impl Filter<Extract = (Authorization<Basic>,), Error = Rejection> + Copy {
+    header::header2().or_else(move |_| future::err(crate::reject::unauthorized(realm)))
+}
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -79,6 +79,12 @@ pub(crate) fn missing_cookie(name: &'static str) -> Rejection {
     known(MissingCookie { name })
 }
 
+// 401 Unauthorized
+#[inline]
+pub(crate) fn unauthorized(realm: &'static str) -> Rejection {
+    Rejection::unauthorized(Box::new(realm.to_string()))
+}
+
 // 405 Method Not Allowed
 #[inline]
 pub(crate) fn method_not_allowed() -> Rejection {
@@ -184,6 +190,7 @@ enum Reason {
 
 enum Rejections {
     Known(Known),
+    Unauthorized(Box<dyn Cause>),
     Custom(Box<dyn Cause>),
     Combined(Box<Rejections>, Box<Rejections>),
 }
@@ -272,6 +279,12 @@ impl Rejection {
     fn custom(other: Box<dyn Cause>) -> Self {
         Rejection {
             reason: Reason::Other(Box::new(Rejections::Custom(other))),
+        }
+    }
+
+    fn unauthorized(other: Box<dyn Cause>) -> Self {
+        Rejection {
+            reason: Reason::Other(Box::new(Rejections::Unauthorized(other))),
         }
     }
 
@@ -376,6 +389,7 @@ impl fmt::Debug for Reason {
             Reason::Other(ref other) => match **other {
                 Rejections::Known(ref e) => fmt::Debug::fmt(e, f),
                 Rejections::Custom(ref e) => fmt::Debug::fmt(e, f),
+                Rejections::Unauthorized(ref e) => fmt::Debug::fmt(e, f),
                 Rejections::Combined(ref a, ref b) => {
                     let mut list = f.debug_list();
                     a.debug_list(&mut list);
@@ -410,6 +424,7 @@ impl Rejections {
                 | Known::MissingExtension(_)
                 | Known::BodyConsumedMultipleTimes(_) => StatusCode::INTERNAL_SERVER_ERROR,
             },
+            Rejections::Unauthorized(..) => StatusCode::UNAUTHORIZED,
             Rejections::Custom(..) => StatusCode::INTERNAL_SERVER_ERROR,
             Rejections::Combined(ref a, ref b) => preferred(a, b).status(),
         }
@@ -417,6 +432,21 @@ impl Rejections {
 
     fn into_response(&self) -> crate::reply::Response {
         match *self {
+            Rejections::Unauthorized(ref e) => {
+                let body = format!("Please Authorize for: {:?}", e);
+                let mut res = http::Response::new(Body::from(body));
+                *res.status_mut() = self.status();
+                res.headers_mut().insert(
+                    CONTENT_TYPE,
+                    HeaderValue::from_static("text/plain; charset=utf-8"),
+                );
+                let realm = e.downcast_ref::<String>().expect("realm must be string");
+                res.headers_mut().insert(
+                    "www-authenticate",
+                    HeaderValue::from_str(format!("Basic realm=\"{}\"", realm).as_str()).unwrap(),
+                );
+                res
+            }
             Rejections::Known(ref e) => {
                 let mut res = http::Response::new(Body::from(e.to_string()));
                 *res.status_mut() = self.status();
@@ -447,6 +477,7 @@ impl Rejections {
     fn find<T: 'static>(&self) -> Option<&T> {
         match *self {
             Rejections::Known(ref e) => e.inner_as_any().downcast_ref(),
+            Rejections::Unauthorized(ref e) => e.downcast_ref(),
             Rejections::Custom(ref e) => e.downcast_ref(),
             Rejections::Combined(ref a, ref b) => a.find().or_else(|| b.find()),
         }
@@ -455,6 +486,9 @@ impl Rejections {
     fn debug_list(&self, f: &mut fmt::DebugList<'_, '_>) {
         match *self {
             Rejections::Known(ref e) => {
+                f.entry(e);
+            }
+            Rejections::Unauthorized(ref e) => {
                 f.entry(e);
             }
             Rejections::Custom(ref e) => {
@@ -507,6 +541,11 @@ unit_error! {
 unit_error! {
     /// The request's content-type is not supported
     pub UnsupportedMediaType: "The request's content-type is not supported"
+}
+
+unit_error! {
+    /// Unauthorized request
+    pub Unauthorized: "Unauthorized request"
 }
 
 /// Missing request header
--- a/src/filters/mod.rs
+++ b/src/filters/mod.rs
@@ -5,6 +5,7 @@
 
 pub mod addr;
 pub mod any;
+pub mod auth;
 pub mod body;
 #[cfg(feature = "compression")]
 pub mod compression;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -120,6 +120,7 @@ pub use self::filters::{
     addr,
     // any() function
     any::any,
+    auth,
     body,
     cookie,
     // cookie() function


src/filters/auth.rs
```diff
@@ -0,0 +1,30 @@
...
+///
+/// let route = warp::auth::basic("Realm name")
+///     .map(|auth_header: Authorization<Basic>| {
+///         println!("authorization header = {:?}", auth_header);
+///     });
+/// ```
+pub fn basic(
+    realm: &'static str,
+) -> impl Filter<Extract = (Authorization<Basic>,), Error = Rejection> + Copy {
+    header::header2().or_else(move |_| future::err(crate::reject::unauthorized(realm)))
```
FreeMasen, line 29: I feel like it would be nicer to have the callback argument be `Basic` instead of `Authorization`

examples/basic_auth.rs
```diff
@@ -0,0 +1,13 @@
+//#![deny(warnings)]
+use headers::{authorization::Basic, Authorization};
+use warp::Filter;
+
+#[tokio::main]
+async fn main() {
+    let routes = warp::auth::basic("Realm name").map(|auth_header: Authorization<Basic>| {
+        // TODO: some password lookups done in here
```
FreeMasen, line 8: It might be nice for the example to actually include a method for password lookups

src/reject.rs
```diff
@@ -410,13 +424,29 @@ impl Rejections {
...
+                let mut res = http::Response::new(Body::from(body));
+                *res.status_mut() = self.status();
+                res.headers_mut().insert(
+                    CONTENT_TYPE,
+                    HeaderValue::from_static("text/plain; charset=utf-8"),
+                );
+                let realm = e.downcast_ref::<String>().expect("realm must be string");
+                res.headers_mut().insert(
+                    "www-authenticate",
+                    HeaderValue::from_str(format!("Basic realm=\"{}\"", realm).as_str()).unwrap(),
```
FreeMasen, line 446: The unauthorized response should be able to handle both `Basic` and `Bearer` replies at least. [The full list of valid values here is 10 entries long](http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml), so some flexibility on the `scheme` would be nice.

Also according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) the `realm` _could_ be omitted, though I am not sure it is wise to allow that

src/reject.rs
```diff
@@ -509,6 +543,11 @@ unit_error! {
     pub UnsupportedMediaType: "The request's content-type is not supported"
 }
 
+unit_error! {
```
FreeMasen, line 546: This is probably not a `unit_error` if the `scheme` is going to be provided

src/filters/auth.rs
mike-kfed: I don't follow what you mean here, can you please elaborate?

examples/basic_auth.rs
mike-kfed: well anything serious needs to pass in some form of database, file or at least a hashmap dragging in dependencies. easiest form is a silly `if name == password` check, but not sure that makes sense.

src/reject.rs
mike-kfed: Given that what I've coded so far feels kinda hacky, supporting multiple schemes calls for an even cleaner implementation. Meaning for example coming up with a new rejection type as mentioned in my description of the PR? I'll try to add `Bearer` but I am not sure this is really readable code after without fundamental changes.

src/filters/auth.rs
FreeMasen: Something like the following
```rust
pub fn basic(
    realm: &'static str,
) -> impl Filter<Extract = (Basic,), Error = Rejection> + Copy {
    header::header2().map(|auth: Authorization<Basic>| auth.0).or_else(move |_| future::err(crate::reject::unauthorized(realm)))
```

examples/basic_auth.rs
FreeMasen: If the goal of an example is to illustrate how to use the filter, it would be good to provide something closer to a real world example instead of a todo comment. Using the `dir` example as a template, the project provides an actual path that is populated with files. When that example is run, it actually serves content, which helps to illustrate _how_ to use that filter.

src/reject.rs
FreeMasen: Updating the [error type](https://github.com/seanmonstar/warp/pull/751/files/7b357e450c292dc252d40325d2aa82c56994a3e9#diff-3b90cd596f94fb9464f356d926c42bb98767b769f938613f9f7e971d3e328c3cR548) to not be a `unit_error` should simplify the realm and scheme extraction from the user's configuration.

src/reject.rs
mike-kfed: what other type of error is it then? I just tried make it return Unauthorized request, I am not a full pro of warps internals ;)

src/filters/auth.rs
mike-kfed: thanks, okay I'll look into that!

src/reject.rs
FreeMasen: [MissingHeader](https://github.com/seanmonstar/warp/pull/751/files/7b357e450c292dc252d40325d2aa82c56994a3e9#diff-3b90cd596f94fb9464f356d926c42bb98767b769f938613f9f7e971d3e328c3cR553) would be an error you could model this off of.

src/reject.rs
FreeMasen: On issue #62 there are a few files still available from the previous PR around an auth filter. I've updated their approach to the current master [here](https://github.com/FreeMasen/warp/blob/auth_filter/src/reject.rs#L111) which might be worth looking at.

examples/basic_auth.rs
mike-kfed: I'll add a mutex guarded Hashmap then, everything else needs 3rd party libs. Okay like that?

src/reject.rs
mike-kfed: thx, will look into it

src/reject.rs
mike-kfed: thx for the feedback, I'll hack away :)

examples/basic_auth.rs
FreeMasen: You may want to make it a `Arc<tokio::sync::Mutex<HashMap<String, String>>>` since almost any implementation is going to need to do some kind of `async` work. The `tokio` dependency will already be there from using `tokio::main`

mike-kfed: I realised a lock-guard is not needed, since this is read-only. example should be fine now

src/filters/auth.rs
mike-kfed: I see how that helps now, much cleaner code afterwards :) should be okay now too

mike-kfed:
While getting rid of unit-error and looking at your attempt with the Authorizer trait, I realized my example is broken. Here's a quick attempt to work around the limitation:

```rust
use headers::authorization::Basic;
use std::collections::HashMap;
use std::sync::Arc;
use warp::Filter;

async fn is_authenticated(
    auth_header: Basic,
    pwdb: Arc<HashMap<&str, &str>>,
) -> Result<String, warp::Rejection> {
    let user = auth_header.username();
    if pwdb.get(user) == Some(&auth_header.password()) {
        Ok(user.to_string())
    } else {
        Err(warp::reject::unauthorized("Basic", "Realm")) // stupid hack making unauthorized public
    }
}

#[tokio::main]
async fn main() {
    let pwdb: HashMap<_, _> = vec![
        ("alice", "wonderland"),
        ("bob", "cat"),
        ("carl", "IePai4ph"),
    ]
    .into_iter()
    .collect();
    let pwdb = Arc::new(pwdb); // no lock-guard needed, it's read-only
    let pwdb = warp::any().map(move || pwdb.clone());
    let routes = warp::auth::basic("Realm name")
        .and(pwdb.clone())
        .and_then(is_authenticated)
        .map(|user: String| format!("Hello, {} you're authenticated", user));

    warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
}
```

Another important thing, the `CONTENT_TYPE` header is hardcoded to `"text/plain; charset=utf-8"` in everything we do. this is another stumbling block for securing REST-APIs (especially when Bearer scheme is used this is a no-go, people will use this for json).

In my application I hacked around all those limitations making a custom rejection handler and more boilerplate. Would be nice to also allow for custom-content-type too while we are at it.

I like your Authorizer trait idea, seems more flexible than my simple `and_then` filter approach.

examples/basic_auth.rs
FreeMasen: My suggestion of using the tokio mutex was mostly to have some async work to perform in the handler, the new example is quite a bit more illustrative though

FreeMasen:
> Another important thing, the CONTENT_TYPE header is hardcoded to "text/plain; charset=utf-8" in everything we do. this is another stumbling block for securing REST-APIs (especially when Bearer scheme is used this is a no-go, people will use this for json).

I am not sure I follow this, the 401 reply is sent back with an empty body so the `Content-Type` of that reply is going to be `plain/text`. Any subsequent handler would be able to reply with whatever `Content-Type` they would like. Are you saying that you'd like configure the reply's body when the header is not found? If so, then adding a new `Rejection` would probably be a requirement that a user could `.recover` from.

> I like your Authorizer trait idea, seems more flexible than my simple and_then filter approach.

I would be curious where you end up with this, I ultimately decided that this wasn't the right abstraction but it could have been my implementation.

mike-kfed:
the problem arises when the authentication fails, then the warp reject code we are building here is active, correct? If that is the case you'll have a content-type of plain/text instead of potentially wanted json. Plus no control of output on auth-error, you might want to send something like `{"error": "blah", "msg": "password expired"}`. Sure if auth succeeds then whatever the user serves is correctly handled, since our filter here was passed.

Basically we need to be able to handle

1. no request-header came in
2. header came in but password/token wrong
3. header came in but invalid
4. header came in auth ok.

cases 1, 3 and 4 are easily doable with crate only code, number 2 is where the user likely wants to customise.

examples/basic_auth.rs
mike-kfed: Sure I understand, but anything that needs async work (e.g. update of password db while server is running) is a way too big example. And showing use of a Mutex when lock-free is actually good enough is not a good example either ;)
however my example is broken, as it does not return a 401 on auth-error. but this is what my comment in the parent thread discusses, how to do this best 🤔

FreeMasen:
> header came in but password/token wrong

This would have to follow down the rejection path. It is a bit more boilerplate but I think allowing for rejections to be handled outside of the primary handler is a nice thing about this pattern. Here is a gist with the basic pattern:

https://gist.github.com/FreeMasen/c0ec97e03bfe93d3d703ee31a605bcbf

mike-kfed:
thanks for the help, that is basically what I built for my app - I think it is a lot of boilerplate necessary to "just" have authentication. If we can find a way to hide that for the users of warp it'd be cooler. I'll return trying to find a nice solution on monday, no computer accessible this weekend :)

mike-kfed:
I made my code a bit more readable and played with your example code.
Problem using the rejection handler is now

- not providing user/pass returns Error 500 because it does not handle `UnauthorizedChallenge` rejections
- if I were to add `UnauthorizedChallenge` handling, that struct would have to be public and not private. 
  -  Let's suppose that it was public: we'd write redundant boiler plate returning the www-authenticate header with realm/scheme which is already in `src/reject.rs`.
- using `rej.into_response()` in the reject handler doesn't work either because `warp::reject::sealed::IsReject` trait is private
- problem I remember from my private project, if you add a websocket server to this you need to add e.g. `MissingConnectionUpgrade` handling, even more boilerplate just because authentication is of interest lots of other stuff needs to be handled too.

I think we need to figure out a better way for `.recover()` in general. I'd propose we polish up the filters::auth and reject modules, write a working example with all boilerplate needed, try to get that merged into master. And then open a new issue that discusses getting rid of the boilderplate? what do you think?

mike-kfed:
quickhack for `into_response()` being private:

```diff
diff --git a/src/reject.rs b/src/reject.rs
index 05e4a86..e2cbb4c 100644
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -337,6 +337,11 @@ impl Rejection {
             false
         }
     }
+
+    /// workaround for recover() handler
+    pub fn make_response(&self) -> hyper::Response<Body> {
+        self.into_response()
+    }
 }

 impl<T: Reject> From<T> for Rejection {

```

then your examples else branch is simply 

```rust
let res = if let Some(inner) = rej.find::<AuthRejections>() {
// ...
} else {
    rej.make_response()
};
```

but this change goes way beyond this issues title I feel.

FreeMasen:
>     * not providing user/pass returns Error 500 because it does not handle `UnauthorizedChallenge` rejections

My expectations is that the UnauthorizedChallenge wouldn't be a rejection but a reply that is already taken care of by the framework. If I instrument an endpoint with `warp::auth::basic("MyRealm", |basic| { //... })` I want warp to send the 401 for me automatically. If that isn't something provided by the filter it really doesn't seem very useful since you can get the same via `warp::header("Authorization")`

mike-kfed:
This is what I mean, as soon as you recover you have to deal with any rejection, the `examples/rejections.rs` even shows that you need to deal with 404 and everything. So yes right now you could build this yourself with `warp::header("Authorization")` no advantage using warp's library filters, which is the point I tried to make - a fundamental change of how this works is needed. (Or I am misunderstanding how  `. recover()` is supposed to be used, my experience so far is it hands over any rejection in the route giving you responsibility to handle it yourself.)

FreeMasen:
> (Or I am misunderstanding how `. recover()` is supposed to be used

I think you are understanding how recover is supposed to be used. In the case that a route is defined using `basic` I would expect that an reply, not a rejection, would be sent by that filter to orchestrate the challenge. CORS preflight requests would be a similar example, an incoming OPTIONS request is not handled by the instrumented route, instead is replied to by warp under the hood.

> Another important thing, the CONTENT_TYPE header is hardcoded to "text/plain; charset=utf-8" in everything we do. this is another stumbling block for securing REST-APIs

I believe this is still the point you are trying to make that I may be misunderstanding. Are you saying that you want to reply to an unauthorized request with some kind of custom body during the Unauthorized Challenge? If so, can you provide a little more detail about the use case you see as being a stumbling block? I would expect the body of the Unauthorized Challenge response to be empty making the content type not really important

mike-kfed:
> I would expect the body of the Unauthorized Challenge response to be empty making the content type not really important

most webservers send a human readable string informing them auth is necessary, otherwise the browser window stays empty when the user clicks "cancel" in his auth-popup. But yeah, when the content is empty, you can also argue it is not necessary to send a content-type at all ;)

FreeMasen:
> the user provided the wrong auth info, i.e. after a custom is_authenticated call

I would expect this should lead to a 403 and not another 401 as defined [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#the_general_http_authentication_framework)

> if you want to write a pure json only api,

I think this is a separate issue entirely. I could possibly see the inclusion of another Server level configuration or some kind of `with(warp::default_content_type("json"))` but in this case I would say worrying about the content-type is out of scope. Alternatively it may already be covered by [the `with_header`](https://docs.rs/warp/0.3.0/warp/reply/fn.with_header.html) fn

mike-kfed:
> I would expect this should lead to a 403 and not another 401 as defined here, also for reference

correct, I patched it to send empty body and no content-type now.

> I think this is a separate issue entirely.

I agree.

Leaves with how to best treat recover for rejections lowering the load of auth implementations. My hack from above is basically the rejections-filter doing the default warp action if not handled myself. what do you think?

mike-kfed:
> CORS preflight requests would be a similar example, an incoming OPTIONS request is not handled by the instrumented route, instead is replied to by warp under the hood.

I just played with this, the cors filter also is implemented with rejections, difference is it's use of wrapsealed, however if your example code was extended with CORS this way, the handle_rejections also gets the CORS rejections to handle:

```rust
    let routes = warp::auth::basic("Realm name")
        .and(pwdb.clone())
        .and_then(is_authenticated)
        .map(|user: String| format!("Hello, {} you're authenticated", user))
        .with(cors)
        .recover(handle_rejections);
```

normally nobody notices that because they probably chain the `.with()` at the end of their route avoiding it by chance. So I guess the auth filter is implemented correctly, it's just a question of how to resolve issue #451

FreeMasen:
> normally nobody notices that because they probably chain the `.with()` at the end of their route avoiding it by chance. So I guess the auth filter is implemented correctly, it's just a question of how to resolve issue #451

I feel like auth should _probably_ be a `with` style filter as apposed to the normal route style filter (these should have better names...). As it stands I haven't been able to figure out how to do `async` work from a `with` style filter, either using the [`wrap_fn`](https://docs.rs/warp/0.3.0/warp/fn.wrap_fn.html) filter or trying to setup [my own `auth` filters](https://github.com/FreeMasen/warp/blob/auth_filter/src/filters/auth.rs).

mike-kfed:
I agree a `with` style filter would help, and it might be worth the effort to get one going. But in the end it is probably easier to make a public interface to convert a rejection into a `impl Reply` which is needed anyway for other use-cases too.

mike-kfed:
Hello @FreeMasen I started a new branch for a with-filter called [`http_basic_auth_cb`](https://github.com/mike-kfed/warp/tree/http_basic_auth_cb), based on PR #819 but allowing for a callback function to authenticate, filter code here [basic.rs](https://github.com/mike-kfed/warp/blob/http_basic_auth_cb/src/filters/auth/basic.rs) how to use it is shown here [tests/auth_basic.rs](https://github.com/mike-kfed/warp/blob/http_basic_auth_cb/tests/auth_basic.rs)

It'd be great if we can work together on making that work with an async callback? or discuss a different API for building an auth filter? Anyways please let me know what you think

PR #741, opened 2020-11-06
Bump `tokio` from `0.2` to `0.3`

Urhengulas:
Bump `tokio` and related dependencies (`hyper`, `tokio-rustls`, `tokio-tungstenite`).

**Checklist**
- [x] get branch of @hawkw merged into `hyper` (https://github.com/hyperium/hyper/pull/2317)
    - [ ] hyper release with changes
- [x] get `tokio-tungstenite` to make a release with `tokio 0.3`
- [ ] update `bytes` to `0.6` for `warp`
- [x] update `bytes` to `0.6` for `hyper` (https://github.com/hyperium/hyper/pull/2339)

Fixes #725

PR #753, opened 2020-11-29
Upgrade to Tokio 1.0

### LINKED ISSUE #725: Upgrade to Tokio v1
**Dependencies which need to upgrade**
- [x] hyper (hyperium/hyper#2369)
- [x] tokio-rustls (tokio-rs/tls#46)
- [x] tokio-tungstenite (snapview/tokio-tungstenite#142)

aknuds1:
Similar to @Urhengulas' PR #741, although more complete, this PR aims to upgrade to Tokio v1.0.

I can successfully run all tests with all features enabled: `cargo test --all-features`.

This is my first time trying to contribute to warp, and I'm not super familiar with the internals, so please consider it a best effort :) 

Fixes #725.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,8 @@ jobs:
             benches: true
           - build: tls
             features: "--features tls"
-          - build: uds
-            features: "--features tokio/uds"
+          - build: net
+            features: "--features tokio/net"
           - build: no-default-features
             features: "--no-default-features"
           - build: compression
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,6 @@ jobs:
             benches: true
           - build: tls
             features: "--features tls"
-          - build: net
-            features: "--features tokio/net"
           - build: no-default-features
             features: "--no-default-features"
           - build: compression
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "3b987d825e9ff2a50c05b9f51345170d621df23f", features = ["brotli", "deflate", "gzip", "stream", "tokio-1"], optional = true }
+async-compression = { version = "0.3.7", features = ["brotli", "deflate", "gzip", "tokio"], optional = true }
 bytes = "1.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 headers = "0.3"
@@ -32,6 +32,7 @@ serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7"
 tokio = { version = "1.0", features = ["fs", "sync", "time"] }
+tokio-util = { version = "0.6", features = ["io"] }
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ tracing = { version = "0.1", default-features = false, features = ["log", "std"]
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { git = "https://github.com/snapview/tokio-tungstenite.git", rev = "71a5b72123db32b318d48964948fc76c943f1548", default-features = false, optional = true }
+tokio-tungstenite = { version = "0.12", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
 tokio-rustls = { version = "0.21", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ tracing = { version = "0.1", default-features = false, features = ["log", "std"]
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { git = "https://github.com/aknuds1/tokio-tungstenite.git", rev = "21c3d35b5ef91c2092dd67f6f2ada71d81c4b544", default-features = false, optional = true }
+tokio-tungstenite = { git = "https://github.com/aknuds1/tokio-tungstenite.git", rev = "cad06f3ac07e7fc2a936308207aab72b31478462", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
 tokio-rustls = { version = "0.22", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,12 +17,12 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "e4d903b8ff9972f056f714c61bd9d0f3321a4463", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
-bytes = "0.6"
+async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "3b987d825e9ff2a50c05b9f51345170d621df23f", features = ["brotli", "deflate", "gzip", "stream", "tokio-1"], optional = true }
+bytes = "1.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 headers = "0.3"
 http = "0.2"
-hyper = { git = "https://github.com/hyperium/hyper.git", rev = "1ba2a141a6f8736446ff4a0111df347c0dc66f6c", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
+hyper = { version = "0.14", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
 log = "0.4"
 mime = "0.3"
 mime_guess = "2.0.0"
@@ -31,7 +31,7 @@ scoped-tls = "1.0"
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7"
-tokio = { version = "0.3", features = ["fs", "stream", "sync", "time"] }
+tokio = { version = "1.0", features = ["fs", "sync", "time"] }
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
@@ -39,7 +39,7 @@ tower-service = "0.3"
 tokio-tungstenite = { version = "0.12", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
-tokio-rustls = { version = "0.21", optional = true }
+tokio-rustls = { version = "0.22", optional = true }
 
 [dev-dependencies]
 pretty_env_logger = "0.4"
@@ -47,7 +47,7 @@ tracing-subscriber = "0.2.7"
 tracing-log = "0.1"
 serde_derive = "1.0"
 handlebars = "3.0.0"
-tokio = { version = "0.3", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
 listenfd = "0.3"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ tracing = { version = "0.1", default-features = false, features = ["log", "std"]
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { git = "https://github.com/aknuds1/tokio-tungstenite.git", rev = "5c557a20aab6bdb343f8aa1d1b61d2f8b17c3312", default-features = false, optional = true }
+tokio-tungstenite = { version = "0.13", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
 tokio-rustls = { version = "0.22", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,12 +17,12 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { version = "0.3.1", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
-bytes = "0.5"
+async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "e4d903b8ff9972f056f714c61bd9d0f3321a4463", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
+bytes = "0.6"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 headers = "0.3"
 http = "0.2"
-hyper = { version = "0.13", features = ["stream"] }
+hyper = { git = "https://github.com/hyperium/hyper.git", rev = "1ba2a141a6f8736446ff4a0111df347c0dc66f6c", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
 log = "0.4"
 mime = "0.3"
 mime_guess = "2.0.0"
@@ -31,15 +31,15 @@ scoped-tls = "1.0"
 serde = "1.0"
 serde_json = "1.0"
 serde_urlencoded = "0.7"
-tokio = { version = "0.2", features = ["fs", "stream", "sync", "time"] }
+tokio = { version = "0.3", features = ["fs", "stream", "sync", "time"] }
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { version = "0.11", default-features = false, optional = true }
+tokio-tungstenite = { git = "https://github.com/snapview/tokio-tungstenite.git", rev = "71a5b72123db32b318d48964948fc76c943f1548", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
-tokio-rustls = { version = "0.14", optional = true }
+tokio-rustls = { version = "0.21", optional = true }
 
 [dev-dependencies]
 pretty_env_logger = "0.4"
@@ -47,7 +47,7 @@ tracing-subscriber = "0.2.7"
 tracing-log = "0.1"
 serde_derive = "1.0"
 handlebars = "3.0.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "0.3", features = ["macros", "rt-multi-thread"] }
 listenfd = "0.3"
 
 [features]
@@ -78,7 +78,6 @@ required-features = ["compression"]
 
 [[example]]
 name = "unix_socket"
-required-features = ["tokio/uds"]
 
 [[example]]
 name = "websockets"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,6 @@ all-features = true
 
 [dependencies]
 async-compression = { version = "0.3.7", features = ["brotli", "deflate", "gzip", "tokio"], optional = true }
-async-stream = "0.3"
 bytes = "1.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 headers = "0.3"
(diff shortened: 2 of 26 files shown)


Cargo.toml
```diff
@@ -17,12 +17,12 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { version = "0.3.1", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
-bytes = "0.5"
+async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "e4d903b8ff9972f056f714c61bd9d0f3321a4463", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
```
Urhengulas, line 20: Do you really need the `rev`, or is it just pointing to master?

src/filters/fs.rs
```diff
@@ -524,3 +525,33 @@ mod tests {
...
+    cx: &mut Context<'_>,
+    buf: &mut B,
+) -> Poll<io::Result<usize>> {
+    if !buf.has_remaining_mut() {
+        return Poll::Ready(Ok(0));
+    }
+
+    let n = {
+        let dst = buf.bytes_mut();
+        let dst = unsafe { &mut *(dst as *mut _ as *mut [MaybeUninit<u8>]) };
```
Urhengulas, line 540: Maybe add `Safety: `dst` may include uninitialized memory, but `ReadBuf` will never uninitialize memory that has already been initialized.` as comment? (See https://github.com/tokio-rs/tokio/discussions/3121)

Urhengulas (commented):
Good job! Only two remarks :D

Cargo.toml
aknuds1: It's more the other way around, master is pointing to this revision. At least it was, which is why I prefer pinning the revision, since a branch by its nature will change and you have no guarantee it will keep working.

The intent is also to switch to official versions of dependencies, before finalizing this PR. The current dependencies on Git commits just represent what I could make work.

src/filters/fs.rs
aknuds1: I just copied this code from tokio-util since I couldn't get the latter to work as a dependency. I'm more or less assuming that tokio-util will be used instead by the time this PR is finalized, although I don't know what the maintainers prefer.

src/filters/ws.rs
```diff
@@ -134,10 +134,7 @@ where
     fn into_response(self) -> Response {
         let on_upgrade = self.on_upgrade;
         let config = self.ws.config;
-        let fut = self
-            .ws
-            .body
-            .on_upgrade()
+        let fut = hyper::upgrade::on(Response::new(self.ws.body))
```
fundon, line 137: We can merge
`Response::new(self.ws.body)` and `let mut res = http::Response::default();`  to

```
let mut res   = Response::new(self.ws.body); 
let fut = hyper::upgrade::on(&mut res)
```

aknuds1: @seanmonstar Is it safe to replace `res` as suggested in the above comment? AFAICT, the difference is that `res` gets constructed from `self.ws.body`.

Cargo.toml
```diff
@@ -31,23 +31,23 @@ scoped-tls = "1.0"
...
 serde_json = "1.0"
 serde_urlencoded = "0.7"
-tokio = { version = "0.2", features = ["fs", "stream", "sync", "time"] }
+tokio = { version = "0.3", features = ["fs", "stream", "sync", "time"] }
 tracing = { version = "0.1", default-features = false, features = ["log", "std"] }
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { version = "0.11", default-features = false, optional = true }
+tokio-tungstenite = { git = "https://github.com/snapview/tokio-tungstenite.git", rev = "71a5b72123db32b318d48964948fc76c943f1548", default-features = false, optional = true }
```
jxs, line 39: `tokio-tungstenite` has released version `0.12` using tokio 0.3

aknuds1: @jxs Great, will update.

aknuds1: Done.

src/filters/fs.rs
jxs: `tokio-util` has since been upgraded to `0.5.1` which has `poll_read_buf`. I think it makes sense to use it and isolate the unsafe code there, but then maybe the `fs` filter should be feature gated as it now introduces an extra dependency.
What do you think @seanmonstar ?

aknuds1: Thanks for the update @jxs :)

jhpratt:
Tokio 1.0 has been released, so this PR should probably be updated to depend on that.

aknuds1:
I'm trying to upgrade to Tokio 1.0.

aknuds1:
The remaining issue is tokio-tungstenite. Trying to upgrade it to tokio 1.0, but it's easier said than done.

aknuds1:
Forked tokio-tungstenite, to upgrade it to tokio 1, and depending on it, but build still fails.

paolobarbolini:
> but build still fails

* `Sleep` is now `!Unpin`
* `bytes_mut` is now `chunk_mut`
* `bytes::buf::{Buf, BufMut}` -> `bytes::{Buf, BufMut}` (it's top-level now)

aknuds1:
If anyone has tips wrt. fixing the build failures, I'm all ears :)

Cargo.toml
```diff
@@ -17,12 +17,12 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { version = "0.3.1", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
-bytes = "0.5"
+async-compression = { git = "https://github.com/aknuds1/async-compression", rev = "3b987d825e9ff2a50c05b9f51345170d621df23f", features = ["brotli", "deflate", "gzip", "stream", "tokio-1"], optional = true }
```
paolobarbolini, line 20: This can now be upgraded to async-compression 0.3.7

aknuds1: @paolobarbolini Looks as if that version still depends on bytes 0.5 if you use the stream feature?

aknuds1: I tried building against it and couldn't make it work.

paolobarbolini: See https://github.com/seanmonstar/reqwest/pull/1076#discussion_r548536489

What you can do is something like [this](https://github.com/seanmonstar/reqwest/pull/1076/files#diff-668dbc2595c2151fbc2dde0df240f68963eb8e30582b5cbc71cf80da947d2800), where you wrap the encoder into `tokio_util::io::StreamReader` and `tokio_util::codec::FramedRead`

src/filters/body.rs
```diff
@@ -231,8 +231,8 @@ impl Decode for Json {
     const MIME: (mime::Name<'static>, mime::Name<'static>) = (mime::APPLICATION, mime::JSON);
     const WITH_NO_CONTENT_TYPE: bool = true;
 
-    fn decode<B: Buf, T: DeserializeOwned>(mut buf: B) -> Result<T, BoxError> {
-        serde_json::from_slice(&buf.to_bytes()).map_err(Into::into)
+    fn decode<B: Buf, T: DeserializeOwned>(buf: B) -> Result<T, BoxError> {
+        serde_json::from_slice(buf.chunk()).map_err(Into::into)
```
paolobarbolini, line 235: `Buf::chunk` may not return the entire buffer (see https://docs.rs/bytes/1.0.0/bytes/trait.Buf.html#tymethod.chunk)

```suggestion
        serde_json::from_slice(&buf.copy_to_bytes(buf.remaining())).map_err(Into::into)
```

Cargo.toml
aknuds1: @paolobarbolini OK, do you have any tips for the remaining failures?

paolobarbolini: [`alive_timer`](https://github.com/seanmonstar/warp/blob/5d068ae90f65cf66784c62daba8a27eaa506fa26/src/filters/sse.rs#L487) needs to be pinned, just like `event_stream` is (add `#[pin]` on top of it).

`Sleep::reset` doesn't take `self` anymore ([before](https://docs.rs/tokio/0.3.6/tokio/time/struct.Sleep.html#method.reset) - [after](https://docs.rs/tokio/1.0.0/tokio/time/struct.Sleep.html#method.reset)), so the two calls to `reset` need to be changed like so:
`first_argument.reset(second_argument)` -> `Sleep::reset(first_argument, second_argument)`

seanmonstar: You can still call it like a method, you just might need an `timer.as_mut().reset(arg)` in there.

aknuds1: Thanks, I fixed remaining build failures. I can look at async-compression next.

src/test.rs
```diff
@@ -92,16 +92,17 @@ use std::task::{self, Poll};
 
 use bytes::Bytes;
 #[cfg(feature = "websocket")]
```
paolobarbolini, line 94: This broke compiling without default features.

```suggestion
```

aknuds1:
@paolobarbolini @seanmonstar I could build with `--all-features` now, but some examples fail when I run tests.

aknuds1:
Is anyone able to help with the remaining build failure?

src/filters/ws.rs
fundon: I ported `ws.rs` to viz-rs. 
You can see that https://github.com/viz-rs/viz/blob/master/viz-core/src/ws.rs.

ebkalderon:
Seems like `cargo fmt` is complaining on the most recent commit. Doesn't seem to be a build failure.

src/filters/ws.rs
fundon: https://github.com/hyperium/hyper/blob/master/src/upgrade.rs#L258-L287

aknuds1:
With the kind help of @paolobarbolini I could fix all the build failures. Now some test failures remain AFAICT.

aknuds1:
Is anyone able to see why e.g. the integration test `echo_pings` fails?

fundon:
@ aknuds1  
https://github.com/hyperium/hyper/blob/master/src/upgrade.rs#L258-L287

```rust
hyper::upgrade::on(Response::new(self.ws.body));
```
The `Response::new(self.ws.body)` need include the `hyper::upgrade::OnUpgrade` from `req.extensions`, or `hyper::upgrade::on(req);`.

aknuds1:
@fundon Do you have a patch I can apply?

fundon:
@aknuds1  I ported `ws.rs` to https://github.com/viz-rs/viz/blob/master/viz-core/src/ws.rs. You can get it.

aknuds1:
Thanks @fundon! With the help of your code, I could figure out how to handle websocket upgrades.

aknuds1:
All the tests pass, it looks as if tokio/uds needs to be removed as a required check in the project settings if I'm not mistaken (since it's removed from tokio and .github/workflows/ci.yml).

Cargo.toml
```diff
@@ -31,23 +31,25 @@ scoped-tls = "1.0"
...
 tracing-futures = { version = "0.2", default-features = false, features = ["std-future"] }
 tower-service = "0.3"
 # tls is enabled by default, we don't want that yet
-tokio-tungstenite = { version = "0.11", default-features = false, optional = true }
+tokio-tungstenite = { git = "https://github.com/aknuds1/tokio-tungstenite.git", rev = "cad06f3ac07e7fc2a936308207aab72b31478462", default-features = false, optional = true }
 percent-encoding = "2.1"
 pin-project = "1.0"
-tokio-rustls = { version = "0.14", optional = true }
+tokio-rustls = { version = "0.22", optional = true }
+async-stream = "0.3"
```
Urhengulas, line 44: Maybe move `async-stream` to be 2nd dependency to keep alphabetical ordering?

Urhengulas (commented):
Good work :)

hasezoey:
@aknuds1 could you update this PR? because currently this dosnt build anymore:
```
    Updating git repository `https://github.com/nickelc/tungstenite-rs.git`
error: failed to get `tungstenite` as a dependency of package `tokio-tungstenite v0.12.0 (https://github.com/aknuds1/tokio-tungstenite.git?rev=cad06f3ac07e7fc2a936308207aab72b31478462#cad06f3a)`
    ... which is depended on by `warp v0.2.5 (https://github.com/aknuds1/warp?branch=chore/upgrade-tokio#f6508fa1)`
    ... which is depended on by `myproject v0.1.0 (/myproject)`

Caused by:
  failed to load source for dependency `tungstenite`

Caused by:
  Unable to update https://github.com/nickelc/tungstenite-rs.git?branch=bytes

Caused by:
  failed to find branch `bytes`

Caused by:
  cannot locate remote-tracking branch 'origin/bytes'; class=Reference (4); code=NotFound (-3)

```

aknuds1:
Thanks for the heads up @hasezoey, fixed.

apexys:
Is there anything we could do to help with this PR? 
We would really like to update our application to tokio 1.0 to use some of the newer features, but depend on warp.

aknuds1:
> Is there anything we could do to help with this PR?
> We would really like to update our application to tokio 1.0 to use some of the newer features, but depend on warp.

@apexys I'm waiting on tokio-tungstenite to release a new version, so I don't have to depend on a Git revision.

aknuds1:
I changed the dependency to the latest release of tokio-tungstenite now, i.e. 0.13, so the PR should be ready to merge.

mraerino:
@seanmonstar there seems to be a challenge around the required CI checks since one of them was renamed - you might need to remove the requirement for this to merge and add the new required check after merge

src/filters/compression.rs
```diff
@@ -56,7 +58,10 @@ pub struct Compression<F> {
 /// ```
 pub fn gzip() -> Compression<impl Fn(CompressionProps) -> Response + Copy> {
     let func = move |mut props: CompressionProps| {
-        let body = Body::wrap_stream(GzipEncoder::new(props.body));
+        let body = Body::wrap_stream(FramedRead::new(
```
jxs, line 61: why did you prefer `FramedRead` over `ReaderStream`?

aknuds1: @jxs To be honest, I don't remember exactly, but it was probably from an example. Do you think ReaderStream is preferable?

paolobarbolini: It was inspired from the reqwest implementation https://github.com/seanmonstar/reqwest/blob/a19eb3419637856d6d42f55dff6e4790613b2933/src/async_impl/decoder.rs#L47

aknuds1: Thanks @paolobarbolini, I can't bring myself to remember where I found this solution.

jxs: thanks @paolobarbolini. @aknuds1 I asked because I was not sure either, Previously `FrameReader` was the way to achieve `AsyncRead` to `Stream` conversion, but since `ReaderStream` [was introduced](https://github.com/tokio-rs/tokio/issues/2705) I would maybe suggest `ReaderStream` as it's simpler, and recommended for this cases on [tokio-util doc](https://docs.rs/tokio-stream/0.1.1/tokio_stream/#conversion-to-and-from-asyncreadasyncwrite)

jxs (approved):
LGTM thanks! :fireworks: 
- `Test (uds, --features tokio/uds)` also needs to be renamed on the repository settings
- should the `tokio_util` dependent filters (`fs`, `compression`) also be featured gated?

src/filters/compression.rs
aknuds1: @seanmonstar Do you have any opinion on this?

aknuds1: I converted from FramedRead into ReaderStream, ass @jxs recommended. I don't really know personally if there are any practical differences, but I can tell that the API is simpler (a second argument to `FramedRead::new` can be dropped). I trust @jxs knows more about this than I do, and I'll let @seanmonstar be the judge.

Cargo.toml
```diff
@@ -17,12 +17,12 @@ edition = "2018"
...
 [dependencies]
-async-compression = { version = "0.3.1", features = ["brotli", "deflate", "gzip", "stream"], optional = true }
-bytes = "0.5"
+async-compression = { version = "0.3.7", features = ["brotli", "deflate", "gzip", "tokio"], optional = true }
+bytes = "1.0"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 headers = "0.3"
 http = "0.2"
-hyper = { version = "0.13", features = ["stream"] }
+hyper = { version = "0.14", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
```
seanmonstar, line 25: Do we need the `client` feature?

src/filters/ws.rs
```diff
@@ -134,10 +140,11 @@ where
     fn into_response(self) -> Response {
         let on_upgrade = self.on_upgrade;
         let config = self.ws.config;
-        let fut = self
-            .ws
-            .body
-            .on_upgrade()
+        let mut resp = Response::new(self.ws.body);
+        if let Some(on_upgrade) = self.ws.on_upgrade {
+            resp.extensions_mut().insert(on_upgrade);
```
seanmonstar, line 145: If you already have the extension, you don't need to insert it here to remove it again. (`on_upgrade` is the same as `fut`).

seanmonstar (changes requested):
Excellent work! I believe this is very close.

.github/workflows/ci.yml
```diff
@@ -45,8 +45,8 @@ jobs:
             benches: true
           - build: tls
             features: "--features tls"
-          - build: uds
-            features: "--features tokio/uds"
+          - build: net
```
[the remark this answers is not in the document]
aknuds1: @seanmonstar OK, I can remove it.

Cargo.toml
aknuds1: @seanmonstar Tested now. It's required for the tests. If I add hyper to dev-dependencies though, with only the client feature enabled, they work. I'll make that change, let me know what you think.

src/filters/ws.rs
aknuds1: @seanmonstar I think I see what you mean, and try to simplify it to the best of my ability. Please have a look.

koivunej:
Should the return type of `warp::body::stream()` now have a stream of `Bytes` instead of `impl Buf` in: https://github.com/aknuds1/warp/blob/bb2f511fc8785784089ebdb12861bbc32e4cef5d/src/filters/body.rs#L77-L78

This following ... some Bytes change I cannot track down, which removed `Buf::to_bytes`.

paolobarbolini:
> This following ... some Bytes change I cannot track down, which removed `Buf::to_bytes`.

It was replaced by [Buf::copy_to_bytes](https://docs.rs/bytes/1.0.1/bytes/buf/trait.Buf.html#method.copy_to_bytes) @koivunej

Kiskae:
It appears 657f1c6 broke the build since `warp::test::handshake` uses `hyper::Client`:

```
   Compiling warp v0.2.5 (https://github.com/aknuds1/warp?branch=chore/upgrade-tokio#bb2f511f)
error[E0433]: failed to resolve: could not find `Client` in `hyper`
   --> /home/x/.cargo/git/checkouts/warp-2008e9c17b453427/bb2f511/src/test.rs:518:36
    |
518 |             let upgrade = ::hyper::Client::builder()
    |                                    ^^^^^^ could not find `Client` in `hyper`

```

aknuds1:
> It appears [657f1c6](https://github.com/seanmonstar/warp/commit/657f1c660ce582432ac884e029f6d3349cfac798) broke the build since `warp::test::handshake` uses `hyper::Client`:
> 
> ```
>    Compiling warp v0.2.5 (https://github.com/aknuds1/warp?branch=chore/upgrade-tokio#bb2f511f)
> error[E0433]: failed to resolve: could not find `Client` in `hyper`
>    --> /home/x/.cargo/git/checkouts/warp-2008e9c17b453427/bb2f511/src/test.rs:518:36
>     |
> 518 |             let upgrade = ::hyper::Client::builder()
>     |                                    ^^^^^^ could not find `Client` in `hyper`
> 
> error: aborting due to previous error
> ```

@Kiskae What's the command to reproduce your failure?

Kiskae:
@aknuds1 Calling `cargo build` with the branch as a dependency seems to trigger it:

```
[package]
name = "repro"
version = "0.1.0"
edition = "2018"

[dependencies]
warp = {git = "https://github.com/aknuds1/warp", branch = "chore/upgrade-tokio"}
```

aknuds1:
@Kiskae Aha, I see the problem, it's because src/test.rs contains public utilities and depends on hyper::Client.

nickelc:
> @Kiskae Aha, I see the problem, it's because src/test.rs contains public utilities and depends on hyper::Client. Guess I'll have to reintroduce the dependency on hyper's client feature @seanmonstar.
i think another option could be to make it an optional `test` feature that requires `hyper/client`

seanmonstar (approved):

aknuds1:
Thanks @seanmonstar! :tada:

PR #764, opened 2020-12-14
Add support MsgPack

kekcheburec:


seanmonstar:
Thanks for the PR!

I don't think MsgPack is quite ubiquitous enough to bundle it directly in warp. It seems like something that should be external. You could publish a crate providing the support.

kekcheburec:
> Thanks for the PR!
> 
> I don't think MsgPack is quite ubiquitous enough to bundle it directly in warp. It seems like something that should be external. You could publish a crate providing the support.

How can I add it using a separate crate. I just looked at the source code for Json and in fact I just had to copy and replace it with a crate for encoding and decoding MsgPack.
Yes, in few places, but it is convenient for embedded systems, where Json is quite cumbersome (in the same Rocket it is)

src/reply.rs
```diff
@@ -142,6 +142,83 @@ impl fmt::Display for ReplyJsonError {
...
+#[cfg(feature = "msgpack")]
+impl Reply for MsgPack {
+    #[inline]
+    fn into_response(self) -> Response {
+        match self.inner {
+            Ok(body) => {
+                let mut res = Response::new(body.into());
+                res.headers_mut().insert(
+                    CONTENT_TYPE,
+                    HeaderValue::from_static("application/msgpack"),
```
NickGeek, line 199: It looks like they also generally use `application/x-msgpack`

Issue #761, opened 2020-12-05
Form Example


WesleyAC:
Warp supports forms, but there is no example in the [examples](https://github.com/seanmonstar/warp/tree/master/examples) showing how to use them. A simple urlencoded form example would be quite useful.

PR #765, opened 2020-12-19
add form example #761

smallfish:


masklinn:
FWIW should probably clearly specify that the form is urlencoded (maybe provide HTTP endpoints to POST there as well to avoid the need for comments?): even though that's specified by the documentation for `warp::filters::body::form` we can't assume the reader will go check that out, especially since it takes no parameters or options. Otherwise there's 100% chance people will be confused when their `multipart/form-data` doesn't work properly.

PR #773, opened 2021-01-06
fix: use append instead of insert

### LINKED ISSUE #609: Multiple headers cannot be given with the same key for responses
Right now `warp::reply::with_header` gives us the ability to `.insert()` into the `HeaderMap` in a response. Unfortunately that overwrites any previous headers with the same key. https://docs.rs/http/0.2.1/http/header/struct.HeaderMap.html#method.insert There's a `.append()` method that might be a better default for `with_header` to use (or might not =), I don't know).

The use-case I have for this is `Set-Cookie` headers; there should be one per cookie. As a workaround I did some search/replace and made my own `append_header` based on `with_header` but using `append` instead of `insert`.

ajpauwels:
fixes #609
by using append, multiple cookies of the same name can be added.
this is required for some headers such as Set-Cookie in order to
set multiple cookies.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -386,7 +386,7 @@ impl<T: Reply> Reply for WithHeader<T> {
     fn into_response(self) -> Response {
         let mut res = self.reply.into_response();
         if let Some((name, value)) = self.header {
-            res.headers_mut().insert(name, value);
+            res.headers_mut().append(name, value);
         }
         res
     }


seanmonstar:
Thanks for the PR! I think changing this behavior might be surprising, since inserting is a replacing action, not appending. I suppose we could have a second method here... But it's also possible to just return an `http::Response`, and that would let someone use any of the methods on `HeaderMap`.

ajpauwels:
That's fair, it's easy enough to implement on your own. Should the issue be closed as unnecessary then?

PR #775, opened 2021-01-12
Add `close_frame` function to `ws::Message`

### LINKED ISSUE #774: Add function to obtain the close frame information from ws::Message
There is no way to obtain the close code or close reason from `ws::Message`. I learnt from this [discussion on discord](https://discord.com/channels/500028886025895936/671042482770935808/796029022281924650) that `Message::into_bytes` returns the close reason but this behavior isn't clear from documentation. Making matters worse, `Message::as_bytes` doesn't have the same behavior, instead [it returns an empty slice](https://github.com/seanmonstar/warp/blob/42fd14fdab8145d27ae770fe4b5c843a99bc2a44/src/filters/ws.rs#L342) when `Message` is of `Close` type.

A function to obtain the close frame information like: 
```rust
/// (u16, &str) = (close_code, close_reason)
fn close_frame(&self) -> (u16, &str)
```

It would also be nice to have `as_bytes` and `into_bytes` to behave the same way.

None available

**Additional context**
N/A

smallglitch:
(Partially) resolves #774

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -325,17 +325,11 @@ impl Message {
     }
 
     /// Try to get the close frame (close code and reason)
-    pub fn close_frame(&self) -> Result<Option<(u16, &str)>, ()> {
-        if let protocol::Message::Close(ref close_frame) = self.inner {
-            let code_and_reason = if let Some(close_frame) = close_frame {
-                Some((close_frame.code.into(), close_frame.reason.as_ref()))
-            } else {
-                None
-            };
-
-            Ok(code_and_reason)
+    pub fn close_frame(&self) -> Option<(u16, &str)> {
+        if let protocol::Message::Close(Some(ref close_frame)) = self.inner {
+            Some((close_frame.code.into(), close_frame.reason.as_ref()))
         } else {
-            Err(())
+            None
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/filters/ws.rs
```diff
@@ -324,6 +324,21 @@ impl Message {
         self.inner.is_pong()
     }
 
+    /// Try to get the close frame (close code and reason)
+    pub fn close_frame(&self) -> Result<Option<(u16, &str)>, ()> {
```
hamza1311, line 328: Why return a `Result<Option>, ()>` instead of `Option`? `()` doesn't mean anything.

smallglitch: To be consistent with [to_str()](https://github.com/seanmonstar/warp/blob/3bae3ecdad654e3c3145674a6387f2f1214e2ebe/src/filters/ws.rs#L342-L348)

hamza1311: `to_str` returns only a `Result`, not `Result<Option>>`. I don't think that's a good API but I'll leave that to @seanmonstar to decide for sure.

seanmonstar: I don't remember why `to_str` does that... I'm not sure it's the best pattern to follow.

Just an `Option` here seems fine.

hamza1311: Should both of the methods just return `Option` to be consistent? Or is just this one fine?

seanmonstar (approved):

seanmonstar:
Thanks so much!

PR #778, opened 2021-01-16
Replace to_str as as_text in ws::Message

hamza1311:
I did this according to [this discussion](https://canary.discord.com/channels/500028886025895936/671042482770935808/7997138255945400720) on discord.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -343,10 +343,19 @@ impl Message {
     }
 
     /// Try to get a reference to the string text, if this is a Text message.
+    #[deprecated(since = "0.3.0", note = "Use the as_text instead")]
     pub fn to_str(&self) -> Result<&str, ()> {
         match self.inner {
             protocol::Message::Text(ref s) => Ok(s),
             _ => Err(()),
+        }
+    }
+
+    /// Try to get a reference to the string text, if this is a Text message.
+    pub fn as_text(&self) -> Option<&str> {
+        match self.inner {
+            protocol::Message::Text(ref s) => Some(s),
+            _ => None,
         }
     }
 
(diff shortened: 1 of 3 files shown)


apiraino:
Hope it's ok if I take the liberty of adding a bit of context on the discussion for those not on said platform:

- **@_smallglitch**: `@seanmonstar` should we make the `to_str()` and `close_frame()` functions on `ws::Message` have a consistent return type (so both return an option; it would be a breaking change)? I forgot to ask if I should change that before I requested a re-review
- **@_seanmonstar**: hm, I kinda think if changing the return type, i'd also rename it to `as_text()`
- **@_seanmonstar**: in that case, we could just deprecate `to_str`, if we decide to make that change
- **@_seanmonstar**: `to_str` might be misleading anyways... it's just give a `&str` if it was a `Text` message

Therefore I think related to https://github.com/seanmonstar/warp/commit/f59de2490dd78ffd9d318928c13ba4c2238d9464

src/filters/ws.rs
```diff
@@ -343,13 +343,22 @@ impl Message {
     }
 
     /// Try to get a reference to the string text, if this is a Text message.
+    #[deprecated(since = "0.3.0", note = "Use the as_text instead")]
```
jxs, line 346: I don't think we need a deprecation note, we can just release a minor version without `to_str` to address this change :)

jxs (changes requested):
Hi, thanks! and sorry for the delay, only one minor suggestion

src/filters/ws.rs
hamza1311: Should I just remove `to_str` then?

jxs: yeah :)

jxs (approved):

PR #788, opened 2021-01-25
Update issue templates: convert "question" template to contact_links

masklinn:
Found that in the Rust project, replaces the "Get Started" button for Question (which opens a template containing the Discord link) by a direct link to the Discord server:
Screen Shot 2021-01-25 at 16 42 39
to
Screen Shot 2021-01-25 at 16 43 12

Also disables the "Open a blank issue" link in the footer.

Incidentally it looks like it might be possible to create a direct invite to a channel? 

I don't know if that only allows inviting people or if it can provide a link. In the latter case it could be better than linking to the server.


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,7 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Question
+    url: https://discord.gg/tokio
```
seanmonstar, line 4: The README includes a link directly to the channel: https://discord.gg/RFsPjyt

masklinn: Excellent, I'll use that then!

masklinn: Done, seems to work fine.

seanmonstar (approved):
Thanks!

PR #792, opened 2021-01-27
Fix: CORS: Allow spaces in Access-Control-Request-Headers values

mitjat:
### Bug description

The current implementation does not correctly interpret the `Access-Control-Request-Headers` value in a CORS preflight if it includes spaces.

For example, if

 * warp cors is configured with `.allow_headers([header::AUTHORIZATION, header::CONTENT_LENGTH])` and
 * preflight request includes `Access-Control-Request-Headers: content-length, authorization`,
 
then the preflight request will be rejected (HTTP 403) because the header value is split on ",", and `" authorization"` (note the space!) is not in the list of allowed headers.

### Fix

This PR adds a `.trim()` to the parsing process to allow for spaces in `Access-Control-Request-Headers` header values. It tweaks a test so that it fails without this PR's fix.

### Note

Technically, the standard prescribes that `Access-Control-Request-Headers` values should _not_ contain spaces -- see https://fetch.spec.whatwg.org/#cors-preflight-fetch, step 5. However
  1) The internet is rife with examples that do contain spaces, including the [official MDN documentation for the header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#examples)
  2) At least some existing http clients generate values with clients, e.g. [jsdom](https://www.npmjs.com/package/jsdom), a major Node library

So it seems like a good idea to be mildly lenient with parsing here.

### DIFF
--- a/src/filters/cors.rs
+++ b/src/filters/cors.rs
@@ -383,7 +383,7 @@ impl Configured {
                         .to_str()
                         .map_err(|_| Forbidden::HeaderNotAllowed)?;
                     for header in headers.split(',') {
-                        if !self.is_header_allowed(header) {
+                        if !self.is_header_allowed(header.trim()) {
                             return Err(Forbidden::HeaderNotAllowed);
                         }
                     }
--- a/tests/cors.rs
+++ b/tests/cors.rs
@@ -119,7 +119,7 @@ async fn success() {
     let res = warp::test::request()
         .method("OPTIONS")
         .header("origin", "https://hyper.rs")
-        .header("access-control-request-headers", "x-bar,x-foo")
+        .header("access-control-request-headers", "x-bar, x-foo")
         .header("access-control-request-method", "DELETE")
         .reply(&route)
         .await;


seanmonstar:
Interesting catch! It seems the Fetch spec even has a note call out about not using a space, "for better or worse". Do any browsers generate the header with space? Or do you have an example you ran into that does?

mitjat:
> Do any browsers generate the header with space? Or do you have an example you ran into that does?

The "Note" section above pretty much sums it up. I haven't tested with browsers. (Edit: I [tested](https://www.test-cors.org/#?client_method=POST&client_credentials=false&client_headers=foo%3A%20bar%0Aquu%3A%20zix&client_postdata=foobar&server_enable=true&server_status=200&server_credentials=true&server_methods=POST&server_headers=foo%2Cquu&server_tabs=local) with Firefox and Chrome just now, and they do both do the right thing, no spaces.)
I did run into this when using the `jsdom` library in javascript to generate requests against a `warp` server; it generates a header with spaces.

It's interesting how the fetch spec is self-aware about the spaces, yes :). I think the "generate conservatively, parse leniently" maxim applies pretty well here. Arguably, it's a `jsdom` bug (and it's worth fixing there also), but it behooves the server to be robust.

seanmonstar:
Yes, that's fair. I was just mainly curious if there were other bigger clients that were hitting this. Thanks!

seanmonstar (approved):

mitjat:
Thanks for merging, Sean!

I also wanted to let you know I really admire your responsiveness and exemplary tone of communication on issue trackers, forums etc. You explain your reasoning (even if it's tiny stuff like your last comment above), you're never condescending, you're willing to answer peripheral questions that border on supplemental tutorial material ... It makes reading warp-related stuff a pleasure. Thank you for making (this little corner of) the world a nicer place.

PR #799, opened 2021-02-05
Adds a very basic proxy server example

cars10:
As a new user of warp I struggled a bit while writing a small proxy server. 

Specifically the combination of `warp::method`, `warp::path::full`, `warp::header::headers_cloned` and `warp::body::bytes` to gain access to all the request information i need was not obvious to me. So after figuring it out (with the help of your discord) i wanted to contribute this example.

Please let me know if you have any issues with the code :)

### DIFF
--- a/examples/dyn_proxy.rs
+++ b/examples/dyn_proxy.rs
@@ -1,0 +1,59 @@
+#![deny(warnings)]
+use warp::hyper::body::{Body, Bytes};
+use warp::hyper::{Client, Request};
+use warp::{
+    http::{method::Method, HeaderMap, Response},
+    path::FullPath,
+    Filter, Rejection,
+};
+
+static PROXY_TARGET: &'static str = "http://httpbin.org";
+
+async fn proxy_request(
+    method: Method,
+    path: FullPath,
+    headers: HeaderMap,
+    body: Bytes,
+) -> Result<Response<Body>, Rejection> {
+    let request = build_request(&method, &path, &headers, body);
+    let client = Client::new();
+    let response = client.request(request).await.unwrap();
+    let response_status = response.status();
+    let response_body = response.into_body();
+
+    Ok(Response::builder()
+        .status(response_status)
+        .body(response_body)
+        .unwrap())
+}
+
+fn build_request(
+    method: &Method,
+    path: &FullPath,
+    headers: &HeaderMap,
+    body: Bytes,
+) -> Request<Body> {
+    let uri = format!("{}/{}", PROXY_TARGET, path.as_str());
+
+    let mut request = Request::builder().method(method.as_str()).uri(uri);
+
+    for (key, value) in headers {
+        request = request.header(key, value);
+    }
+
+    request.body(Body::from(body)).unwrap()
+}
+
+#[tokio::main]
+async fn main() {
+    let routes = warp::method()
+        .and(warp::path::full())
+        .and(warp::header::headers_cloned())
+        .and(warp::body::bytes())
+        .and_then(proxy_request);
+
+    println!("Proxy server to {} running.", PROXY_TARGET);
+    println!("Example request:");
+    println!("curl -X GET http://localhost:3030/ip");
+    warp::serve(routes).run(([127, 0, 0, 1], 3030)).await;
+}


renancloudwalk:
Looks useful to me.
I don't know what you guys think about this, but I always miss some testing on the sample files, I was also about to open a PR adding some testing to the examples, or at least one testing example by itself.

Also I don't understand why deny warnings on the examples (not only yours).

Thanks!

cars10:
> Looks useful to me.

Great :)

> I don't know what you guys think about this, but I always miss some testing on the sample files, I was also about to open a PR adding some testing to the examples, or at least one testing example by itself.

True. I really like that rocket has tests for most (or all?) examples in the repo. That being said though, in this example it is quite hard to write meaningful tests because it depends an an external service. You got any ideas on how to test this?

> Also I don't understand why deny warnings on the examples (not only yours).

I simply used it because it is also present on all the other tests and i wanted to stick to the same style. As i could not find any contributing or style guidelines  :)

Thanks for the feedback!

examples/dyn_proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    request.body(Body::from(body)).unwrap()
+}
+
+#[tokio::main]
+async fn main() {
+    let routes = warp::method()
+        .and(warp::path::full())
+        .and(warp::header::headers_cloned())
+        .and(warp::body::bytes())
```
seanmonstar, line 52: You could use `warp::body::steam()`, so you don't need to buffer the body in the proxy, but just stream it along.

examples/dyn_proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+    method: &Method,
+    path: &FullPath,
+    headers: &HeaderMap,
+    body: Bytes,
+) -> Request<Body> {
+    let uri = format!("{}/{}", PROXY_TARGET, path.as_str());
+
+    let mut request = Request::builder().method(method.as_str()).uri(uri);
+
+    for (key, value) in headers {
```
seanmonstar, line 40: If you skip using the request builder, you can assign the full headers instead having to iterate here. Like: `*req.headers_mut() = headers;`.

examples/dyn_proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+    headers: HeaderMap,
+    body: Bytes,
+) -> Result<Response<Body>, Rejection> {
+    let request = build_request(&method, &path, &headers, body);
+    let client = Client::new();
+    let response = client.request(request).await.unwrap();
+    let response_status = response.status();
+    let response_body = response.into_body();
+
+    Ok(Response::builder()
```
seanmonstar, line 24: You may wish to include the response headers as well.

examples/dyn_proxy.rs
```diff
@@ -0,0 +1,59 @@
...
+
+async fn proxy_request(
+    method: Method,
+    path: FullPath,
+    headers: HeaderMap,
+    body: Bytes,
+) -> Result<Response<Body>, Rejection> {
+    let request = build_request(&method, &path, &headers, body);
+    let client = Client::new();
+    let response = client.request(request).await.unwrap();
```
seanmonstar, line 20: It's likely eventually there will be some network error sending the request, it'd probably be better to return a 503 response in that case. Otherwise, the unwrap will crash the proxy server.

seanmonstar (changes requested):
Cool, thanks for submitting! The example seems simple and useful to learn from. Happy to add this.

examples/dyn_proxy.rs
cars10: Great, i actually tried to do that but couldn't work it out. Thanks, fixed

examples/dyn_proxy.rs
cars10: Done :)

examples/dyn_proxy.rs
cars10: Good catch, fixed that.

examples/dyn_proxy.rs
cars10: I tried to use `warp::body::stream()` here, but i could not figure out how to send a `client.request` with a `Stream`. It seems that `client.requests` expects a type that implements `HttpBody + Send + 'static` but i could not convert the stream to work it out. I might be missing something here?

cars10:
> Cool, thanks for submitting! The example seems simple and useful to learn from. Happy to add this.

Thanks for the great response. Really helped me in better understanding warp and improving the example.

examples/dyn_proxy.rs
jxs: Hi @cars10,
something like:
```Rust
fn build_request(
    method: Method,
    path: FullPath,
    headers: HeaderMap,
    body: impl Stream<Item = Result<impl Buf, Error>> + Send + 'static,
) -> Request<Body> {
    let uri = format!("{}/{}", PROXY_TARGET, path.as_str());

    let body = body.map_ok(|mut buf| buf.copy_to_bytes(buf.remaining()));
    let mut request = Request::new(Body::wrap_stream(body));
    *request.method_mut() = method;
    *request.uri_mut() = uri.parse().unwrap();
    *request.headers_mut() = headers;
    request
}
```

cars10: Done! Thanks for the hint.

jxs (approved):

PR #800, opened 2021-02-05
Allow construction of websocket pong messages

### LINKED ISSUE #782: Allow construction of websocket pong messages
I am writing websocket server and I'm mocking the `warp::filters::ws::WebSocket` with a custom type implementing `Sink<Message>` and `Stream<Item=Result<Message, Error>>`. To be able to test the timeout behavior of the code I need to manually respond to Ping messages with a Pong message. Currently the API doesn't provide a way to construct a Pong message.

I'd like to be able to construct a [`Message`](https://docs.rs/warp/0.2.5/warp/filters/ws/struct.Message.html) representing a pong message. The API could be very similar to the existing `ping` function. For example the following (untested) code could be used:
```rust
/// Construct a new Pong `Message`.
///
/// Note that this is rarely needed as Ping messages are automatically responded with the corresponding Pong message by
/// the underlying socket. This is mainly provided for some unconventional use cases, like writing mock implementations
/// for tests
pub fn pong<V: Into<Vec<u8>>>(v: V) -> Message {
    Message {
        inner: protocol::Message::Pong(v.into()),
    }
}
```

An alternative could be to provide a `From` implementation for `tungstenite::protocol::Message`.

henkkuli:
Fixes #782

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -291,6 +291,17 @@ impl Message {
         }
     }
 
+    /// Construct a new Pong `Message`.
+    ///
+    /// Note that one rarely needs to manually construct a Pong message because the underlying tungstenite socket
+    /// automatically responds to the Ping messages it receives. Manual construction might still be useful in some cases
+    /// like in tests or to send unidirectional heartbeats.
+    pub fn pong<V: Into<Vec<u8>>>(v: V) -> Message {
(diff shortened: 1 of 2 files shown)


jxs (approved):

seanmonstar (approved):
Thank you!

PR #801, opened 2021-02-09
Use append instead of insert when adding headers using WithHeader

fabianboesiger:
This change allows to set multiple header values with the same key using `warp::reply::with_header`. Previously, the old value was overwritten by the new one.


jxs:
Hi, see https://github.com/seanmonstar/warp/pull/773#issuecomment-755787779

fabianboesiger:
Oh, sorry I didn't see that PR.

PR #802, opened 2021-02-10
Implemented Reply for a infallible result

comath:
`http::Error` implements `From<Infallible>`, this just shortcuts that so we can have an easy time with infallible results.

Another option is to implement the following:
```
impl<T,E> Reply for Result<T, E>
where
    T: Reply + Send,
    E: Reply + Send,
{ ... }
```

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -435,11 +435,7 @@ where
     fn into_response(self) -> Response {
         match self {
             Ok(t) => t.into_response(),
-            Err(e) => {
-                // Maybe panic?
-                tracing::error!("Infallible Error: {:?}", e);
-                StatusCode::INTERNAL_SERVER_ERROR.into_response()
-            }
+            Err(e) => match e {}
         }
     }
 }
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -427,6 +427,23 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/reply.rs
```diff
@@ -427,6 +427,23 @@ where
...
+{
+    #[inline]
+    fn into_response(self) -> Response {
+        match self {
+            Ok(t) => t.into_response(),
+            Err(e) => {
+                // Maybe panic?
+                tracing::error!("Infallible Error: {:?}", e);
+                StatusCode::INTERNAL_SERVER_ERROR.into_response()
+            }
```
MikailBag, line 442: If I understand correctly, this should work better
```suggestion
            Err(e) => match e {}
```

comath: I did not know that about `Infallible`, I changed it to what you suggested.

kaj:
For reference: The "other option", `impl Reply for Result<T, E> where T: Reply, E: Reply`, is suggested in #458 .

comath:
> For reference: The "other option", `impl Reply for Result<T, E> where T: Reply, E: Reply`, is suggested in #458 .

I should have looked around some more, sorry :). I thought it'd be much much more complicated than this simple addition.

jxs (approved):

Issue #815, opened 2021-03-09
Range: bytes header for warp::fs::dir hosted files doesn't receive last chunk


jaredwolff:
**Version**
warp v0.3.0

**Platform**
FreeBSD pyrinas-dev.jaredwolff.com 12.2-RELEASE-p1 FreeBSD 12.2-RELEASE-p1 GENERIC  amd64

**Description**

Not sure if this should go here or to the underlying `hyper` implementation but I'll give it a shot here first.

I'm currently hosting some static binaries that can be downloaded by embedded devices for OTA updates. One of the important pieces of functionality is the need for partial downloads (say 2048 byte chunks at a time). While this, mostly, works, I've run into an issue related to retrieving the last chunk. 

For instance I know that the file located at  `https://pyrinas-dev-ota.jaredwolff.com/images/352656102712240.bin` is 200595 bytes in size. (Btw, it's a test binary on publicly available code)

When I attempt to pull the last section using curl I get a 416 error: (I originally got this error in embedded which led me to investigate more.)

```
curl/7.73.0/bin/curl https://pyrinas-dev-ota.jaredwolff.com/images/352656102712240.bin -H "Content-Type: application/octet-stream" -H "Range: bytes=198656-200595" --output bob.bin
```

But, when I reduce the size of the full file, it retrieves the last 1939 bytes successfully. 

```
/usr/local/Cellar/curl/7.73.0/bin/curl https://pyrinas-dev-ota.jaredwolff.com/images/352656102712240.bin -H "Content-Type: application/octet-stream" -H "Range: bytes=198656-200594" --output bob.bin
```

I am running this through a reverse proxy (Caddy). I did double check that the Caddy implementation is not funky by doing the same thing with a statically hosted file elsewhere. (Same Caddy instance, different domain) No problems though with the .png I was hosting.

```
curl  https://www.jaredwolff.com/testing-hardware-using-rust/images/Screen_Shot_2021-01-06_at_10.30.55_AM.png -H "Range: bytes=125801-125808" --output test.bin
```

All the above files are available for anyone to test this. Seems like the implementation is using a Range indexed from 0 rather than from 1 which may explain why reducing to 200594 worked.

By the way here's what the code to host the files looks like:

```rust
    // Only one folder that we're interested in..
    let images = warp::path("images").and(warp::fs::dir(&IMAGE_FOLDER_PATH));

    // Run the `warp` server
    warp::serve(images)
        .run(SocketAddrV4::new(
[...]

PR #818, opened 2021-03-15
filters: fs: end range parameter edge case

jaredwolff:
This fix originates from #815. 

This fixes the edge case where  the bound == file size. 
By default the bound is always incremented by one
but if it is equal to the file size it should *not* increment.

### DIFF
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -375,7 +375,14 @@ fn bytes_range(range: Option<Range>, max_len: u64) -> Result<(u64, u64), BadRang
 
             let end = match end {
                 Bound::Unbounded => max_len,
-                Bound::Included(s) => s + 1,
+                Bound::Included(s) => {
+                    // For the special case where s == the file size
+                    if s == max_len {
+                        s
+                    } else {
+                        s + 1
+                    }
+                }
                 Bound::Excluded(s) => s,
             };
 
--- a/tests/fs.rs
+++ b/tests/fs.rs
@@ -229,3 +229,43 @@ async fn byte_ranges() {
     assert_eq!(res.headers()["content-length"], contents.len().to_string());
     assert_eq!(res.headers().get("content-range"), None);
 }
+
+#[tokio::test]
+async fn byte_ranges_with_excluded_file_size() {
+    let _ = pretty_env_logger::try_init();
+
+    let contents = fs::read("README.md").expect("fs::read README.md");
+    let file = warp::fs::file("README.md");
+
+    // range including end of file (non-inclusive result)
+    let res = warp::test::request()
+        .header("range", format!("bytes=100-{}", contents.len()))
+        .reply(&file)
+        .await;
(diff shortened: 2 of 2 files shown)


jxs:
Hi, and thanks for your interest!
Hum, shouldn't this check be on the `headers` crate, as you showed on https://github.com/seanmonstar/warp/issues/815#issuecomment-794660948 so that we then get the `end` as `Bound::Excluded` ?

jaredwolff:
@jxs thanks for your input. Seems like the `parse_bound` function defaults to `Bound::Included` unless I'm wrong here. I don't see where `Bound::Excluded` is used.

jxs:
no no, you are right, what I was asking/wondering was if it shouldn't be on `parse_bound` that `end` should be considered `Bound::Excluded`

jaredwolff:
Ahh duh. Sorry I didn't understand what you were saying at first. 

If there's a way to know the `max_len` i.e. file size within the `headers` lib then I'm game but I don't think there is. You need to know the file size that way you can determine whether or not the end bound is incremented or left alone. That information is only available within `warp` itself. But if there's a crafty way of doing it I'm all ears. :)

seanmonstar:
I don't think the problem is the inclusive handling right there, the spec (RFC 7233) says that bytes ranges are always inclusive. I think the problem is an off-by-one error at https://github.com/seanmonstar/warp/blob/6d7298c084a25e7446a995539fc99683a40b19fa/src/filters/fs.rs#L418. For example, the range `50-100` means bytes 50 through 100 inclusive, which is 51 bytes. But that line would calculate 50 as the length. But, just fixing that one line might make it stop working when using just `max_len`...

jaredwolff:
Hey @seanmonstar thanks for weighing in. I know you're a busy guy!

I did play with different `Range` header values and it's operating as expected. (Binary image is pulled down in 2048 byte intervals). Happy to modify the PR and test what's here if there's a cleaner solution.

seanmonstar:
So, what I mentioned doesn't happen? `bytes=50-100` downloads 50 bytes or 51?

jaredwolff:
51 bytes.

seanmonstar:
Hm..... *scratches chin*... time to stare at it some more...

seanmonstar:
Well, I suppose for peace of mind, do you think we could add a few unit tests (preferably a new function) at the bottom of `tests/fs.rs`, that exercises these few cases? Especially helpful to have a test that failed before, and with this fix it works.

jaredwolff:
@seanmonstar yea let me whip something up.

jaredwolff:
Tested with the before and after. Before this fix it fails (as expected) and with the fix it passes that particular unit test. Ran all the others for good measure. All good there. Let me know if you want me to tweak anything else!

seanmonstar (approved):
Hey, thanks for finding a fix and getting us some tests!

jaredwolff:
Woo. Thanks @seanmonstar

PR #819, opened 2021-03-16
Add warp::auth::basic() filter

586837r:
You use it like this:
```rust
use warp::Filter;

let auth = warp::auth::basic()
    .realm("access")
    .allow("user", "1234");

let route = warp::any()
    .map(warp::reply)
    .with(auth);
```

#751 also implements basic authentication.
    
Contrary to theirs, mine:
- returns an `impl Reply` instead of a rejection on authorization failure
- uses an immutable array of authorizations
- has a very simple and limited api
- realm is optional


mike-kfed:
I like the implementation using builder pattern and usage as a `with()` filter.

Only objections are

- the user/password storage is not encrypted, therefore unsafe (also not sure this should be part of warp, something that can for example authenticate against htpasswd files could/should be an extra crate)
- no way to use a custom lookup function for e.g. a user/pass database (or when used for `Bearer` auth to verify tokens)

gralpli:
Just an additional thought: the password comparison is not constant-time and might therefore be vulnerable to timing attacks.

PR #821, opened 2021-03-20
Upgrade `Message` To An Enum

danii:
I saw the old documentation said something about upgrading this struct to a non exhaustive enum in the future, plus I prefer matching over something like this as an enum, so, I made it an enum.

This shouldn't cause any breaking changes as far as I'm aware. Someone could've wrote something like `let Message {..} = some_message;`, but really that has no practical use in any context, as such I highly doubt anyone wrote that in any production code.

This should IMO improve usability of `Message`s by quite a lot.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -203,7 +203,7 @@ impl Stream for WebSocket {
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         match ready!(Pin::new(&mut self.inner).poll_next(cx)) {
-            Some(Ok(item)) => Poll::Ready(Some(Ok(Message { inner: item }))),
+            Some(Ok(item)) => Poll::Ready(Some(Ok(Message::from_protocol(item)))),
             Some(Err(e)) => {
                 tracing::debug!("websocket poll error: {}", e);
                 Poll::Ready(Some(Err(crate::Error::new(e))))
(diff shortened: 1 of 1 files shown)


src/filters/ws.rs
```diff
@@ -297,90 +325,78 @@ impl Message {
...
     /// like in tests or to send unidirectional heartbeats.
     pub fn pong<V: Into<Vec<u8>>>(v: V) -> Message {
-        Message {
-            inner: protocol::Message::Pong(v.into()),
-        }
+        Message::Pong(v.into())
     }
 
     /// Construct the default Close `Message`.
     pub fn close() -> Message {
```
jxs, line 332: I'd just maintain the old behavior wdyt? [You can close the socket without it being abruptly](https://docs.rs/tungstenite/0.13.0/tungstenite/protocol/struct.WebSocket.html#method.close)

src/filters/ws.rs
```diff
@@ -297,90 +325,78 @@ impl Message {
...
-            inner: protocol::Message::Close(Some(protocol::frame::CloseFrame {
-                code: protocol::frame::coding::CloseCode::from(code.into()),
-                reason: reason.into(),
-            })),
-        }
+        Message::Close(code.into(), reason.into())
     }
 
     /// Returns true if this message is a Text message.
     pub fn is_text(&self) -> bool {
```
jxs, line 342: on one side I'd just deprecate this old methods as one can just do `matches!` now, what do you also think @seanmonstar ?

jxs (changes requested):
Hi, and thanks for the PR! only a couple minor things left

danii:
Hello, I apologize for my procrastination. I know it isn't professional.

I see that the pull request #903 has been made which does the same, is it the plan to close mine and merge the new one? If so shall I close mine?

PR #827, opened 2021-03-29
Add pub fn to get a Response from a Rejection

jtroo:
This is inspired from my desire to easily turn a Rejection into a Response so that my recover handling code is easier. There are some routes where I want to turn all rejections into replies and not fall back to other routes, and I don't want to handle all the built-in rejections manually.

Example:

    /// My custom error type
    pub enum Error {
        // ... snip
    }

    impl warp::reject::Reject for Error {}

    pub async fn handle_rejection(
        err: Rejection
    ) -> std::result::Result<warp::reply::Response, std::convert::Infallible> {
        if let Some(e) = err.find::<Error>() {
            // handle my custom rejection
        } else {
            Ok(err.default_response())
        }
    }

My original comment:
https://github.com/seanmonstar/warp/issues/712#issuecomment-808990088

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -317,6 +317,30 @@ impl Rejection {
             false
         }
     }
+
+    /// Get the default `Response` for this `Rejection`. With a custom `Rejection`, this will
+    /// always return a response with status code `INTERNAL_SERVER_ERROR`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let rejection = warp::reject();
+    /// assert_eq!(rejection.default_response().status(), warp::http::StatusCode::NOT_FOUND);
+    ///
+    /// #[derive(Debug)]
+    /// struct MyCustomErr;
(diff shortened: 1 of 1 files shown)


jtroo:
Could also change this part of the [rejections example](https://github.com/jtroo/warp/blob/master/examples/rejections.rs) to use this fn. Any thoughts?

```
    } else {
        // We should have expected this... Just log and say its a 500
        eprintln!("unhandled rejection: {:?}", err);
        code = StatusCode::INTERNAL_SERVER_ERROR;
        message = "UNHANDLED_REJECTION";
    }
```

jtroo:
Just realized that there was another open PR that accomplishes the same thing but using the `Into` trait: https://github.com/seanmonstar/warp/pull/820

Either one would work for me. The approach in this PR has the benefit of the documentation to warn users about the behaviour of custom rejections.

Issue #834, opened 2021-04-19
HEAD request support for fs::dir() filter


zenria:
Hi, 

filters::fs::dir() only reply to GET requests, it should also reply to HEAD requests.

joseluisq:
I think this is related to or possibly dup of https://github.com/seanmonstar/warp/issues/532.

PR #835, opened 2021-04-19
Allow HEAD requests to be served by fs::dir() filter

zenria:
This is a fix for feature request #834

### DIFF
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -81,6 +81,8 @@ pub fn file(path: impl Into<PathBuf>) -> impl FilterClone<Extract = One<File>, E
 pub fn dir(path: impl Into<PathBuf>) -> impl FilterClone<Extract = One<File>, Error = Rejection> {
     let base = Arc::new(path.into());
     crate::get()
+        .or(crate::head())
+        .unify()
         .and(path_from_tail(base))
         .and(conditionals())
         .and_then(file_reply)


joseluisq:
How about the empty body for `HEAD` responses? https://tools.ietf.org/html/rfc7231#section-4.3.2

zenria:
The underlying HTTP server (not sure if it's hyper or tower) does not sent the body when handling HEAD requests.

This has already been done in the fs::file() filter: 

https://github.com/seanmonstar/warp/blob/4a03710adcc7acee2c773a97d25ffd3de18efea9/src/filters/fs.rs#L48-L57

joseluisq:
Great.
Another question does the `file` need `HEAD` method support too or not?

zenria:
I think, yes it needs it, but this is already working due to the any() filter. 

I didn't want to use the any() filter for the fs:dir() as it appears cumbersome to respond to POST or PUT requests on static files. So I think we should also disable this on the fs:file() filter and use only get+head filter

joseluisq:
> I think, yes it needs it, but this is already working due to the any() filter.

Do you mean this is working for your addition or do you mean this `file` filter is already working that's why you don't want to break someone else code?

But yeah I'm also wondering in what cases we want to support `fs::file()` with any method.

zenria:
> > I think, yes it needs it, but this is already working due to the any() filter.
> 
> Do you mean this is working for your addition or do you mean this `file` filter is already working that's why you don't want to break someone else code?

the second option: I did not want to break someone else code!

> 
> But yeah I'm also wondering in what cases we want to support `fs::file()` with any method.

:-)

jxs (approved):

jxs:
thanks @zenria! and thanks @joseluisq for the reviewing :)

>The underlying HTTP server (not sure if it's hyper or tower) does not sent the body when handling HEAD requests.

yeah it's hyper, [see here](https://github.com/hyperium/hyper/blob/master/src/proto/h1/role.rs#L831)

amousset:
FYI this change actually broke some code when updating from 0.3.1 to 0.3.2. We had a custom handler for HEAD but relied on `fs::dir()` for GET requests (something probably quite unusual I admit :) ). Adding a `method::get()` filter before the `fs::dir()` fixed the problem.

PR #839, opened 2021-04-27
Add docs for ping handling

0xdeafbeef:
It's super confusing, that we should manually poll rx part of the socket to make pings handled automatically.

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -10,7 +10,7 @@ use super::header;
 use crate::filter::{filter_fn_one, Filter, One};
 use crate::reject::Rejection;
 use crate::reply::{Reply, Response};
-use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt};
+use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt, StreamExt};
 use headers::{Connection, HeaderMapExt, SecWebsocketAccept, SecWebsocketKey, Upgrade};
 use http;
 use hyper::upgrade::OnUpgrade;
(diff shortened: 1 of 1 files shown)


jxs:
Hi, and thanks for the submission. I think this is rather implicit by the poll nature of Rust's futures. If you don't poll to read, there's no way `tungstenite` can reply. wdyt about rephrasing the note more like that?

src/filters/ws.rs
```diff
@@ -10,7 +10,7 @@ use super::header;
 use crate::filter::{filter_fn_one, Filter, One};
 use crate::reject::Rejection;
 use crate::reply::{Reply, Response};
-use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt};
+use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt, StreamExt};
```
jxs, line 13:
```suggestion
use futures::{future, ready, FutureExt, Sink, Stream, StreamExt, TryFutureExt};
```

src/filters/ws.rs
```diff
@@ -10,7 +10,7 @@ use super::header;
 use crate::filter::{filter_fn_one, Filter, One};
 use crate::reject::Rejection;
 use crate::reply::{Reply, Response};
-use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt};
+use futures::{future, ready, FutureExt, Sink, Stream, StreamExt, TryFutureExt};
```
jxs, line 13:
```suggestion
use futures::{future, ready, FutureExt, Sink, Stream, TryFutureExt};
```

jxs (approved):

jxs:
thanks!

PR #841, opened 2021-05-01
Add ext provider

arniu:


Aankhen:
Thank you for reworking the MR! I’m looking forward to using it once it’s merged.

paulzhang5511:
Good.

jxs:
Hi, thanks for the submission!   
I know you replied to my questions on the Discord channel, but looking into the example I still can't understand the advantage from the `with_db` filter used on the [todos example](https://github.com/seanmonstar/warp/blob/master/examples/todos.rs#L100). I can understand that you are able to `get` and `provide` the values on different places of the route definition but how is that helpful, can you describe?
thanks :)

dcormier:
This would help me a lot. I want to be able to put something on the "request context" (for lack of a better term) and have it optionally available at a request handler. I have not been able to find a way to do that. This seems perfect.

arniu:
@jxs 1. With `ext::provide`, we do not write `with_db` any more, and do not have to pass `db` to filters one by one; 2. Without `ext::provide`, `ext::get` is useless.

PR #843, opened 2021-05-02
docs: Add examples to query

### LINKED ISSUE #657: Optional query parameters seem either very convoluted or conceringly missing
I've found issues in the past referencing how to possibly get this done like #86; however, this seems like an incredibly heavy-handed and honestly hacky solution to do something fairly standard in most web frameworks I've dealt with. The original author of that issue even pointed out that it seemed too complicated, but it was closed and moved on. 

This becomes especially increasingly frustrating as you add more optional query parameters and your code forks based on how they are provided; i.e. why should I have to create 5 different handler functions based on the parameters supplied when I can just have one that figures out what to do with the parameters supplied? 

Why can't we just create an endpoint such as `/foo` with optional query parameters `bar` and `baz` as such:
```
warp::path("foo").and(warp::query())
        .map(|options: Option<QueryOptions>| match names {
            Some(opt) => format!("bar: {}, baz: {}", opt.bar, opt.baz),
            None => "no query params provided".to_owned(),
        })
```

Thinking of an instance where I have some grouping of optional query parameters, like for example, pagination query parameters. If I supply `size`, I would like to also enforce that `page_number` or something else be provided as well. Thus trying to group everything together under something like `QueryOptions` instead of just trying to figure out what to do if someone provides only `size` or only some other thing and figuring out how to proceed for each of those. 

Per the title, I can't tell if this is just missing support in warp right now, or this is more complicated than I think it reasonably should be to figure out.

Thanks

FlorianDr:
Fixes #657

### DIFF
--- a/src/filters/query.rs
+++ b/src/filters/query.rs
@@ -10,6 +10,58 @@ use crate::reject::{self, Rejection};
 /// Creates a `Filter` that decodes query parameters to the type `T`.
 ///
 /// If cannot decode into a `T`, the request is rejected with a `400 Bad Request`.
+///
+/// # Example
+///
+/// ```
+/// use std::collections::HashMap;
+/// use warp::{
+///     http::Response,
+///     Filter,
+/// };
+///
+/// let route = warp::any()
+///     .and(warp::query::<HashMap<String, String>>())
+///     .map(|map: HashMap<String, String>| {
+///         let mut response: Vec<String> = Vec::new();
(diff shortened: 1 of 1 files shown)


jxs:
Hi, thanks! There seems to be some issues with examples, can you take a look?

jxs (approved):

jxs:
thanks!

Issue #323, opened 2019-11-29
Multipart should Stream File Upload


cetra3:
Having a look through the multipart filter source, it looks like it loads the entire file into memory:

https://github.com/seanmonstar/warp/blob/5c269562a823c5340f3dfc14bdd11af592c03dea/src/filters/multipart.rs#L112-L116

Maybe returning a `Stream` of `Bytes` would be more appropriate than a `Vec<u8>`, especially for larger files.  I can submit a PR to make the change for you if there is interest in changing this.

seanmonstar:
Yea, as part of the initial implementation, the easiest was to just buffer the full contents. I certainly want to provide a way to prevent that. The tricky part that I punted on was coming up with a decent API.

cetra3:
Ok, so I have had a look at the current multipart code & there are a couple of issues:

* The `multipart` crate is not async as far as I can tell, and `read_entry` is blocking when reading in fields.  If you are reading in 1GB of data this is an issue.

* Personal taste but setting a limit of 2MB for a multipart upload as a default is quite "magical".  Other libraries I have used that set these sorts of limits by default cause a lot of pain in production.  I would advocate no limit unless set

surban:
Is there a workaround we could currently use for streaming file uploads?

cetra3:
@surban I have been working on this slightly with a branch in another lib called [mpart-async](https://github.com/cetra3/mpart-async). 

I have added a [warp example](https://github.com/cetra3/mpart-async/blob/8c69108344ee3f5795bfa507e941b87198394da5/examples/warp.rs#L19-L42) which is still not as ergonomic as I'd like, but it is sort of where I am heading with this:

```rust
async fn mpart(
    mime: Mime,
    body: impl Stream<Item = Result<impl Buf, warp::Error>> + Unpin,
) -> Result<impl warp::Reply, Infallible> {
    let boundary = mime.get_param("boundary").map(|v| v.to_string()).unwrap();

    let mut stream = MpartStream::new(boundary, body.map_ok(|mut buf| buf.to_bytes()));

    while let Ok(Some(mut field)) = stream.try_next().await {
        println!("Field received:{}", field.name().unwrap());
        if let Ok(filename) = field.filename() {
            println!("Field filename:{}", filename);
        }

        while let Ok(Some(bytes)) = field.try_next().await {
            println!("Bytes received:{}", bytes.len());
        }
    }

    Ok(format!("Thanks!\n"))
}
```

You can test with `curl`:

[...]

PR #846, opened 2021-05-08
Reimplement multipart and make it stream the file upload

paolobarbolini:
Hello, seeing how long #323 has been open I decided to give it a go myself. I should probably have asked in the issue or on Discord, but I really wanted to give this issue a shot, so here I am.

This rewrites the multipart filter to do all of the multipart decoding by itself, dropping the dependency on the `multipart` crate. This reduces the number of dependencies, and makes it so the body is streamed instead of having to read it entirely into memory (see #323).

No changes have been done to the public API. The only breaking change will be observed by implementations that try to read from `Part`s obtained before the last call to `FormData::poll_next`. In those cases an error will be returned.

The decoder is mostly zero copy. The only case in which data is copied is when the underlying body yields very small chunks, since the boundary searcher expects that two consecutive chunks have a combined length greater than the boundary length.

NOTE: this implementation currently doesn't handle truncated multipart bodies.


seanmonstar:
So, I'm beyond excited at the work here! Thanks so much, this is super cool. I'd really like to fix warp's multipart filters to use full streaming. I have two meta thoughts about this though:

- Would this make sense as a separate crate, so that others could utilize it outside of warp?
- To have a full multipart parser, it'd be a good idea to have a sufficient test suite testing everything about it.

paolobarbolini:
> Would this make sense as a separate crate, so that others could utilize it outside of warp?

A separate crate could be interesting here. Looking around crates.io there are a few other implementations but they seem a bit bloated to me, so a new crate could change that. I could move this implementation to a separate crate and update the PR when I get around to publishing it.

> To have a full multipart parser, it'd be a good idea to have a sufficient test suite testing everything about it.

That's definitely something that's missing now and that I would add before calling this PR ready or publishing a crate. Another advantage of having this be a separate crate could be avoiding bloating warp's tests.

paolobarbolini:
I published the implementation as a separate crate and updated this PR to use it. Since I see you use the squash and merge GitHub feature I thought it would be fine to keep the commit history, so I just made a commit that removes the implementation from warp and wraps my crate.

Eragonfr:
Is there an update on this, I feel like this PR gets some importance as it allow to drop `multipart` dependencies. As it is no longer updated and now it also gives a new warning in rust nightly 
```
warning: the following packages contain code that will be rejected by a future version of Rust: buf_redux v0.8.4, multipart v0.18.0
```
The GitHub repo of multipart and buf_redux has been archived, as far as I know there is no current plan to update them.

willclevine:
Following up on @Eragonfr 's comment.  Is there any update on this?  There is now a vulnerability in `remove_dir_all` that will go away if `multipart` is removed:
```
error[vulnerability]: Race Condition Enabling Link Following and Time-of-check Time-of-use (TOCTOU)
...
    = Solution: Upgrade to >=0.8.0
    = remove_dir_all v0.5.3
      └── tempfile v3.3.0
          └── multipart v0.18.0
              └── warp v0.3.3

jaysonsantos:
Hey folks, what do you think about using [multer](https://github.com/rousan/multer-rs) for multipart? It seems more up-to date and maintained than the other options.

seanmonstar:
I tried fixing up the conflicts, and it just got annoying. I then tried to cherry-pick manually, which made it worse. So now, I have a new PR where I just copy and pasted the changes into the files, and made it a new commit. That's in #1031.

(Using multer would be fine, I tried for a few minutes, until I realized it only has `async fn`s, which means I'd need to use `stream::unfold` and a `BoxStream`. I'd merge a new PR doing that, if someone wanted to do it.

PR #848, opened 2021-05-10
Upgrade tokio-tungstenite to 0.14

coolreader18:
I also noticed there was some unnecessary indirection/boxing in the `FilterFn` implementation, so I removed that.


jxs (approved):

jxs:
nice catch, thanks!

PR #851, opened 2021-05-17
Host in trace info (fixed #847)

### LINKED ISSUE #847: warp::log::custom should use warp::filters::host::optional()
Right now, `info.host()` just returns the Host header, which does not work for HTTP/2. It should use the same logic as the host filter.

jakajancar:


jakajancar:
@seanmonstar @jxs CR please

jxs (approved):
LGTM!

src/filters/log.rs
```diff
@@ -156,11 +156,8 @@ impl<'a> Info<'a> {
     }
 
     /// View the host of the request
-    pub fn host(&self) -> Option<&str> {
-        self.route
-            .headers()
-            .get(header::HOST)
-            .and_then(|v| v.to_str().ok())
+    pub fn host(&self) -> Option<Authority> {
```
seanmonstar, line 159: So as to not be a breaking change, can this keep returning `&str`?

src/filters/trace.rs
```diff
@@ -46,12 +46,17 @@ pub fn request() -> Trace<impl Fn(Info) -> Span + Clone> {
             "request",
             remote.addr = Empty,
             method = %info.method(),
+            host = Empty,
```
seanmonstar, line 49: Is it generally useful to always include the host? I'd imagine for most servers, the actual host string isn't interesting, and someone would only want to see it if they served multiple hosts on the same server.

seanmonstar (changes requested):

src/filters/trace.rs
jakajancar: That's true, on the other hand for me host is, but I really don't care about version and proto, for example.

I would include it since in multi-host setups path is not interpretable without it, but if we don't want to, not a big deal either, as the `Fn(Info) -> Span` function can be defined in own code. Let me know.

src/filters/log.rs
jakajancar: I think it can. However then:

  - It's inconsistent with `warp::host()`
  - As a user, it's nice that you're made aware that it can also contain port (high potential of biting you)
  - As a user, it's nice that you don't have to parse into `Authority` yourself (esp. since it already was `Authority` internally)

I would just bite the bullet if we can and go for the cleaner API. Otherwise I can change it, yes.

jakajancar:
@seanmonstar This is still relevant to me, want to make some decisions on the open issues?

PR #853, opened 2021-05-18
Add 2h keepalive by default 

jq-rs:
Add 2h keepalive by default to drop unused TCP sessions (e.g. behind a firewall).

Fix unused deprecation error given by new compilers.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -362,7 +362,6 @@ impl fmt::Debug for Reason {
 }
 
 #[doc(hidden)]
-#[deprecated(note = "Use warp::reply::json and warp::reply::with_status instead.")]
 impl serde::Serialize for Rejection {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
--- a/src/server.rs
+++ b/src/server.rs
@@ -3,6 +3,7 @@ use std::net::SocketAddr;
 #[cfg(feature = "tls")]
 use std::path::Path;
 use std::sync::Arc;
+use std::time::Duration;
 
 use futures::{Async, Future, Poll, Stream};
 use hyper::server::conn::AddrIncoming;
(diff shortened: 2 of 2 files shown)


jq-rs:
Fixes #854. New 0.1.x release is kindly requested if this PR is approved.

jxs:
Hi, and thanks for the PR. I wonder if we should also allow for the user to configure it via `Server`. CC @seanmonstar

src/reject.rs
```diff
@@ -362,7 +362,6 @@ impl fmt::Debug for Reason {
 }
 
 #[doc(hidden)]
-#[deprecated(note = "Use warp::reply::json and warp::reply::with_status instead.")]
```
jxs, line 365: you refer that new compiler gives a deprecation warning but I was not able to reproduce with `1.52` nor `nightly`

jq-rs: Thanks for taking a look! This is what I see with deprecated part:
```
   Compiling warp v0.1.23 (/home/user/www/warp)
error: this `#[deprecated]` annotation has no effect
   --> /home/ubuntu/www/warp/src/reject.rs:365:1
    |
365 | #[deprecated(note = "Use warp::reply::json and warp::reply::with_status instead.")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: remove the unnecessary deprecation attribute
    |
    = note: `#[deny(useless_deprecated)]` on by default

error: could not compile `warp`
```
```
cargo --version
cargo 1.52.0 (69767412a 2021-04-21)
```

jq-rs: This starts to fail with cargo 1.49.0 (d00d64df9 2020-12-05), 1.48 still works while compiling [mles-webproxy](https://crates.io/crates/mles-webproxy)-crate.

jq-rs:
If the 2 h default is not acceptable, I can certainly implement an API implementation, please do suggest what kind of API would be preferable. To me, this feature itself is a must-have for public internet web servers.

jxs:
if you are on a rush, you can convert your [warp filters to `Service`](https://docs.rs/warp/0.3.1/warp/fn.service.html#example) and set the keepalive on hyper

jq-rs:
Configurable solution available in #1075, closing this.

PR #855, opened 2021-05-18
Impl Reply for Cow<'static, [u8]>

exrook:


jxs:
Hi, and thanks! 
Can you describe the use case for the need of this type?

serprex:
I was needing this, ended up writing all my code to use `Bytes` instead. Some responses were cached, meanwhile others computed

PR #856, opened 2021-05-24
docs(tests): Add missing await to Future

### LINKED ISSUE #793: Testing code in docs does not compile
Hi,
(Similar to issue #611) I am testing some filters using trying to use the examples from the docs but running into a compiler error:

```
use warp::Filter;

fn sum() -> impl Filter<Extract = (u32,), Error = warp::Rejection> + Copy {
    warp::path::param()
        .and(warp::path::param())
        .map(|x: u32, y: u32| x + y)
}

fn math() -> impl Filter<Extract = (String,), Error = warp::Rejection> + Copy {
    warp::post().and(sum()).map(|z: u32| format!("Sum = {}", z))
}

#[cfg(test)]
mod tests {
    use super::*;

    #[tokio::test]
    async fn test_sum() {
        let filter = sum();

        // Execute `sum` and get the `Extract` back.
        let value = warp::test::request()
            .path("/1/2")
            .filter(&filter)
            .await
            .unwrap();
        assert_eq!(value, 3);

        // Or simply test if a request matches (doesn't reject).
        assert!(!warp::test::request().path("/1/-5").matches(&filter));
    }
}
```

```
error[E0600]: cannot apply unary operator `!` to type `impl warp::Future`
  --> src/main.rs:30:17
   |
30 |         assert!(!warp::test::request().path("/1/-5").matches(&filter));
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot apply unary operator `!`
```

All that is missing here is to call .await on the Future within the assert! macro. I can submit a PR for this.

FlorianDr:
Fixes #793

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -51,6 +51,7 @@
 //!         !warp::test::request()
 //!             .path("/1/-5")
 //!             .matches(&filter)
+//!             .await
 //!     );
 //! }
 //! ```


src/test.rs
```diff
@@ -35,7 +35,6 @@
 //! # use warp::Filter;
 //! #[tokio::test]
 //! async fn test_sum() {
-//! #    let sum = || warp::any().map(|| 3);
```
seanmonstar, line 38: Isn't removing this going to cause the next line to fail?

FlorianDr: hey @seanmonstar, thank you for your comment. I'm new to Rust. Would you help me understand the `#`-prefixed lines in the example code? In the meantime, I reverted the change.

seanmonstar: This is something specific to Rust documentation. It will "hide" the line from the rendered documentation. It's useful to setting up things to make the compiler happy, but would other just be a distraction from the example.

So in this case, the rest of the code references a `sum` filter, but I didn't want to write out a full sum example here (users can go find it in the routing docs), so I just hid the "fake" one.

FlorianDr: Awesome! Thank you so much the for the explanation 🙏.. Pretty neat! Does the PR look good to you otherwise? Have a great Friday afternoon!

jxs (approved):

jxs:
thanks!

PR #858, opened 2021-05-29
Upgrade to multipart 0.18

adamchalmers:
The main advantage of its 0.18 release is that it uses rand 0.8, not 0.7. I was inspired to open this PR because my project (which uses warp) is pulling in both rand 0.8 (directly) and 0.7 (by way of warp, by way of multipart).

Multipart's MSRV is now 1.36, I couldn't see a MSRV for warp, but if the MSRV conflicts with warp's then please decline this PR :)


jxs (approved):

jxs:
Hi, and thanks!
warp tries to follow the same MSRV as hyper and tokio, so I'd say `rand` 1.36 is completely fine :)

PR #860, opened 2021-05-31
Remove unnecessary clone in `websockets_chat.rs`

ekzhang:
Hi, thanks for maintaining this great library. I think there is a line in the `websockets_chat.rs` library to clone an `Arc<_>` that isn't necessary, since the `users` variable is only passed around by a reference of type `&Arc<_>` anyway. I removed this cloned variable, and it still seems to compile and run fine.

Let me know if this makes sense.


jxs (approved):

jxs:
Hi, nice catch thanks! :rocket:

PR #861, opened 2021-06-01
Expose tracing/log feature

johannescpk:
[`log::Record`](https://docs.rs/log/0.4.14/log/struct.Record.html)'s are `!Send + !Sync` and `tracing`'s `log` feature automatically uses those, which can be problematic for consumers. Exposing the feature allows consumers to not enable the `log` feature.

The feature is called `trace-log` since both `log` and `tracing-log` are blocked by crates with the same name.


jxs (approved):

jxs:
Hi, yeah this makes sense, [hyper did the same](https://github.com/hyperium/hyper/issues/2326).
Thanks!

PR #865, opened 2021-06-10
Add a test with query string containing spaces

### LINKED ISSUE #823: Tests with spaces in query string
Using the test framework to simulate requests like:
```warp::test::request().path("/api?somekey=someparam")```
Works fine.

But if you insert a space like:
```warp::test::request().path("/api?somekey=some param")```
It crashes with "test request path invalid" at https://github.com/seanmonstar/warp/blob/b6d1fc0719604ef1010aec00544408e6af1289a5/src/test.rs#L200
Because it attempts to directly parse the given path as a URI. But spaces in URI should be percent encoded, so it is rejected.

And, if you do percent encode it:
```warp::test::request().path("/api?somekey=some param")```
It parses fine, but instead of sending a space to the request handler, it actually sends a ```%20```
So the handler performs incorrectly.

However, if you make a web request with that same path: ```"https://someurl.com/api?somekey=some param```, it parses fine, and sends the space to the handler.

So tests cannot be written to match web requests (with a space in query parameters).

devashishdxt:
Fixes #823.

### DIFF
--- a/tests/query.rs
+++ b/tests/query.rs
@@ -16,6 +16,17 @@ async fn query() {
 }
 
 #[tokio::test]
+async fn query_with_space() {
+    let as_map = warp::query::<HashMap<String, String>>();
+
+    let req = warp::test::request().path("/?foo=foo%20bar&baz=quux");
+
+    let extracted = req.filter(&as_map).await.unwrap();
+    assert_eq!(extracted["foo"], "foo bar");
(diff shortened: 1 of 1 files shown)


jxs:
Hi! and thanks for the submission! but see https://github.com/seanmonstar/warp/issues/823#issuecomment-858867751.

PR #866, opened 2021-06-18
docs: Note that most futures from Server run on current runtime

emesterhazy:
The documentation for most of the warp::Server methods (bind,
try_bind_with_graceful_shutdown, etc) indicate that the futures
they return can be run on any runtime. However, these functions
panic if they are called outside the context of a Tokio 1.x runtime.

This commit makes a minor update to the documentation to note that
these function return futures that run on the /current/ runtime.

Note that try_bind can actually be called outside the context of a
Tokio runtime, so the docstring for that method remains unchanged.

### DIFF
--- a/src/filters/path.rs
+++ b/src/filters/path.rs
@@ -308,7 +308,7 @@ pub fn tail() -> impl Filter<Extract = One<Tail>, Error = Infallible> + Copy {
     })
 }
 
-/// Represents that tail part of a request path, returned by the `tail()` filter.
+/// Represents the tail part of a request path, returned by the [`tail()`] filter.
 pub struct Tail {
     path: PathAndQuery,
     start_index: usize,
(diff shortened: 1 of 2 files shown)


emesterhazy:
I think there's also a typo in the documentation for `Peek` and `Tail`:


> Peek | Represents that tail part of a request path, returned by the&nbsp;tail()&nbsp;filter.

Shouldn't this say "returned by the `peek()` filter?

> Tail | Represents that tail part of a request path, returned by the&nbsp;tail()&nbsp;filter.

Both of these should probably say "Represents *the* tail part..." instead of *that*.

Thoughts? I can make those tweaks as well.

jxs:
Hi, and thanks!
>Thoughts? I can make those tweaks as well.

yeah, seems to me that `Peek` should be:
>Represents the tail part of a request path, returned by the peek() filter.

and Tail should be:
>Represents the tail part of a request path, returned by the tail() filter.

so if you want to go ahead and update these as well that would be great!
Also since you are on it wdyt of making `peek()` and `tail()` references links to the matching functions page? (you can see an example on how to do it [here](https://docs.rs/warp/0.3.1/src/warp/filters/path.rs.html#8))

emesterhazy:
Good idea on the links. I'm not sure when, but at some point the compiler got smarter and all that I needed to do is wrap the functions in brackets. While I was at it I turned the reference to `full()` into a link as well.

jxs (approved):

jxs:
awesome thanks!

PR #868, opened 2021-06-23
Expose `FilterBase` to cargo doc

xd009642:
So when trying to pass a filter as an argument to a function I was struggling to get the trait bounds to work generically going off the compiler error messages or published documentation. I asked on the discord and got the answer and looking at the code it is clear what the solution was but the documentation could be more helpful but this is kind of complicated by the `FilterBase` trait being private. @Kestrer suggested doing a PR to expose `FilterBase` just in docs so here it is. I guess the one concern is maybe the docs as they would be after this change may need some more big bold text to tell people `FilterBase` isn't available for them to use and they shouldn't rely on it's implementation details. Potentially `#[doc(hidden)]` on the trait methods so only the `Extract` and `Rejection` types are shown in the documentation?

The documentation before:

Before

The documentation after showing me the `Extract` value I needed to get my code to work

After

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -1,3 +1,4 @@
+#![allow(missing_docs)]
 mod and;
 mod and_then;
 mod boxed;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -103,6 +103,8 @@ mod transport;
 
 pub use self::error::Error;
 pub use self::filter::Filter;
+#[cfg(doc)]
+pub use crate::filter::FilterBase;
 // This otherwise shows a big dump of re-exports in the doc homepage,
 // with zero context, so just hide it from the docs. Doc examples
 // on each can show that a convenient import exists.


xd009642:
For some more motivation this is the minimal example I was trying to get working:

```rust
use std::net::SocketAddr;
use warp::{Filter, Rejection, Reply};

fn path_as_first_arg(
    f: impl Filter + Clone + Send + Sync,
) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone + Sync + Send {
    // In reality this is more involved setting up 3 subroutes based on
    // content-type and other headers
    f.and(warp::get()).map(|| String::from("Hello"))
}

#[tokio::main]
async fn main() {
    let routes =
        path_as_first_arg(warp::path!("foo" / "bar")).or(path_as_first_arg(warp::path!("baz")));

    warp::serve(routes)
        .run("0.0.0.0:1804".parse::<SocketAddr>().unwrap())
        .await;
}
```

The solution is to do ` f: impl Filter<Extract=(), Error=Rejection> + Clone + Send + Sync,` but the error messages weren't particularly helpful (at least to me):

```
error[E0277]: the trait bound `<<impl Filter + Clone + Send + Sync as warp::filter::FilterBase>::Extract as warp::generic::Tuple>::HList: warp::generic::Combine<_>` is not satisfied
 --> src/main.rs:9:7
  |
9 |     f.and(warp::get()).map(|| String::from("Hello"))
  |       ^^^ the trait `warp::generic::Combine<_>` is not implemented for `<<impl Filter + Clone + Send + Sync as warp::filter::FilterBase>::Extract as warp::generic::Tuple>::HList`

error[E0277]: the trait bound `Rejection: reject::sealed::CombineRejection<<impl Filter + Clone + Send + Sync as warp::filter::FilterBase>::Error>` is not satisfied
 --> src/main.rs:9:7
  |
9 |     f.and(warp::get()).map(|| String::from("Hello"))
  |       ^^^ the trait `reject::sealed::CombineRejection<<impl Filter + Clone + Send + Sync as warp::filter::FilterBase>::Error>` is not implemented for `Rejection`
  |
help: consider introducing a `where` bound, but there might be an alternative better way to express th
is requirement
  |
6 | ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone + Sync + Send where Rejection: reject::sealed::CombineRejection<<impl Filter + Clone + Send + Sync as warp::filter::FilterBase>::Error> {
  |                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: the method `map` exists for struct `warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>`, but its trait bounds were not satisfied
  --> src/main.rs:9:24
   |
9  |     f.and(warp::get()).map(|| String::from("Hello"))
   |                        ^^^ method cannot be called on `warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>` due to unsatisfied trait bounds
   | 
  ::: /home/daniel/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.3.1/src/filter/and.rs:13:1
   |
13 | pub struct And<T, U> {
   | --------------------
   | |
   | doesn't satisfy `_: Iterator`
   | doesn't satisfy `_: warp::Filter`
   | doesn't satisfy `_: warp::filter::FilterBase`
   |
   = note: the following trait bounds were not satisfied:
           `warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::filter::FilterBase`
           which is required by `warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::Filter`
           `&warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::filter::FilterBase`
           which is required by `&warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::Filter`
           `&mut warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::filter::FilterBase`
           which is required by `&mut warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: warp::Filter`
           `warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: Iterator`
           which is required by `&mut warp::filter::and::And<impl Filter + Clone + Send + Sync, impl warp::Filter+Copy>: Iterator`

error: aborting due to 3 previous errors
```

jxs:
Hi, and thanks for your interest!  
Yeah unfortunately the Rust compiler isn't helpful in a lot of situations regarding warp, but this PR basically opens up `FilterBase` for implementation, afaik [`#[cfg(doc)]` is used If you want to make sure an item is seen by Rustdoc regardless of what platform it's targeting](https://doc.rust-lang.org/rustdoc/advanced-features.html#cfgdoc-documenting-platform-specific-or-feature-specific-information), so here it's redundant.  
That's why you also had to add `#![allow(missing_docs)]` for the compilation to not break as it becoming public needs to be documented

xd009642:
So I thought `#[cfg(doc)]` would only make it public for `cargo doc` going from this comment on the discussion on discord: https://discordapp.com/channels/500028886025895936/671042482770935808/857296970829922314 that kind of lead me to the assumption the `allow(missing_docs)` might use the doc coverage stuff in cargo doc to work out if things weren't covered by doc tests. This is the fiddly part of sealed traits I'm really not sure on without looking things up or trying them out and I'm not on a PC with rust installed right now but I'll look at it tomorrow and see if I can figure a way around it

PR #872, opened 2021-06-27
rework fs filter byte-ranges, fixes #867

### LINKED ISSUE #867: Warp returns incorrect results for 'Range: bytes=-something'
**Version**
warp v0.3.1

**Platform**

`Linux branna 5.8.0-53-generic #60-Ubuntu SMP Thu May 6 07:46:32 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

**Description**

A `Range:` header like `bytes=-10` is supposed to return the last ten bytes of the file. But `warp::fs` instead returns the first ten bytes of the file.

Simple reproducer:

```rust
#[tokio::main]
async fn main() {
    warp::serve(warp::fs::file("/usr/share/dict/words"))
        .run(([127, 0, 0, 1], 3030))
        .await;
}
```

What I got: a response with `content-range: 0-10` and the first 10 bytes of the file:

```
❯ curl -iH 'Range: bytes=-10' http://localhost:3030
HTTP/1.1 206 Partial Content
content-range: bytes 0-10/976241
content-length: 11
content-type: application/octet-stream
accept-ranges: bytes
last-modified: Sun, 01 Mar 2020 04:07:46 GMT
date: Sun, 20 Jun 2021 02:59:07 GMT

A
A's
AMD
A%                                                                                        
```

What I expected: a response with `content-range: 976231-976240/976241` and the last 10 bytes of the file.

Spec reference:

>    A client can request the last N bytes of the selected representation
>    using a suffix-byte-range-spec.
> 
>      suffix-byte-range-spec = "-" suffix-length
>      suffix-length = 1*DIGIT
> 
>    If the selected representation is shorter than the specified
>    suffix-length, the entire representation is used.
> 
>    Additional examples, assuming a representation of length 10000:
> 
>    o  The final 500 bytes (byte offsets 9500-9999, inclusive):
> 
>         bytes=-500
> 
- https://datatracker.ietf.org/doc/html/rfc7233#section-2.1

jxs:


src/filters/fs.rs
```diff
@@ -369,23 +369,21 @@ fn bytes_range(range: Option<Range>, max_len: u64) -> Result<(u64, u64), BadRang
...
+                }
+                (Bound::Included(b), Bound::Unbounded) => (b, max_len),
+                (Bound::Unbounded, Bound::Included(e)) => {
+                    if e > max_len {
+                        return Err(BadRange);
                     }
+                    (max_len - e, max_len)
                 }
-                Bound::Excluded(s) => s,
+                _ => unreachable!(),
```
jxs, line 386: `Range::iter()`  [only returns `Bounds::included` or `Bounds::Unbounded`](https://github.com/hyperium/headers/blob/master/src/common/range.rs#L71-L84)

src/filters/fs.rs
```diff
@@ -369,23 +369,21 @@ fn bytes_range(range: Option<Range>, max_len: u64) -> Result<(u64, u64), BadRang
...
-                Bound::Excluded(s) => s + 1,
-            };
-
-            let end = match end {
-                Bound::Unbounded => max_len,
-                Bound::Included(s) => {
+            let (start, end) = match (start, end) {
+                (Bound::Unbounded, Bound::Unbounded) => (0, max_len),
+                (Bound::Included(b), Bound::Included(e)) => {
                     // For the special case where s == the file size
```
joseluisq, line 375: I guess the comment means "e instead of s", right?
But why not use `a` and `b` var names for the destructured tuples?

src/filters/fs.rs
joseluisq: Rust says that `unreachable!()` always panic so should we avoid it with something like `return Err(BadRange);`? what do you think?

joseluisq: I got it! I think my previous comment is irrelevant since all cases are covered and `unreachable!()` just clarify the last case for compiler.

src/filters/fs.rs
jxs: yeah, should be `s` instead of `b` thanks! `s` and `e` match the `start` and `end` variables used on the `let` above

Issue #77, opened 2018-08-28
Rejecting with warp::reply::not_found results in a 405 Method Not Allowed error


Geal:
Hello,
with the following code, I would get a 405 error while I expect a 404:

```rust
fn main() {
  let routes = warp::get2().and(warp::path("test").and_then(send_test));
  warp::serve(routes).run(([127, 0, 0, 1], 3000));
}

fn send_test() -> Result<String, warp::Rejection> {
  Err(warp::reject::not_found())
}
```

Is there something I am doing wrong?

seanmonstar:
No, you aren't wrong in thinking that. Currently, the rejections in warp collect and the highest priority one is chosen to render as a response. It's assumed that Not Found is the lowest priority, so the method rejection from the post plus the not found rejection you create means the 405 is shown.

The priority of these could maybe need adjustment.

Geal:
maybe there could be a way to cut backtracking? ie, if we know we got to a specific prefix or argument, we're in the right route, and should return the error from this route

algermissen:
Can you explain how I could return such a response from a header filter? I think that is not possible right now if I am not mistaken?

seanmonstar:
> Can you explain how I could return such a response from a header filter?

You could make use of `Filter::recover` to alter the rejection, or `Filter::or_else` can be useful to provide a "default" value overriding any rejection.

mexus:
My vote is to a way to cut the backtracking. Currently I'm using the approach proposed by @blm768 , but as he mentioned it's not quite ergonomic.
IMHO any reject provided by a user's code should by default stop the processing and just return the rejection, and probably it's not a too bad idea to have a special reject like `reject::try_something_different()` (bad naming, but I hope you'll get the point) that should be returned by the combinators to go for furthers `or`s.

kadfak:
There's an example to customize errors as described by @blm768 for anyone interested: [examples/errors.rs](https://github.com/seanmonstar/warp/blob/1b51ce9f6a507a858c146562ce8f373ed1ec1d7d/examples/errors.rs)

algermissen:
What do other's think about the need for context-specific rejection payload? I still see no way to do that with the rejection design as-is. Or I am missing something.

My use case is 401 Unauthorized, where it is usually only known in that specific route, what exactly to return as Challenge parameters. Another use case would be HTTP Problems (https://tools.ietf.org/html/rfc7807). 

Thoughts?

boustrophedon:
[...]

PR #874, opened 2021-06-29
Add support for custom HTTP methods

### LINKED ISSUE #871: Support for HTTP extension methods
I'm trying to implement WebDAV handlers using warp, which requires filters for non-standard HTTP methods, as documented in RFC 4918. While `http::Method` allows defining custom methods, warp only provides a pre-defined set of filters for the standard methods and keeps `method_is` in `filters::method` private, so I couldn't figure out any other way to filter for custom methods.
  
Making `method_is` public would probably be enough already. warp could provide some sugar for `Method::from_bytes`, but that is not absolutely necessary.  

I could't figure out any other way to filter for the method in warp.

**Additional context**
N/A

e-user:
Fixes #871.

I had to turn all method filters from `Copy` to `Clone` because custom `Method`s cannot be created as `const`, which makes passing them as `&'static` impossible in practice (or at least, I couldn't figure out a way to do it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v0.x.x (TBD)
+- **Features**:
+  - Add `custom_method` filter.
+
 ### v0.3.1 (March 24, 2021)
 
 - **Features**:
--- a/src/filters/method.rs
+++ b/src/filters/method.rs
@@ -12,6 +12,7 @@ use http::Method;
 use crate::filter::{filter_fn, filter_fn_one, Filter, One};
 use crate::reject::Rejection;
 use std::convert::Infallible;
+use std::str::FromStr;
 
 /// Create a `Filter` that requires the request method to be `GET`.
 ///
@@ -22,8 +23,8 @@ use std::convert::Infallible;
 ///
 /// let get_only = warp::get().map(warp::reply);
 /// ```
-pub fn get() -> impl Filter<Extract = (), Error = Rejection> + Copy {
-    method_is(|| &Method::GET)
+pub fn get() -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    method_is(|| Method::GET.clone())
 }
 
 /// Create a `Filter` that requires the request method to be `POST`.
@@ -35,8 +36,8 @@ pub fn get() -> impl Filter<Extract = (), Error = Rejection> + Copy {
 ///
 /// let post_only = warp::post().map(warp::reply);
 /// ```
-pub fn post() -> impl Filter<Extract = (), Error = Rejection> + Copy {
-    method_is(|| &Method::POST)
+pub fn post() -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    method_is(|| Method::POST.clone())
 }
 
 /// Create a `Filter` that requires the request method to be `PUT`.
@@ -48,8 +49,8 @@ pub fn post() -> impl Filter<Extract = (), Error = Rejection> + Copy {
 ///
 /// let put_only = warp::put().map(warp::reply);
 /// ```
-pub fn put() -> impl Filter<Extract = (), Error = Rejection> + Copy {
-    method_is(|| &Method::PUT)
+pub fn put() -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    method_is(|| Method::PUT.clone())
 }
 
 /// Create a `Filter` that requires the request method to be `DELETE`.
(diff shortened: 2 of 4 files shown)


jxs:
Hi, thanks for your interest!
Hm I am torn on this one, the way you implemented on this PR makes it a breaking change (existing methods stop being `Copy`) we could make it so for `custom_method` not using `method_is` and repeating it's code, therefore removing the `Copy` constraints but then I think it makes it more convoluted than the way you implemented it on https://github.com/seanmonstar/warp/issues/871#issuecomment-869131464 which I don't find it ugly tbh.

e-user:
Hi @jxs, thanks for coming back to me on this one!

I can see the problem. About my workaround, I figured out it's not actually working; once I use a custom error to signal method rejection and turn it into an `Ok(StatusCode::METHOD_NOT_ALLOWED)` in my `recover` handler, the whole top-level chain fails because it's not a `Rejection`, anymore. In other words, the next method-based filter chain is not considered, anymore. One change to make it work could be instead to open up the other `Rejection` methods besides `not_found` and `custom`, so I could use `method_not_allowed`, instead.

So either I would change the PR to implement `custom_method` as you suggested, or close this one in favor of a new one that opens up the `Rejection` methods to the public. Which one do you think is preferable?

jxs:
>I can see the problem. About my workaround, I figured out it's not actually working; once I use a custom error to signal method rejection and turn it into an Ok(StatusCode::METHOD_NOT_ALLOWED) in my recover handler, the whole top-level chain fails because it's not a Rejection, anymore. In other words, the next method-based filter chain is not considered, anymore. One change to make it work could be instead to open up the other Rejection methods besides not_found and custom, so I could use method_not_allowed, instead.

but that's expected, after returning `Ok()` from a `recover` handler you can then apply new filters on the extracted value. The behavior will be the same if the rejection was known and handled. If you want to continue the filter chain you can use `or_else` when your filter rejects based on the method, and then plugging `recover` in the end of the filter chain

e-user:
> but that's expected, after returning Ok() from a recover handler you can then apply new filters on the extracted value. The behavior will be the same if the rejection was known and handled. If you want to continue the filter chain you can use or_else when your filter rejects based on the method, and then plugging recover in the end of the filter chain

To be honest, I spend quite some time trying to produce a minimal version of my problem and ultimately didn't even understand why my original version with `custom_method` from this PR works as intended. I ultimately realized my actual problem is stemming from the fact that I need rejections to be treated differently depending on where they appear, which is only possible with two distinct recovery handlers.

Maybe you can tell me whether I'm making things more complex than they ought to be, but for the record and if anyone should come across the same problem, here's the gist of it:

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    env_logger::init_from_env(Env::default().default_filter_or("info"));
    let address: SocketAddr = "[::]:9090".parse()?;

    let foo_route = get().and(warp::path!("foo")).map(|| "Success");
    let bar_route = post().and(warp::path!("bar")).map(|| "Success");

    warp::serve(foo_route.or(bar_route)).run(address).await;

    Ok(())
}
```

For `GET localhost:9090/baz`, this will yield `405 Method Not Allowed` instead of the expected / desired `404 Not Found`, because the `or(bar_route)` "wins" and yields its own 405 over the original 404. So I had to attach a `recover` handler after each route in my original implementation that turns the 404 `Rejection` into an `Ok` with `StatusCode::NOT_FOUND`, i.e.

```rust
if reject.is_not_found() {
    Ok(StatusCode::NOT_FOUND)
} 
```

(same goes for my custom errors that need to be turned from `Rejection`s to corresponding `Ok`s)

At this point, since I'm using a custom `Rejection` for custom HTTP methods, I had to append another final `recover` handler at the end of the filter chains to make sure a `405 Method not allowed` gets yielded instead of a `500`:

```rust
pub async fn handle_rejection(reject: Rejection) -> Result<impl Reply, Rejection> {
    if let Some(e) = reject.find::<GatewayHandlerError>() {
        match e {
            GatewayHandlerError::MethodNotAllowed => Ok(StatusCode::METHOD_NOT_ALLOWED),
            _ => Err(reject),
        }
    } else {
        log::info!("keeping rejection {:?}", reject);
        Err(reject)
    }
}
```

I'm therefore closing this PR, but will consider makind another one to add this as an example and/or documentation on how to use custom methods with warp. @jxs please let me know if I'm missing anything or this can be done any easier than above.

jxs:
>For GET localhost:9090/baz, this will yield 405 Method Not Allowed instead of the expected / desired 404 Not Found, because the or(bar_route) "wins" and yields its own 405 over the original 404. So I had to attach a recover handler after each route in my original implementation that turns the 404 Rejection into an Ok with StatusCode::NOT_FOUND, i.e.

that's because of the order you have the filters:

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let address: SocketAddr = "[::]:9090".parse()?;

    let foo_route = warp::path!("foo").and(warp::get()).map(|| "Success");
    let bar_route = warp::path!("bar").and(warp::post()).map(|| "Success");

    warp::serve(foo_route.or(bar_route)).run(address).await;

    Ok(())
}
```
and `GET localhost:9090/baz` returns the expected 404,  see https://github.com/seanmonstar/warp/issues/77 for more info

PR #875, opened 2021-06-29
Add `route` filter to get insights into the current request.

### LINKED ISSUE #812: Make route::with public for external
I want to redirect all calls for a specific path from http to https. For this I cannot read the full uri to identify the the scheme and authority.

One solution is to add warp::with_uri method. another would be to make public route::with with the route class. this would be possible for different integrations into warp. Keeping route private from external crates make it hard to glued the missing gaps in warp.

No alternatives.

**Additional context**
NA

e-user:
Fixes #812.

While `trace` and `log` provide access to a limited view on `Route`, they both don't allow the data to leave the passed in closures, so it's unusable for creating filters based on the contained information. Access to the original URL, HTTP version and other data is not just useful, but required for several use cases. Some of the data in `Info` is duplicate if existing `header` filters are considered and might be removed, if desired. For now, I copied what's in `trace`'s `Info` and added `uri`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### v0.x.x (TBD)
+- **Features**:
+  - Add `route` filter.
+
 ### v0.3.1 (March 24, 2021)
 
 - **Features**:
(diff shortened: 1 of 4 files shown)


e-user:
I figured out that everything but the HTTP version can be obtained from other places scattered throughout the API. Since a much more simple filter could be added for the HTTP version, I'm closing this PR since it doesn't really add much.

Issue #126, opened 2018-10-29
Docs touch-up for Rejections


seanmonstar:
With the changes to the Rejection system (#115), the docs could use some improvements.

- The docs for `warp::reject::custom` should link to `Filter::recover`.
- The docs for `warp::Rejection` should link to the `warp::reject` module documentation.
- The `warp::reject` module docs should expand on how to recover from rejections.

boxdot:
Looks like this was already done in the commit b262f934bf1c263ea1c42219f2b8e3121715748d.

jxs:
hum, after docs for `warp::reject::custom` link to `Filter::recover`,
and docs for `warp::Rejection` link to `warp::reject` module documentation, 
but `warp::reject` module docs doesn't yet refer to `Filter::recover` or the errors example, should this be addressed @seanmonstar ?

calvinbrown085:
I can grab 
> The warp::reject module docs should expand on how to recover from rejections.

calvinbrown085:
Is there a specific example on what kind of rejection you'd want to recover from?

jxs:
fixed with #876

PR #876, opened 2021-06-29
expanded documentation on how to handle rejections

aujxn:
I saw an easy open issue #126 to expand the docs about handling rejections. I'm new to the library but would like to get involved with the project.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -11,20 +11,56 @@
 //! new custom [`Filter`](../trait.Filter.html)s and still want other routes to be
 //! matchable in the case a predicate doesn't hold.
 //!
+//! As a request is processed by a Filter chain, the rejections are accumulated into
+//! a list contained by the [`Rejection`](struct.Rejection.html) type. Rejections from
+//! filters can be handled using [`Filter::recover`](../trait.Filter.html#method.recover).
+//! This is a convenient way to map rejections into a [`Reply`](../reply/trait.Reply.html).
+//!
+//! For a more complete example see the
+//! [Rejection Example](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs)
+//! from the repository.
+//!
 //! # Example
 //!
 //! ```
-//! use warp::Filter;
+//! use warp::{Reply, Filter, reject, Rejection, http::StatusCode};
 //!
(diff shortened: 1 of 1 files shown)


src/reject.rs
```diff
@@ -11,20 +11,56 @@
...
+//!
+//! // Custom rejection handler that maps rejections into responses.
+//! async fn handle_rejection(err: Rejection) -> Result<impl Reply, std::convert::Infallible> {
+//!     if err.is_not_found() {
+//!         Ok(reply::with_status("NOT_FOUND", StatusCode::NOT_FOUND))
+//!     } else if let Some(e) = err.find::<InvalidParameter>() {
+//!         Ok(reply::with_status("BAD_REQUEST", StatusCode::BAD_REQUEST))
+//!     } else {
+//!         // Could use eprintln! if not using logging crate
+//!         error!("unhandled rejection: {:?}", err);
```
jxs, line 42: we can just `eprintln!` here, and remove the comment

src/reject.rs
```diff
@@ -11,20 +11,56 @@
...
+//! async fn main() {
+//!
+//!     // Filter on `/:id`, but reject with InvalidParameter if the `id` is `0`.
+//!     // Recover from this rejection using a custom rejection handler.
+//!     let route = warp::path::param()
+//!         .and_then(|id: u32| async move {
+//!             if id == 0 {
+//!                 Err(warp::reject::custom(InvalidParameter))
+//!             } else {
+//!                 Ok("something since id is valid")
```
jxs, line 57: we can just have "id is valid!" here

jxs (changes requested):
Hi, and thanks for getting this started! :+1:  Only missing fixing the issues reported by CI and a couple of notes I left

src/reject.rs
aujxn: Yeah, this makes more sense thanks for the help

src/reject.rs
aujxn: Perfect

aujxn:
Awesome! Thanks for the help, `cargo test` succeeds now I should have tried that before submitting the PR.

jxs (approved):

jxs:
awesome, thanks! :rocket:

PR #877, opened 2021-07-01
Add example how to use custom HTTP methods

### LINKED ISSUE #871: Support for HTTP extension methods

e-user:
Fixes #871.

This has been created as a result out of #874.

### DIFF
--- a/examples/custom_methods.rs
+++ b/examples/custom_methods.rs
@@ -1,6 +1,5 @@
 #![deny(warnings)]
 use std::net::SocketAddr;
-use std::str::FromStr;
 
 use warp::hyper::StatusCode;
 use warp::{hyper::Method, reject, Filter, Rejection, Reply};
--- a/examples/custom_methods.rs
+++ b/examples/custom_methods.rs
@@ -14,13 +14,11 @@ const BAR_METHOD: &'static str = "BAR";
 
 fn method(name: &'static str) -> impl Filter<Extract = (), Error = Rejection> + Clone {
     warp::method()
-        .and_then(move |m: Method| {
-            async move {
-                if m == name {
-                    Ok(())
-                } else {
-                    Err(reject::custom(MethodError))
-                }
+        .and_then(move |m: Method| async move {
+            if m == name {
+                Ok(())
+            } else {
+                Err(reject::custom(MethodError))
             }
         })
         .untuple_one()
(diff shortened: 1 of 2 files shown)


examples/custom_methods.rs
```diff
@@ -0,0 +1,65 @@
...
+use std::str::FromStr;
+
+use warp::hyper::StatusCode;
+use warp::{hyper::Method, reject, Filter, Rejection, Reply};
+
+#[derive(Debug)]
+struct MethodError;
+impl reject::Reject for MethodError {}
+
+fn method(name: &str) -> impl Filter<Extract = (), Error = Rejection> + Clone {
```
seanmonstar, line 12: If we assume most methods are going to be known at compile time, we could make this a `&'static str`, and then you don't need to parse it as a `Method`, or clone it, and can just compare `if m == name`, and be done with it.

e-user: Really good point. Addressed in the new commit.

examples/custom_methods.rs
```diff
@@ -0,0 +1,64 @@
...
+struct MethodError;
+impl reject::Reject for MethodError {}
+
+const FOO_METHOD: &'static str = "FOO";
+const BAR_METHOD: &'static str = "BAR";
+
+fn method(name: &'static str) -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    warp::method()
+        .and_then(move |m: Method| {
+            async move {
```
seanmonstar, line 18: Looks like rustfmt is suggesting to move the `async move {` up to be on the same line as the start of the closure.

e-user: Interesting, it seems like a missing `Cargo.toml` made my rustfmt fail in applying the correct formatting. Fixed by running it with the corresponding option manually.

e-user:
@seanmonstar Would you mind starting another workflow, please?

Issue #622, opened 2020-06-24
[Question] Interrupting an `or` filter chain


Geobert:
Hi there!

I have this default filter:
```
let get_default = warp::any()
        .map(warp::reply)
        .map(|reply| warp::reply::with_status(reply, StatusCode::NOT_FOUND));
```

so any non expected request returns 404 instead of 500 (especially favicon.ico requests).

It's set up like:

```
let get_routes = warp::get().and(get_filter1.or(get_filter2).or(get_assets).or(get_default));
```

but then, if `get_filter1` reply an error (401 in my case), warp jumps to `get_default` and the client receive 404 instead of my 401.

How to interrupt the or chain?

Or how to define get_default differently to do that?

seanmonstar:
It looks like your "default" filter could instead be written using `Filter::recover`, so you can inspect what rejections occurred and send a custom response.

Geobert:
> 
> 
> It looks like your "default" filter could instead be written using `Filter::recover`, so you can inspect what rejections occurred and send a custom response.

Like always, thanks for your answer!

I'll try that approach and post my solution here for posterity :)

Geobert:
I already have a `recover`:

```rust
let routes = get_routes.or(post_routes).recover(customize_error);
```

and I noticed that any undefined request return `MethodNotAllowed` rejection so I added in `customize_error`:
```rust
// other error customization code
} else if let Some(_) = err.find::<warp::reject::MethodNotAllowed>() {
        code = StatusCode::NOT_FOUND;
        message = "NOT_FOUND".to_string();
    } else {
        code = StatusCode::INTERNAL_SERVER_ERROR;
        message = "UNHANDLED_REJECTION".to_string();
}
```

songday:
Hi, I encountered this as well.
I defined many REST endpoints, `5 GET methods and 2 POST methods`

once I request one of POST methods, I found `Rejection` in `recover` has
```
&err = Rejection(
    [
        MethodNotAllowed,
        MethodNotAllowed,
        MethodNotAllowed,
        MethodNotAllowed,
        MethodNotAllowed,
    ],
)
```

these 5 `MethodNotAllowed` came from those 5 GET methods
this cause `if let Some(_) = err.find::<warp::reject::MethodNotAllowed>()` will be matched and I can't get `the real error`

why did this happen?

PR #878, opened 2021-07-05
Implemented `Filter::then`

### LINKED ISSUE #594: Difference between and_then and map
I'm having a little difficulty understanding the difference between the `map` and `and_then` filters.

For `Result` the difference seems to be that `and_then` replaces the entire `Result` with the function return value (if the original value is Ok) whereas `map` only replaces the Ok value. In other words you `map(foo)` is the same as `and_then(Ok(foo))`.

But in Warp they seem to be entirely different things because `and_then` is async and `map` isn't. Can you explain this? And is there a simple sync version of `and_then` that I can use instead of writing an async block without any async code?

### LINKED ISSUE #712: Improved rejections

gtsiam:
Tipping point for me to jump into `Future` type hell to implement this was [this](https://github.com/seanmonstar/warp/issues/712#issuecomment-697031645) comment - but I'd say it's worth it. I've used it in my personal project for a while, and the code I've been able to write is much cleaner, I think.

Changes:
- Added `Filter::then()` and complementary type `Then`. Exactly like `map`, but async. The name is meant to reflect [future's `then()`](https://docs.rs/futures/0.3.17/futures/future/trait.FutureExt.html#method.then) function.
- `Filter::and_then()` documentation now points to `Filter::then()` as an alternative, since my understanding is that:
   > Rejections are meant to say "this filter didn't accept the request, maybe another can." 

Fix #712
Fix #594

Providing Reply implementations for Result would also be very useful, so this is loosely related to #458

---

Post-edit note: This function was renamed from `map_async` to `then` to more closely match `future`'s api. However it is worth noting that just as `Filter::map` isn't exactly like `future`'s `map` on account of rejections, `Filter::then` isn't exactly like `future`'s `then`. However `Filter::then` is the same, conceptually, as `future`'s `then`, so this seems like the best choice.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -7,6 +7,7 @@ mod or;
 mod or_else;
 mod recover;
 pub(crate) mod service;
+mod then;
 mod unify;
 mod untuple_one;
 mod wrap;
@@ -27,6 +28,7 @@ pub(crate) use self::map_err::MapErr;
 pub(crate) use self::or::Or;
 use self::or_else::OrElse;
 use self::recover::Recover;
+use self::then::Then;
 use self::unify::Unify;
 use self::untuple_one::UntupleOne;
 pub use self::wrap::wrap_fn;
@@ -197,12 +199,44 @@ pub trait Filter: FilterBase {
         }
     }
 
-    /// Composes this `Filter` with a function receiving the extracted value.
+    /// Composes this `Filter` with an async function receiving
+    /// the extracted value.
+    ///
+    /// The function should return some `Future` type.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use warp::Filter;
+    ///
+    /// // Map `/:id`
+    /// warp::path::param().then(|id: u64| async move {
+    ///   format!("Hello #{}", id)
+    /// });
+    /// ```
+    fn then<F>(self, fun: F) -> Then<Self, F>
+    where
+        Self: Sized,
+        F: Func<Self::Extract> + Clone,
+        F::Output: Future + Send,
+    {
+        Then {
+            filter: self,
+            callback: fun,
+        }
+    }
+
+    /// Composes this `Filter` with a fallible async function receiving
+    /// the extracted value.
     ///
     /// The function should return some `TryFuture` type.
     ///
     /// The `Error` type of the return `Future` needs be a `Rejection`, which
     /// means most futures will need to have their error mapped into one.
+    ///
+    /// Rejections are meant to say "this filter didn't accept the request,
+    /// maybe another can". So for application-level errors, consider using
+    /// [`Filter::then`] instead.
     ///
     /// # Example
     ///
--- a/src/filter/then.rs
+++ b/src/filter/then.rs
@@ -6,27 +6,25 @@ use futures::{ready, TryFuture};
 use pin_project::pin_project;
 
 use super::{Filter, FilterBase, Func, Internal};
-use crate::reject::CombineRejection;
 
 #[derive(Clone, Copy, Debug)]
-pub struct AndThen<T, F> {
+pub struct Then<T, F> {
     pub(super) filter: T,
     pub(super) callback: F,
 }
 
-impl<T, F> FilterBase for AndThen<T, F>
+impl<T, F> FilterBase for Then<T, F>
 where
     T: Filter,
     F: Func<T::Extract> + Clone + Send,
-    F::Output: TryFuture + Send,
-    <F::Output as TryFuture>::Error: CombineRejection<T::Error>,
+    F::Output: Future + Send,
 {
-    type Extract = (<F::Output as TryFuture>::Ok,);
-    type Error = <<F::Output as TryFuture>::Error as CombineRejection<T::Error>>::One;
-    type Future = AndThenFuture<T, F>;
+    type Extract = (<F::Output as Future>::Output,);
+    type Error = T::Error;
+    type Future = ThenFuture<T, F>;
     #[inline]
     fn filter(&self, _: Internal) -> Self::Future {
-        AndThenFuture {
+        ThenFuture {
             state: State::First(self.filter.filter(Internal), self.callback.clone()),
         }
     }
@@ -34,12 +32,11 @@ where
 
 #[allow(missing_debug_implementations)]
 #[pin_project]
-pub struct AndThenFuture<T, F>
+pub struct ThenFuture<T, F>
 where
     T: Filter,
     F: Func<T::Extract>,
-    F::Output: TryFuture + Send,
-    <F::Output as TryFuture>::Error: CombineRejection<T::Error>,
+    F::Output: Future + Send,
 {
     #[pin]
     state: State<T::Future, F>,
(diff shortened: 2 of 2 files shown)


apiraino:
hi @gtsiam I was called from the issue you linked, I'm really interested in understanding your PR and how in practice could be used. Please fix my understanding.

By reading again Sean's [comment](https://github.com/seanmonstar/warp/issues/712#issuecomment-697031645) my hunch is that your PR is trying to draw a clearer line between a `Rejection` (i.e. "current Filter cannot handle this request, goto next one or fail with a `warp::reject::not_found`") and an actual HTTP error code I want to return to the client (possibly bypassing all remaining filters?). All to avoid the current (imho) hack of abusing the Rejection type by adding custom rejections and then using `.recover()` on them to emit the correct reply. 

Can you please show here an example of how to use your `map_async` returning a Future that (I'm assuming your intent) will be converted into a Reply?

Regardless, thank you from me for working on this.

gtsiam:
Pretty much yes, except you're not bypassing any filters - `map_async` is _exactly_ like `map`, only async. You still have to map your error, if it isn't already, into something `impl Reply`.

For example I have something like this:
```rust
async fn login(state: State, req: LoginRequest) -> Result<LoginResponse, LoginError> { /* ... */ }
fn json_response<T: Serialize, E: Serialize>(res: Result<T, E>) -> reply::Json { /* ... */ }

// In my filter function:
path!("login")
    .and(post())
    .and(with_state)
    .and(body::json())
    .map_async(login)
    .map(json_response);
```
Notice that `json_response` takes in a `Result<T, E>`, not just `T`. Also nothing compels me to use result here, other than the fact that my code calls for it. Just like `map`, `map_async` doesn't care what you return. It only forwards it to the next filter.

---

Warp's api today looks like this:
| Closure returns | Function |
|-|-|
|`T`|`map()`|
|`Result<T, Rejection>`|???|
|`Future<Output = T>`|???|
|`Future<Output = Result<T, Rejection>>`|`and_then()`|

The idea is to fill in the second gap with `map_async()` so that you don't have to deal with rejections for application-level errors in async code. Arguably, there are two gaps, but I doubt a sync version of `and_then()` is really worth the effort.

This would make the transition from sync to async as simple as changing your function to return `impl Future` and using `map_async` instead of `map`.

Then there's the `?` operator. Imagine, for example, you have to make a call to a database using async:
```rust
#[derive(Debug, From, Serialize)]
enum LoginError {
    // Other variants here. InternalError is a catch-all for, well, internal errors.
    Internal(#[from(forward)] api::InternalError),
}

// No need for impl Reject for LoginError. Warp never deconstructs the returned Result.

async fn login(state: State, req: LoginRequest) -> Result<LoginResponse, LoginError> {
    let conn = state.pool.get()?; // No need to map_err(|err| reject::custom(err)) !
    
    // [...] Stuff that doesn't really matter for this example
}

```
You are already able to do just this with `map()`, but only for sync code.

There's also [infallible async endpoints](https://github.com/seanmonstar/warp/issues/594#issuecomment-641484239) where you have to do stuff like this because of type inference:
```rust
.and_then(|data| async move {
        let response = data.get().await;
        Ok::<_, Rejection>(warp::reply::json(&response))
    });
```

But with `map_async`:
```rust
.map_async(|data| async move {
        let response = data.get().await;
        warp::reply::json(&response)
    });
```

Also a fundamental issue with using `Rejection` for application errors is that if you forget to handle an error in `.recover()`, then warp has to check all other handlers. Where are rust's compile-time guarantees?

I'd wager there's also some overhead to using recover to check for a lot of errors, though I could be wrong. `Rejection` is not magic. It is, essentially, a linked list of errors. Every `find` has to go through it (though I'm not sure how long it grows in practice, I'd have to check when I have time for it)

(Also while writing, I just realized this probably closes #594 too - I'll edit the description)

apiraino:
Fantastic explaination, thank you so much. You clarified what I hoped for (i.e. remove all the custom Rejections which are really confusing).

> Also a fundamental issue with using Rejection for application errors is that if you forget to handle an error in .recover(), then warp has to check all other handlers. Where are rust's compile-time guarantees?

Also, I suspect, this could potentially solve another annoying side-effect: if a request goes through the chain of filters and reaches the end without being "picked up" and handled I could avoid the `MethodNotAllowed` pile-up (see [this comment for example](https://github.com/seanmonstar/warp/issues/622#issuecomment-666183891)) that got me a couple of times.

lkts:
This is a great addition. I have just followed the same trap - using `and_then` for async handler, type signature suggests my errors should map to `Rejection`, i try to implement `recover` and start question myself if it's the right thing to do and apparently it is not.

kaj (approved):
I can't say I understand all the details of implementation and pinning of the future, but it looks good to me, and seems to be similar to and different from the `and_then` implementation in the right ways.  I like the doc changes to `Filter::and_then`.

The changes is `src/filter/and_then.rs` is not really relevant to implementing map_async, but nice cleanup.

gtsiam:
Good to hear :)

I ended up with this after copy-pasting the `and_then` implementation, renaming it, then incrementally fixing it until I ended up with something that resembled the `map_async` I wanted - then came the fun part of trying to reason through it to make sure it made sense - so yeah... point is; should be very similar.

As for `src/filter/and_then.rs`, entirely my own pet-peeve what can I say.

gtsiam:
I was hoping the convinient "Update" branch button would just rebase it... guess not. Let me do it manually then.

gtsiam:
Actually it just occurred to me github could probably merge that anyway and rebasing messes with reviews... so uh.... oops?
Well, not much I can do about it now.

kaj:
My review still stands, but I'm no maintainer ...

apiraino:
@seanmonstar @jxs genlty bumping for a review of this PR

thanks!

option-greek:
Hello all,
   If I have to use this feature (map_async), which version of warp should I be pulling in from crates.io?

apiraino:
@option-greek this is a branch from a fork waiting for a review from the maintainers. As such, it is not published anywhere. If you want to test it, you need to checkout this fork and use it in your project with something like:
```
warp = { git="https://github.com/gtsiam/warp.git", rev="b16c88" }
```

(I didn't test the above import)

option-greek:
Got it. I'll try it out and post here the results. Btw, didn't realise
import directly from github and that too from a commit was possible!

rubycut:
Any ETA for this PR?

seanmonstar (approved):

gtsiam:
🎉

@rubycut ETA: -1m

jakajancar:
This is great to have warp return an error from a handler and be done with it, instead of continuing to try a bunch of other filters.

However, how should one handle various "extractor" filters that are fallible? For example, if you have an `.and(user(db))` filter that internally calls e.g. `warp::header()` but then connects to the database which might fail. If such a filter just rejects, other peer filters will still be attempted (perhaps making the very same query 10 times). And you cannot just map to `impl Reply` because you're actually extracting some value that you want to use later.

In this case, "fatal rejections" would still be needed to avoid unnecessary processing, no?

gtsiam:
Let me experiment a bit and I'll draft a few proposals in the coming days. In the meantime, please spin up a new issue for this.

PR #881, opened 2021-07-16
rework multipart::Part::data()

jxs:
this stemmed from having to error handle twice when calling `Part::data()`:
- it seems `Part::data()` doesn't need to be `async`, therefore:
- do we need `Part` to impl `Stream`? It's `Stream` impl is basically `Poll::Ready(self.0.take_data())`
- it also seems to never Err, so we don't need it's signature to be `Result`

### DIFF
--- a/src/filters/multipart.rs
+++ b/src/filters/multipart.rs
@@ -8,7 +8,6 @@ use std::io::{Cursor, Read};
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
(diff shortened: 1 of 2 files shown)


paolobarbolini:
How's this going to work with #846?

jxs:
ah, duh! I completely forgot the existing issue with `Multipart` and `Stream` as my use case did not need it. I am going to close this one and take the time to change the label on #846 since you have marked it ready for review

PR #885, opened 2021-07-28
Simplify websockets_chat example to avoid unnecessary `Result`.

yaymukund:
When I used the websockets example, I found myself wondering why you need to `send(Ok(Message...))` rather than `send(Message...)`. It turns out [`stream.forward()` is only implemented on `TryStreamExt`](https://github.com/rust-lang/futures-rs/issues/2004).

So I rewrote the example to use `send()` which avoids the unnecessary `Result`.

### DIFF
--- a/examples/websockets_chat.rs
+++ b/examples/websockets_chat.rs
@@ -5,7 +5,7 @@ use std::sync::{
     Arc,
 };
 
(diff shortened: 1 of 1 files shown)


jxs (approved):
LGTM, thanks!

PR #887, opened 2021-08-18
Make TLS ciphersuites configurable

AlessandroBono:
Hello, I would like to make the TLS ciphersuites configurable. Let me know how to proceed if you are not comfortable to expose `SupportedCipherSuite` from `rustls`.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -12,6 +12,8 @@ use hyper::server::conn::AddrIncoming;
 use hyper::service::{make_service_fn, service_fn};
 use hyper::Server as HyperServer;
 use tokio::io::{AsyncRead, AsyncWrite};
+#[cfg(feature = "tls")]
+use tokio_rustls::rustls::SupportedCipherSuite;
 use tracing::Instrument;
 
 use crate::filter::Filter;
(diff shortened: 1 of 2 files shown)


jxs:
Hi, and thanks for the interest!
Yeah ideally we don't want to expose internal dependencies to avoid them introducing breaking changes. But this would involve three indirections right? `SupportedCipherSuite` and `Tls12CipherSuite` /`Tls13CipherSuite` honestly I'd suggest punting this one, wdyt @seanmonstar .   
Meanwhile if you [convert your warp filters to `Service`](https://docs.rs/warp/0.3.1/warp/fn.service.html#example) and roll them on Hyper with `tokio_rustls` `TlsAcceptor` see [here also for an axum example](https://github.com/tokio-rs/axum/blob/main/examples/tls-rustls/src/main.rs#L26-L47) on how to that

seanmonstar:
Yea, I agree with @jxs. We've purposefully not exposed the `rustls` dependency, which allows us to upgrade to new versions transparently. So, I wouldn't want to expose the `SupportedCipherSuite` type.

We _could_ consider allowing you to just pass a list of strings, and parse them internally. I don't know if that's a good idea, just one I had. The other alternative is what we've suggested for anyone needing more advanced configuration: convert the filters into a service using `warp::service`, and then use rustls and hyper to make the listener part.

PR #891, opened 2021-09-06
Moved IsReject from sealed to expose its function

tiptenbrink:
**Why this change?**
Warp supports almost all HTTP error codes. As such, when making a bad request to a warp server, warp will throw the appropriate error and return the correct status code. However, when using custom error handling (with a recover filter) it is not possible to use these builtin rejections. This simple change exposes IsReject, which exposes the status code and other things for use in custom error handling.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -73,7 +73,7 @@ use http::{
 };
 use hyper::Body;
 
-pub(crate) use self::sealed::{CombineRejection, IsReject};
+pub(crate) use self::sealed::{CombineRejection};
 
 /// Rejects a request with `404 Not Found`.
 #[inline]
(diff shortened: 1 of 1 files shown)


seanmonstar:
> However, when using custom error handling (with a recover filter) it is not possible to use these builtin rejections.

You can currently check a rejection using `.find::<Type>()`. The sealed methods are meant to stay private.

tiptenbrink:
> > However, when using custom error handling (with a recover filter) it is not possible to use these builtin rejections.
> 
> You can currently check a rejection using `.find::<Type>()`. The sealed methods are meant to stay private.

Yes, but as far as I'm aware, this means having to check for each different Rejection error individually if you want to find out the correct status code (or other info). There is no blanket way to just pick a few errors you want to handle in a custom way, while leaving the rest to be handled as default. In the rejections.rs example on the repo, custom error handling leads to every error that is not specifically handled (of which there are more than a dozen) to instead default to a 500 Internal Server Error, which isn't very pretty.

seanmonstar:
> custom error handling leads to every error that is not specifically handled (of which there are more than a dozen) to instead default to a 500 Internal Server Error

They shouldn't default to a 500. After specifically handling the ones you care about, you can return the rejections and most will be turned into a corresponding 4xx response. Usually only the "something really bad happened" and "custom" rejections will render as a 500.

tiptenbrink:
> 
> They shouldn't default to a 500. After specifically handling the ones you care about, you can return the rejections and most will be turned into a corresponding 4xx response. Usually only the "something really bad happened" and "custom" rejections will render as a 500.

My apologies, I do see there is a way to fully emulate the default error behavior for those. I do hope a way to still expose some info about rejections will be possible, because I would like to wrap some additional text around the default rejection string. But since I cannot access those, I believe that's still not possible.
Also, thanks a lot for this library, I've enjoyed using it!

PR #898, opened 2021-09-12
Fix rust edition idioms

nickelc:


nickelc:
:thinking: The dead code shows up since `nightly-2021-09-11`

seanmonstar (approved):
Thanks!

PR #906, opened 2021-10-04
Replace `futures` with `futures-util`

nylonicious:
Replace usage of `futures` with `futures-util` crate.

- `cargo tree` looks much cleaner
- Removes unused `futures-io` dependency.


nylonicious:
I think it should be fine now.

PR #907, opened 2021-10-08
Implement Sink and Stream for WsClient

FSMaxB:
This implements `Sink` and `Stream` for `WsClient`.

### Why is this useful?
If you use `tokio-tungstenite` as your websocket client, you will usually have a `tokio-tungstenite::WebSocketStream` to work with, which exclusively uses the `Sink` and `Stream` traits for sending and receiving messages. Implementing `Sink` and `Stream` for `WsClient` as well allows for writing code that works with both the real `tokio-tungstenite::WebSocketStream` and `WsClient` by relying only on the `Sink` and `Stream` traits in the implementation, especially if you combine it with #903 which provides conversions between `warp::ws::Message` and `tungstenite::protocol::Message`.

### Why not use `tokio::sync::mpsc::unbounded_channel`
Initially I tried to implement this using the tokio mpsc channels that `WsClient` was using, but found it to be impossible.
Implementing `Stream` isn't an issue because of the `Stream` implementation provided in `tokio-stream` and because `tokio::sync::mpsc::UnboundedReceiver` has a `poll_recv` method exactly for that purpose.
Implementing `Sink` however doesn't work because there is no way that `poll_flush` or `poll_close` can notify a task to be polled again.

Therefore this PR switches the channel implementation from `tokio` to `futures`, since `futues::channel::mpsc::UnboundedSender` and `UnboundedReceiver` already implement `Sink` and `Stream` and the implementation on `WsClient` then only need to proxy their calls to the underlying channel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ all-features = true
 async-compression = { version = "0.3.7", features = ["brotli", "deflate", "gzip", "tokio"], optional = true }
 bytes = "1.0"
 futures-util = { version = "0.3", default-features = false, features = ["sink"] }
+futures-channel = { version = "0.3.17", features = ["sink"]}
 headers = "0.3"
 http = "0.2"
 hyper = { version = "0.14", features = ["stream", "server", "http1", "tcp", "client"] }
--- a/src/test.rs
+++ b/src/test.rs
@@ -92,6 +92,7 @@ use std::pin::Pin;
 use std::task::{self, Poll};
 
 use bytes::Bytes;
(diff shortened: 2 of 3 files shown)


FSMaxB:
I'm going to rebase this shortly since merging #906 broke this pull request. This PR now also needs to add `futures-channel` on top of the `futures-util` dependency currently on master.

jxs (approved):

jxs:
Hi, and sorry for the delay! Makes sense, thanks!

PR #909, opened 2021-10-16
Implemented Reply for Result<impl Reply, impl Reply>

gtsiam:
#878 was originally meant to work with this change, however #458 which originally provided it was closed after #878 was merged. And so, I'm providing a minimal subset of #458 to merge. I'm leaving the example changes from #458 up to a future pr, since those could be subject to discussions unrelated to this particular change.

**Changes**:
- Implemented `Reply` for `Result<impl Reply, impl Reply>`.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -411,18 +411,24 @@ impl Reply for ::http::StatusCode {
     }
 }
 
-impl<T> Reply for Result<T, ::http::Error>
+impl Reply for ::http::Error {
+    #[inline]
+    fn into_response(self) -> Response {
+        tracing::error!("reply error: {:?}", self);
+        StatusCode::INTERNAL_SERVER_ERROR.into_response()
+    }
+}
+
+impl<T, E> Reply for Result<T, E>
 where
-    T: Reply + Send,
+    T: Reply,
+    E: Reply,
 {
     #[inline]
     fn into_response(self) -> Response {
         match self {
             Ok(t) => t.into_response(),
-            Err(e) => {
-                tracing::error!("reply error: {:?}", e);
-                StatusCode::INTERNAL_SERVER_ERROR.into_response()
-            }
+            Err(e) => e.into_response(),
         }
     }
 }


MaxFangX (approved):
Not a maintainer, but I rebased these changes [on my own branch](https://github.com/MaxFangX/warp/tree/reply-for-result) and the new impl works wonderfully for removing the `match` statements everywhere in my handlers and replacing them with concise `?`s.

Using this impl I could define a new error type and impl `Reply` on it, as well as `From` for library error types:
```rust
use thiserror::Error;
use warp::reply::Response;
use warp::Reply;

#[derive(Error, Debug)]
pub enum ApiError {
    #[error("Database error")]
    DatabaseError,
    #[error("Object not found")]
    NotFound,
    #[error("Object could not be serialized into JSON")]
    JsonSerializeError,
}

impl Reply for ApiError {
    fn into_response(self) -> Response {
        self.to_string().into_response()
    }
}

impl From<sea_orm::DbErr> for ApiError {
    fn from(_err: sea_orm::DbErr) -> ApiError {
        ApiError::DatabaseError
    }
}

impl From<serde_json::Error> for ApiError {
    fn from(_err: serde_json::Error) -> ApiError {
        ApiError::JsonSerializeError
    }
}
```

The resulting handler looks like this:
```rust
pub async fn get_node(db: Db) -> Result<impl Reply, ApiError> {
    // Each use of ? concisely calls into one of the From impls
    let node = node::Entity::find()
        .filter(node::Column::UserId.eq(1))
        .one(&*db)
        .await?
        .ok_or(ApiError::NotFound)?;

    let json = serde_json::to_string(&node)?;
    Ok(json)
}
```

Which allowed me to terminate without any `recovers`, `Rejection`s, or other unintended constructs:
```rust
let route = warp::path("node")
    .and(warp::post())
    .and(with_db(db.clone()))
    .and(warp::body::json())
    .then(v1::get_node);
```

Rebasing on master had 0 conflicts so it should be super quick, but if @gtsiam is busy I can make a fresh PR.

Thank you for implementing this!

gtsiam:
I don't like pinging like this, but @jxs @seanmonstar: Since there seems to be interest in this, could someone take a look? I intentionally made the change as simple as humanly possible to make merging easy.

MaxFangX:
Iterating on your idea @jakajancar, it turns out that if you call `into_response()` provided by the `Reply` trait, you can avoid the heap allocation.

Given e.g. a handler that returns a `Result<S, E>` where `S` implements `Serialize` and `E` implements `Reply`
```rust
#[derive(Error, Debug)]
pub enum ApiError {
    #[error("Database error")]
    DatabaseError,
    #[error("Object not found")]
    NotFound,
}

impl Reply for ApiError {
    fn into_response(self) -> Response<Body> {
        Response::builder()
            .status(StatusCode::INTERNAL_SERVER_ERROR)
            .body(self.to_string().into())
            .expect("Could not construct Response")
    }
}

#[derive(Serialize)]
pub struct NodeInfo {
    pub pubkey: String,
    pub num_channels: usize,
}

pub async fn node_info(
    channel_manager: Arc<ChannelManagerType>,
    peer_manager: Arc<PeerManagerType>,
) -> Result<NodeInfo, ApiError> {
    // do stuff that builds the NodeInfo
    let resp = NodeInfo {
        pubkey,
        num_channels,
    };

    Ok(resp)
}
```

One can have a helper like this:
```rust
/// Converts Result<S, E> into Response<Body>, avoiding the need to call
/// reply::json(&resp) in every handler or to implement warp::Reply manually
fn into_response<S: Serialize, E: Reply>(
    reply_res: Result<S, E>,
) -> Response<Body> {
    match reply_res {
        Ok(resp) => reply::json(&resp).into_response(),
        Err(err) => err.into_response(),
    }
}
```

With a route that looks like this:
```rust
let node_info = warp::path("node_info")
    .and(warp::get())
    .and(inject::channel_manager(channel_manager))
    .and(inject::peer_manager(peer_manager))
    .then(node_info)
    .map(into_response);
```

I think it is pretty clean, and it works with warp today.

cbeck88:
I iterated towards a similar solution -- it would be nice if infrastructure like `fn into_response` above were in some shared utility crate, or warp itself, since many people are likely to reinvent this and then carry it.

gtsiam:
Indeed. But at this point my experience contributing to warp indicates that it is simply abandonware. Generally I've come to expect that any library by @seanmonstar that isn't hyper is or will become [abandonware](https://github.com/hyperium/mime/issues/135) at this point.

If you want a good framework, go with axum. If you already have a large codebase, just write up a quick `ReplyExt` trait for this method and move on with your life :/

seanmonstar:
It is true that I can spend less attention on warp. My previous job filled up my time with corporate overhead, and a prioritization towards things the company specifically used. It's not a great excuse, it's just the reason. I'm sorry that makes contributing frustrating.

I am happy to onboard others who can pay more attention, as I did with @jxs a couple years ago. The way I'm able to build trust like that is when someone starts doing reviews, and reviews in a way that aligns with the previous issues/reviews and goals of the project.

---

With regards to this PR, I remember seeing it a while ago and feeling uncertain about `Result<impl Reply, impl Reply>`, since I feel like the Rejection and Error system in warp is probably not quite as elegant as I had hoped. But then other things crowded that out my mind. Looking again now, I think this is perfectly fine. If at some point the Rejection system were to be redesigned, and this were in conflict, we can always change it then. (Basically, I'm stopping a "what-if-future" from getting in the way of "would-be-nice-now".)

seanmonstar (approved):

seanmonstar:
Thank you @gtsiam for the PR, and for reminding me existed!

gtsiam:
That was fast!

@seanmonstar If you're looking for maintainers, consider myself volunteered.

PR #912, opened 2021-11-04
Feat: split compression features for Gzip and Brotli support

Some-Dood:
Hello there! This PR splits the `compression` feature into two: `compression-brotli` (Brotli) and `compression-gzip` (Deflate and Gzip). As their names suggest, each feature is responsible for enabling support for their respective compression algorithms.

This is done in an effort to minimize the `cargo tree` for those who do not intend to use one or the other. Allowing the `warp` consumer to choose which compression algorithms to support gives finer-grained control over their dependencies, which is nice!

The `async-compression` crate pulls in the `brotli` crate when that feature is enabled. Similarly, it pulls in the `flate2` crate when the `deflate` feature or the `gzip` feature is enabled. Observe that the `deflate` and `gzip` features have been merged as the `compression-gzip` feature in `warp` since they both indirectly pull in the `flate2` crate anyway.

In order to maintain backwards-compatibility, the `compression` feature is now set to enable both `compression-brotli` and `compression-gzip`, which is the current behavior prior to this change.

To that end, if there are any issues with this PR, I'd be glad to resolve them. Thanks! 🎉

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ edition = "2018"
 all-features = true
 
 [dependencies]
-async-compression = { version = "0.3.7", features = ["brotli", "deflate", "gzip", "tokio"], optional = true }
+async-compression = { version = "0.3.7", features = ["tokio"], optional = true }
 bytes = "1.0"
 futures-util = { version = "0.3", default-features = false, features = ["sink"] }
 futures-channel = { version = "0.3.17", features = ["sink"]}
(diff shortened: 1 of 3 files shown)


seanmonstar (approved):
This looks excellent, thank you!

I have another PR running through CI which I think will cause a conflict with this one, so you may need to update in a few...

Some-Dood:
Alrighty, everything seems to be ready! I have rebased the changes onto the latest `master` to keep the Git history clean. 🚀

PR #919, opened 2021-11-23
Add option to ignore content-length with multipart body

outamaa:
The current implementation of `multipart::form` requires that the `Content-Length` header exists. However, there are valid situations where this is not the case, for example when `Transfer-Encoding` header is present. See e.g. [RFC2616, section 4.4](https://datatracker.ietf.org/doc/html/rfc2616#section-4.4).

This PR adds an option to not check the content length, thus allowing the `multipart::form` filter to be used in cases where `Content-Length` header is not present. The change does not break the existing API.

### DIFF
--- a/src/filters/multipart.rs
+++ b/src/filters/multipart.rs
@@ -25,7 +25,7 @@ const DEFAULT_FORM_DATA_MAX_LENGTH: u64 = 1024 * 1024 * 2;
 /// Create with the `warp::multipart::form()` function.
 #[derive(Debug, Clone)]
 pub struct FormOptions {
-    max_length: u64,
+    max_length: Option<u64>,
 }
 
 /// A `Stream` of multipart/form-data `Part`s.
@@ -51,7 +51,7 @@ pub struct Part {
 /// in turn is a `Stream` of bytes.
 pub fn form() -> FormOptions {
     FormOptions {
-        max_length: DEFAULT_FORM_DATA_MAX_LENGTH,
+        max_length: Some(DEFAULT_FORM_DATA_MAX_LENGTH),
     }
 }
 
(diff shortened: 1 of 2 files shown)


jxs:
Hi, thanks for your interest! and sorry for the delay. I think this should rather be part of https://github.com/seanmonstar/warp/pull/846/ what do you think?

outamaa:
> Hi, thanks for your interest! and sorry for the delay.

No problem. Hope this proves useful to somebody else besides me. :)

> I think this should rather be part of #846 what do you think?

I really have no strong preference one way or the other, though I do think of this as a separate feature to multipart being streamable. How would you suggest going forward with this?

xd009642:
Small bump as I was wondering what's taken to progress this as #846 is closed and multipart stuff in warp seems to have been redone recently but this feature still isn't present :eyes:

outamaa:
I now merged the changes from upstream master into this branch.

src/filters/multipart.rs
```diff
@@ -61,7 +61,13 @@ impl FormOptions {
     ///
     /// Defaults to 2MB.
     pub fn max_length(mut self, max: u64) -> Self {
-        self.max_length = max;
+        self.max_length = Some(max);
+        self
+    }
+
+    /// Set no limit to maximum byte length for the body.
+    pub fn no_max_length(mut self) -> Self {
```
seanmonstar, line 69: We could change to `max_length(impl Into<Option<u64>>)`. All previous calls will still work, but you can now pass `None`. And then document that.

outamaa: Sounds good, thanks!

outamaa: Done.

seanmonstar (approved):

PR #924, opened 2021-12-11
Make file serving helpers public

### LINKED ISSUE #171: [Question] Dynamic file Filter
Hi, I have a question on how to properly use the `fs::file` Filter.

I'd like to have something like this: 
```rust
	let route = warp::get2()
		.and(path!("files" / String))
		.and(warp::path::end())
		.and_then(get_file_path)
		.and_then(|filepath| warp::fs::file(filepath));
```
Where `get_file_path` would be a `Future` returning a `PathBuf`.

Is there any way to achieve this?

Thanks, and great project BTW :smiley:

arctic-hen7:
This makes some of the contents of [fs.rs](https://github.com/seanmonstar/warp/blob/master/src/filters/fs.rs) public so that users can serve static files at paths based on the request without having to implement file streaming manually.

Fixes #171.

### DIFF
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -132,12 +132,14 @@ fn sanitize_path(base: impl AsRef<Path>, tail: &str) -> Result<PathBuf, Rejectio
     Ok(buf)
 }
 
-#[derive(Debug)]
-struct Conditionals {
-    if_modified_since: Option<IfModifiedSince>,
-    if_unmodified_since: Option<IfUnmodifiedSince>,
-    if_range: Option<IfRange>,
-    range: Option<Range>,
+/// Conditionals that define certain aspects of file serving. These are usually based on the headers of a request, though they can be manually constructed.
+#[allow(missing_docs)] // TODO add docs for these (consult @seanmonstar)
+#[derive(Debug, Default)] // We derive `Default` as well to make constructing this manually easier
+pub struct Conditionals {
+    pub if_modified_since: Option<IfModifiedSince>,
+    pub if_unmodified_since: Option<IfUnmodifiedSince>,
+    pub if_range: Option<IfRange>,
+    pub range: Option<Range>,
 }
 
 enum Cond {
@@ -195,7 +197,8 @@ impl Conditionals {
     }
 }
 
-fn conditionals() -> impl Filter<Extract = One<Conditionals>, Error = Infallible> + Copy {
+/// A filter for extracting certain headers related to static file serving.
+pub fn conditionals() -> impl Filter<Extract = One<Conditionals>, Error = Infallible> + Copy {
     crate::header::optional2()
         .and(crate::header::optional2())
         .and(crate::header::optional2())
@@ -242,9 +245,9 @@ impl File {
     }
 }
 
-// Silly wrapper since Arc<PathBuf> doesn't implement AsRef<Path> ;_;
+/// A wrapper for `Arc<PathBuf>` that implements `AsRef<Path>`.
 #[derive(Clone, Debug)]
-struct ArcPath(Arc<PathBuf>);
+pub struct ArcPath(pub Arc<PathBuf>);
 
 impl AsRef<Path> for ArcPath {
     fn as_ref(&self) -> &Path {
@@ -258,7 +261,9 @@ impl Reply for File {
     }
 }
 
-fn file_reply(
+/// An internal helper for serving static files. Usually, you'll want to use this with `warp::fs::file`, but if the file's path is based on something extracted with a filter,
+/// you'll need to use this manually.
+pub fn file_reply(
     path: ArcPath,
     conditionals: Conditionals,
 ) -> impl Future<Output = Result<File, Rejection>> + Send {
--- a/src/filters/fs.rs
+++ b/src/filters/fs.rs
@@ -134,7 +134,7 @@ fn sanitize_path(base: impl AsRef<Path>, tail: &str) -> Result<PathBuf, Rejectio
 
 /// Conditionals that define certain aspects of file serving. These are usually based on the headers of a request, though they can be manually constructed.
 #[allow(missing_docs)] // TODO add docs for these (consult @seanmonstar)
-#[derive(Debug, Default)] // We derive `Default` as well to make constructing this manually easier
+#[derive(Debug, Default)]
 pub struct Conditionals {
     pub if_modified_since: Option<IfModifiedSince>,
     pub if_unmodified_since: Option<IfUnmodifiedSince>,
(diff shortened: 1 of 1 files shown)


arctic-hen7:
@seanmonstar I'm slightly unsure about how you want some of these internals documented. I've mentioned you in the code for `Conditionals`, the fields of which I haven't documented yet.

arctic-hen7:
@seanmonstar is there anything blocking this in particular?

arctic-hen7:
Okay, not sure if anyone is going to review this but for anyone with the same problems as in #171 I published a temporary fix crate (I needed this immediately) as `warp-fix-171` on crates.io. Hope this helps someone other than me!

Obviously, if this gets merged I'll deprecate that crate.

src/filters/fs.rs
```diff
@@ -132,12 +132,14 @@ fn sanitize_path(base: impl AsRef<Path>, tail: &str) -> Result<PathBuf, Rejectio
...
 }
 
-#[derive(Debug)]
-struct Conditionals {
-    if_modified_since: Option<IfModifiedSince>,
-    if_unmodified_since: Option<IfUnmodifiedSince>,
-    if_range: Option<IfRange>,
-    range: Option<Range>,
+/// Conditionals that define certain aspects of file serving. These are usually based on the headers of a request, though they can be manually constructed.
+#[allow(missing_docs)] // TODO add docs for these (consult @seanmonstar)
```
tobz, line 136: Probably just makes sense to try and come up with a rough first draft of these comments, then people can actually review them.  Easier to give feedback, sometimes, than it is to create from scratch. :)

src/filters/fs.rs
```diff
@@ -132,12 +132,14 @@ fn sanitize_path(base: impl AsRef<Path>, tail: &str) -> Result<PathBuf, Rejectio
...
+#[derive(Debug, Default)] // We derive `Default` as well to make constructing this manually easier
```
tobz, line 137: I think you can drop the comment here, since the whole reason of having a `Default` impl is kind of captured by `Default`'s own doc comments anyways.

src/filters/fs.rs
```diff
@@ -258,7 +261,9 @@ impl Reply for File {
     }
 }
 
-fn file_reply(
+/// An internal helper for serving static files. Usually, you'll want to use this with `warp::fs::file`, but if the file's path is based on something extracted with a filter,
```
tobz, line 264: Good practice is to split the first and second sentence with a blank line, and rustdoc will then use the first sentence as the "high level" text description in `docs.rs`, etc, but all of it will be used if you go directly to the docs for the specific item in question.

tobz (changes requested):

src/filters/fs.rs
arctic-hen7: I think from memory the reason I didn't do that is because I don't actually understand what these do yet. Any pointers?

src/filters/fs.rs
arctic-hen7: I never knew this, thanks for the tip!

src/filters/fs.rs
tobz: Well, these are the various HTTP headers that might come in on a request that could be pertinent to file serving.

So, for each of them, maybe a simple one-line explanation of what the header does and then a link to relevant docs, for example, the RFC that they were standardized in, or perhaps the Mozilla MDN docs for the given header.

arctic-hen7: Thanks for the clarification, this should be fixed now!

tobz (approved):

tobz:
Yep, the comments look a lot better now and will be very useful to developers. 👍🏻

RcolyarBI (approved):

arctic-hen7:
@seanmonstar are there any plans to merge this? I'm very happy to update this to be in sync with `master` if so.

arctic-hen7:
@quinn if it helps there's my public `warp-fix-171` crate, which has this fix. Unless you need the latest features from Warp itself, which that doesn't have.

arctic-hen7:
@seanmonstar I notice the repo has had some activity lately, any chance of this being merged? I've updated everything to be in sync with the master branch.

*(For others: until this is merged, I've released another version of the `warp-fix-171` patch crate.)*

src/filters/fs.rs
```diff
@@ -135,12 +135,19 @@ fn sanitize_path(base: impl AsRef<Path>, tail: &str) -> Result<PathBuf, Rejectio
...
-struct Conditionals {
-    if_modified_since: Option<IfModifiedSince>,
-    if_unmodified_since: Option<IfUnmodifiedSince>,
-    if_range: Option<IfRange>,
-    range: Option<Range>,
+/// Conditionals that define certain aspects of file serving. These are usually based on the headers of a request, though they can be manually constructed.
+#[derive(Debug, Default)]
+pub struct Conditionals {
+    /// Only respond with the file if it has been modified since this date. [See MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since).
+    pub if_modified_since: Option<IfModifiedSince>,
```
seanmonstar, line 142: I would keep the fields private. That way we can add more, or change them. We also haven't exposed the `headers` crate publicly.

src/filters/fs.rs
```diff
@@ -245,9 +253,9 @@ impl File {
     }
 }
 
-// Silly wrapper since Arc<PathBuf> doesn't implement AsRef<Path> ;_;
+/// A wrapper for `Arc<PathBuf>` that implements `AsRef<Path>`.
 #[derive(Clone, Debug)]
-struct ArcPath(Arc<PathBuf>);
+pub struct ArcPath(pub Arc<PathBuf>);
```
seanmonstar, line 258: I would keep this private.

src/filters/fs.rs
```diff
@@ -261,7 +269,11 @@ impl Reply for File {
     }
 }
 
-fn file_reply(
+/// An internal helper for serving static files.
+///
+/// Usually, you'll want to use this with `warp::fs::file`, but if the file's path is based on something extracted with a filter,
+/// you'll need to use this manually.
+pub fn file_reply(
```
seanmonstar, line 276: Also kept private.

seanmonstar (changes requested):
I think adding a filter to get `Conditionals` seems like a good idea! I'd suggest a little different though: Add a method to `Conditionals`, that is `fn reply<P: AsRef<Path>>(self, path) -> impl Reply`.

arctic-hen7:
Yeah you're right that is much cleaner, I'll refactor this PR as soon as I can.

(Terribly sorry for the late reply by the way, I've been really busy the last month!)

arctic-hen7:
Closing because #1049 supersedes this (according to its description).

PR #949, opened 2022-03-22
Expose the custom resolver to the warp interface

behos:
This allows users to inject their own CertResolver directly to rustls
for custom cert resolution.

Currently warp provides an interface for using a single certificate for
the duration of the lifecycle of a service.

In some cases, it's useful to be able to inject a custom resolver to
allow more granular use cases. For example, we want to inject a resolver
which periodically reloads the certs from disk to allow live renewals
without interrupting the service.

Exposing the CertResolver configuration allows any custom use case to be
implemented without having to account for every possibility within warp.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -6,12 +6,16 @@ use std::future::Future;
 use std::net::SocketAddr;
 #[cfg(feature = "tls")]
 use std::path::Path;
+#[cfg(feature = "tls")]
+use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


seanmonstar:
Warp purposefully doesn't expose any types from rustls, so that it can upgrade versions internally without a breaking change.

It also provided the TLS stuff as a convenience, but anything that needs advanced configuration should be done without Warp's helper.

PR #954, opened 2022-04-06
Make sealed public

qwfy:
This commit makes `warp::redirect::sealed::Sealed` public, such that customized `AsLocation` can be implemented outside of this crate.

Motivation:

`http::uri::Uri`, currently the only implements of `AsLocation`, does not handle fragments. As a result, redirecting to an url with a fragment cannot be done.

### DIFF
--- a/src/redirect.rs
+++ b/src/redirect.rs
@@ -6,6 +6,7 @@
 use http::{header, StatusCode};
 
 pub use self::sealed::AsLocation;
+pub use self::sealed::Sealed;
 use crate::reply::{self, Reply};
 
 /// A simple `301` permanent redirect to a different location.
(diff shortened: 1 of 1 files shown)


kaj:
The whole reason for the `Sealed` trait is to not be public.  _If_ we want to make it possible to impl `AsLocation` for other types (which I agree would be useful, but maybe also give some problems, see the docs), I think the solution is not to make `Sealed` public, but to remove the `AsLocation: Sealed` dependency.

kaj:
Also, redirection to a location with a fragment identifier can be done, just not as simple.

Example: https://github.com/kaj/r4s/blob/5c127de04188b6d9d733fa02a84504fc1dfe0eb4/src/server/mod.rs#L567-L571

qwfy:
I'm not sure, the [original comment](https://github.com/seanmonstar/warp/pull/954/commits/6f6464cd9f3653fe01c1d2f358f0996e19aa9025#diff-f9c2a8c37b690ffe48b65bc57bd72784eb946d600d3b00cd5073259a7e32f001R118) says the sealed exists to allow new implementations, though the name "sealed" suggests otherwise, maybe I'm missing a design pattern here.

kaj:
I think "this sealed trait" in that sentence refers to `AsLocation` itself.  It exists so that warp can decide to impl for more types without chaning the `redirect` method, and it is sealed so that such new impls won't be in conflict with any impls provieded by crates using warp.

PR #981, opened 2022-07-03
Use status code for trace filter log level / message.

jakajancar:
Currently, the `warp::trace()` filter assumes that:
  - Ok response from filter => INFO / "finished processing with success"
  - Err / rejection from filter => ERROR / "unable to process request" (5xx); or WARN / "unable to serve request" (4xx)

However, this doesn't work well with custom rejections:

Normally you will `.recover()` the rejection and return a 4xx/5xx reply. `warp::trace()` will currently log INFO / "finished processing with success", as if everything is OK, whereas it should log WARN/ERROR, same as for known rejections.

This PR changes the logging to be based only on the status code.

### DIFF
--- a/src/filters/trace.rs
+++ b/src/filters/trace.rs
@@ -238,35 +238,39 @@ mod internal {
     use tracing::Span;
 
     fn finished_logger<E: IsReject>(reply: &Result<(Traced,), E>) {
-        match reply {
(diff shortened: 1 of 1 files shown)


jxs (approved):

jxs:
Hi, yeah this makes sense thanks!

PR #984, opened 2022-08-04
Implement HTTP handling for `TlsServer`

daxpedda:
This allows hosting a HTTP and HTTPS server on the same port. The main use-case is upgrading a connection to HTTPS automatically when going to e.g. "localhost:3030", as browsers connect via HTTP unless explicitly specified.

I implemented in a way to be optional while not disturbing the default path.

This is a first step towards #417. I learned a lot since then :sweat_smile:.

### DIFF
--- a/examples/dual.rs
+++ b/examples/dual.rs
@@ -1,0 +1,26 @@
+#![deny(warnings)]
+
+// Don't copy this `cfg`, it's only needed because this file is within
+// the warp repository.
+// Instead, specify the "tls" feature in your warp dependency declaration.
+#[cfg(feature = "tls")]
+#[tokio::main]
+async fn main() {
+    use warp::Filter;
+
+    // Match any request and return hello world!
+    let routes = warp::any().map(|| "Hello, World!");
+
+    warp::serve(routes)
+        .tls()
+        .http(true)
+        .cert_path("examples/tls/cert.pem")
+        .key_path("examples/tls/key.rsa")
+        .run(([127, 0, 0, 1], 3030))
+        .await;
+}
+
+#[cfg(not(feature = "tls"))]
+fn main() {
+    eprintln!("Requires the `tls` feature.");
+}
--- a/src/server.rs
+++ b/src/server.rs
@@ -561,6 +561,13 @@ where
         });
         (addr, fut)
     }
+
+    /// Enable this server to **also** handle HTTP requests.
+    ///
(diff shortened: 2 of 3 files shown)


jxs:
Hi,
This is great work, looks correct to me! I think we can implement the forward from HTTP to HTTPS in this PR and rename the `http` function to something like `upgrade` wdyt?

daxpedda:
My thinking was that maybe somebody is interested in this feature without the automatic upgrade.
I was planning to implement the upgrade functionality as a filter, but if the feature of handling both protocols to begin with is undesired, what you propose makes sense.

So is this standalone feature undesired?

notgull:
I'm thinking about the downside of allowing this without the "upgrade" is that using HTTP that doesn't immediately redirect to HTTPS is problematic. At my place of work, it's considered a vulnerability. If anything, it should be opt-out of upgrading, not opt in, since it makes your program significantly more insecure.

daxpedda:
Yeah, that sounds reasonable to me. I don't think I will even implement the opt-out, because personally I don't have a use-case, so we can do it when somebody requests it.

Will look into implementing the automatic upgrade soon.

jxs:
>My thinking was that maybe somebody is interested in this feature without the automatic upgrade.

ah good question, but What would be the use case for that?

>I'm thinking about the downside of allowing this without the "upgrade" is that using HTTP that doesn't immediately redirect to HTTPS is problematic. At my place of work, it's considered a vulnerability. 

interesting, yeah I was also wondering about that, do you have more info on the possible attack vector?

daxpedda:
> > My thinking was that maybe somebody is interested in this feature without the automatic upgrade.
> 
> ah good question, but What would be the use case for that?

I don't know, I was only doing it to make the implementation easier, I can't even come up with a use-case.
> > I'm thinking about the downside of allowing this without the "upgrade" is that using HTTP that doesn't immediately redirect to HTTPS is problematic. At my place of work, it's considered a vulnerability.
> 
> interesting, yeah I was also wondering about that, do you have more info on the possible attack vector?

I'm assuming if you set up a HTTPS server you don't actually want anybody connecting over HTTP. So if one forgets to setup the automatic upgrade, a user could potentially use the website completely over HTTP instead of over HTTPS. An attacker could use this by just sending you the HTTP link, even if you check the domain, you might not notice the scheme.

PR #988, opened 2022-08-12
Fix test example so it doesn't fail

vargad:
Comment says "Or simply test if a request matches (doesn't reject)"
therefore I would expect for assert to test for success. Also this
example (for sum) will always match. It's confusing the code example
actually tests if it doesn't match.


jxs (approved):

jxs:
Hi,
yeah this makes sense, thanks!

PR #990, opened 2022-08-19
Add local() SocketAddr filter

daviessm:
I found a need for using the local part of the open connection in a filter. This PR simply adds a local_addr wherever there's a remote_addr (except in tracing).

It needs a bump to `hyper` as local_addr() was only added to `AddrStream` in v0.14.19.


Cargo.toml
```diff
@@ -23,7 +23,7 @@ futures-util = { version = "0.3", default-features = false, features = ["sink"]
 futures-channel = { version = "0.3.17", features = ["sink"]}
 headers = "0.3"
 http = "0.2"
-hyper = { version = "0.14", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
+hyper = { version = ">=0.14.19", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
```
seanmonstar, line 26: The `>=` isn't needed (in fact, it makes it accept ANY new version). Setting `0.14.19` will mean at least that, while still within `0.14.x`.

daviessm: Got it, fixed in 1bc098d32c678ddfbfedbb1cdda2992009cdc705, thanks

PR #992, opened 2022-08-22
Expose the inner websocket reference to allow code-reuse.

yuguorui:
Currently, warp::filters::ws::{WebSocket,Message} are two thin layers
built upon crate tokio_tungstenite, user may prefer to use the vanilla
WebSocket interface to share the code between the server and client.

Expose the inner websocket reference with get_{inner, inner_mut}.

Signed-off-by: yuguorui <yuguorui@pku.edu.cn>

### DIFF
--- a/src/filters/ws.rs
+++ b/src/filters/ws.rs
@@ -200,6 +200,16 @@ impl WebSocket {
     pub async fn close(mut self) -> Result<(), crate::Error> {
         future::poll_fn(|cx| Pin::new(&mut self).poll_close(cx)).await
     }
+
+    /// Get inner WebSocketStream.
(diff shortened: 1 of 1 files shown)


seanmonstar:
The reason for the thin layers is exactly so that we *don't* expose the types. We want to be able to upgrade tungstenite without users depending on an exact version.

yuguorui:
Yes, exposing different versions of tungstenite to users by default could be a disaster. But could we have any non-default feature flag like `expose_ws`?

If the answer is affirmative, then I will be happy to provide the corresponding PR.

seanmonstar:
Making it possible to access the type would make it impossible to upgrade the version.

yuguorui:
Thanks for your quick response, I have closed this.

Issue #738, opened 2020-11-03
path with an extra initial /


mjibson:
I wrote a client with a small bug where it makes a request to a warp server with an extra `/` at the front of the path: `//api/v1/blah`. My reading of the code suggests that it is currently not possible to make a warp filter that can match this because the path function rejects empty parts and parts with slashes. Is there a way to do this?

anmacdonald:
I've also hit this. I've found an interesting blog on the topic: https://www.ctrl.blog/entry/relative-double-slashes-url.html

Currently, it looks like the only way round this problem is to use a reverse proxy to merge the slashes on warp's behalf.

jxs:
Hi! You can try using [`warp::service`](https://docs.rs/warp/latest/warp/fn.service.html) to convert the filters to a service and then get the `hyper` [request and alter it's path](https://docs.rs/hyper/latest/hyper/struct.Request.html#method.uri_mut).
See how warp uses `service_fn` for an example where you can get the `hyper::Request`

anmacdonald:
Cheers, @jxs! That's a big help. I'll investigate this workaround.

clickonetwo:
@jxs Did you get anywhere with this?  @seanmonstar Any chance we could get a new filter behavior to canonicalize multiple slashes?

brotskydotcom:
I have submitted PR #993 which adds a feature `ignore-empty-path-segments` that works around this problem by having the path segment filters ignore multiple slashes between segments.

PR #993, opened 2022-09-02
Add a feature to ignore empty path segments (// in path).

brotskydotcom:
When the new `ignore-empty-path-segments` feature is enabled, multiple leading slashes at the start of the path are ignored (the first segment starts after them), and multiple slashes at the end of any matched segment are ignored (so the next segment starts after them).  In addition, the `end` filter will ignore multiple slashes that follow the final segment. Both the `fullpath` and `tail` filters will still return any multiple slashes in their matched text, but their `segments` iterator will ignore the empty segments (in fact it does this today even without this feature being set).

This change does _not_ include this feature in the list of default features.

This change includes path filter testing on paths with multiple slashes in them.

This change fixes seanmonstar/warp#738.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,9 @@ default = ["multipart", "websocket"]
 websocket = ["tokio-tungstenite"]
 tls = ["tokio-rustls"]
 
+# allow double slashes in paths
+ignore-empty-path-segments = []
+
 # Enable compression-related filters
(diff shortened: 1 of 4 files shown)


brotskydotcom:
@seanmonstar This PR does not update docs.  If you are interested in merging it but need me to write docs first just let me know.  All review comments appreciated.

shinjitumala:
Any updates on this?

Issue #859, opened 2021-05-31
Non-constant time comparison of secret header values can enable timing attacks


codahale:
**Version**
`master`

**Platform**
all

**Description**
Currently, `warp::filters::header::exact` uses [a simple `==` comparison on header values to filter](https://github.com/seanmonstar/warp/blob/master/src/filters/header.rs#L148). For non-secret values (e.g., `dnt=1`), that’s totally fine. For secret values (e.g., [bearer tokens in headers](https://github.com/seanmonstar/warp/blob/master/examples/todos.rs#L86)), an attacker can measure the time the server takes to reject various values and [determine the secret value byte by byte](https://codahale.com/a-lesson-in-timing-attacks/).

Comparing secret values to client-provided values should be done with constant-time comparison routines, such as those in [subtle](https://doc.dalek.rs/subtle/trait.ConstantTimeEq.html).

seanmonstar:
Thanks for the write-up. I'm aware of timing attacks. I don't believe this is an issue with warp itself. Anyone comparing genuine user secrets hopefully don't have the secrets in plaintext, but are using a strong hash function (like bcrypt or similar), which many may provide a safer compare function as well.

codahale:
Password hashing isn’t typically used for bearer tokens, but bearer tokens are still instances of secret information.

If the intent is that `exact` not be used for comparing client-provided values to secret information, it may be wise to change the linked example to reflect that. As it stands, it compares client-provided bearer tokens to a static value in an insecure manner with a comment immediately above it saying that the given usage is “to show how authentication filters are used“.

PR #998, opened 2022-10-11
Implemented constant time exact header filter

kleinschrader:
Hi,

this pull request serves to solve an issue also brought up in #859 .

This PR introduces warp::filters::header::exact_constant_time
A way to filter by an exact head value that will always take the same amount of time, preventing Timing Attack.

While originally closed with the reason that user secrets should be hashed anyways,
i believe that this filter still has a reason to exist as often api-keys are provided via header and set in the programm via env variables. This use case is very common in micro services, k8s, etc.

### DIFF
--- a/src/filters/header.rs
+++ b/src/filters/header.rs
@@ -189,35 +189,33 @@ pub fn exact_constant_time(
                 let len = value.len();
                 let target_len = incoming_val.len();
 
-                // to prevent out of bounds exceptions later.
-                if len == 0 {
-                    return Err(reject::invalid_header(name));
+                let mut total_diff = match len == target_len {
+                    true => 0,
+                    false => 1,
                 };
-
-                // although we can already check if the strings are equal here
-                // by length but due to the timing constraint we still have to check for
-                // equals by byte
-                let mut is_equal = len == target_len;
-
+            
                 let bytes = value.as_bytes();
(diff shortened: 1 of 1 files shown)


src/filters/header.rs
```diff
@@ -155,6 +155,78 @@ pub fn exact(
...
+
+                // although we can already check if the strings are equal here
+                // by length but due to the timing constraint we still have to check for
+                // equals by byte
+                let mut is_equal = len == target_len;
+
+                let bytes = value.as_bytes();
+                let tgt_bytes = value.as_bytes();
+
+                for n in 0..len {
```
seanmonstar, line 205: Are we sure this will actually compile to constant time? I know there's crates that exist specifically to prevent the compiler from optimizing it...

kleinschrader: You were absolutely correct, the compiler optimized most stuff out when a hard coded string was provided.

I now rewrote the function to prevent the compiler from doing this.


Here i build a test program with the --release flag and disassebled it using idafree.
As you can see there is no way past the loop and it has to completely evaluate each byte no matter what.

seanmonstar:
My feeling is that since there are several opinions on how to do this, and since it's possible to implement it outside warp, I'm leaning towards closing and not providing something out of the box.

Issue #1002, opened 2022-11-04
Poor performance of Rejections::into_response()


alexeiakimov:
**Version**

0.3.3

**Platform**

Linux xxx 5.15.0-52-generic #58-Ubuntu SMP Thu Oct 13 08:03:55 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux

**Description**

warp::reject::Rejections::into_response has poor performance if the Rejections instance is a large combination of
known or custom rejections.

I tried this code:

```rust
#[test]
    fn convert_rejection_into_response() {
        let mut rejections = Rejections::Custom(Box::new(std::io::Error::from_raw_os_error(100)));
        for _ in 0..50 {
            rejections = Rejections::Combined(
                Box::new(Rejections::Known(Known::MethodNotAllowed(
                    MethodNotAllowed { _p: () },
                ))),
                Box::new(rejections),
            );
        }
        let reason = Reason::Other(Box::new(rejections));
        let rejection = Rejection { reason };
        println!("{:?}", rejection);
        // The next line is the problem
        println!("{:?}", rejection.into_response());
    }
```

I expected to see this happen: `rejection.into_response()` returns within 1 second

Instead, this happened: `rejection.into_response()` cannot return within 1,5 minutes

seanmonstar:
1.5 minutes? Wow. That with `--release`?

alexeiakimov:
I would like to add that the test above is not artificial, in the real application I have

1. A composite filter like `f1 or f2 or f3 or ... fn` where each fk is a simple combination of path and HTTP method.
2. A request has a correct path and method, but invalid JSON body which fails to be parsed.

In such a case the rejection is a combination of many MethodNotAllowed and one parsing error.

alexeiakimov:
I would like also to add some analysis of how `into_response` is implemented.
Let `n` be the number of simple (i.e. known or custom) rejections in a simple linear combination. Then let

1. `R(n)` be the time to convert the rejection into response
2. `P(n)` be the time to compute the preferred rejection in worst case
3. `S(n)` be the time to compute the rejection status

Then we have
```
S(n) = P(n) + S(n-1) + C
P(n) = S(n-1) + C = P(n-1) + P(n-2) +...
R(n) = P(n) + R(n-1) = P(n) + P(n-1) + ... = 2*(P(n-1) + P(n-2) +...) = 2*R(n-1)
```
[...]

PR #1005, opened 2022-11-09
#1002 Performance of Rejections::into_response is improved.

alexeiakimov:
The present PR improves the performance of `Rejections::into_response()`. Compared to the original approach it calls `status` only for known or custom instances, which allows to avoid exponential complexity. A test is added to check that the response is computed in a reasonable time, with the original implementation the test fails to complete in 1 minute.

### DIFF
--- a/src/reject.rs
+++ b/src/reject.rs
@@ -436,7 +436,7 @@ impl Rejections {
                 | Known::BodyConsumedMultipleTimes(_) => StatusCode::INTERNAL_SERVER_ERROR,
             },
             Rejections::Custom(..) => StatusCode::INTERNAL_SERVER_ERROR,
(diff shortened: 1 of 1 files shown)


alexeiakimov:
The examples code does not compile with Rust 1.67.0-nightly which complains that `impl Reply` is not enough to satisfy the bound for `Filter::Exract`. The compiler suggests to use `impl Reply + Tuple` instead. This problem is not related to the changes introduced by the present PR and caused by changes in the nightly compiler. @seanmonstar could you please confirm that the compilation issue are not caused by changes being introduced? Thank you.

seanmonstar (approved):
Thanks for improving this!

PR #1007, opened 2022-11-17
Make rustls-pemfile optional

glandium:


glandium:
The same patch was merged in #1069.

PR #1010, opened 2022-12-14
Style: handlebars template clone in map

optimizasean:
- Adding clone in map for newer users (to make reusable)
- Added note in comment on reusable enclosure to clone


optimizasean:
Kind of an unimportant edit, but will help newer rust people like me since I couldn't figure out why the "reusable closure" was not reusable. Thought I had to download some other library or enable some nightly features or something since they aren't "reusable" in essence.

Thanks to @haloopdy in the warp discord (in the tokio server) for helping me understand where I was stuck.

PR #1020, opened 2023-01-30
Replace unmaintained/outdated github actions

nickelc:
The toolchain is now installed with [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,21 +12,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v3
 
       - name: Install rust
-        uses: actions-rs/toolchain@v1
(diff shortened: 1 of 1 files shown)


jxs:
Hi, and thanks for this!
Interesting, it seems `rust 1.66` broke `warp`'s `impl Reply` on return types. Opened https://github.com/rust-lang/rust/issues/107729 CC @seanmonstar

seanmonstar:
I think we can solve this by changing the types to `(impl Reply,)`.

SylvainGarrigues:
> I think we can solve this by changing the types to `(impl Reply,)`.

This indeed makes the warp examples compile. Any plans to merge these modifications?

jxs:
> > I think we can solve this by changing the types to `(impl Reply,)`.
> 
> This indeed makes the warp examples compile. Any plans to merge these modifications?

would you like to submit a PR addressing that? :)

PR #1021, opened 2023-02-25
[tracing] Bump tracing packages

sgirones:
Bump tracing packages to the latest tags


seanmonstar:
Thanks for the PR! However, there's no need to do this. Cargo will automatically grab the latest version that fits within the semver range.

sgirones:
ah cool, so we just need to remove the patch version on `tracing` and `tracing-subscriber`?

sgirones:
ok, nevermind, I see how it works now. Sorry for the noise

PR #1027, opened 2023-03-17
Added feature to access generic file

Buzzec:
The `generic` file is needed to build functions that involve some filters (ex. `or`) and to allow generic inputs.


Buzzec:
Pr seems to be ignored

PR #1033, opened 2023-03-31
Implement multipart with multer

jaysonsantos:
This is a continuation of #846

@seanmonstar I've managed to implement the change to use multer without resorting to stream::unfold but, for that I had to implement a change on their project to expose the poll next item function.

```
diff --git a/src/multipart.rs b/src/multipart.rs
index 7b69dbb..ea7410a 100644
--- a/src/multipart.rs
+++ b/src/multipart.rs
@@ -233,7 +233,7 @@ impl<'r> Multipart<'r> {
 
     // CORRECTNESS: This method must only be called only when it is guaranteed
     // that `self.state` is an exlusive `Arc`!
-    fn poll_next_field(&mut self, cx: &mut Context<'_>) -> Poll<Result<Option<Field<'r>>>> {
+    pub fn poll_next_field(&mut self, cx: &mut Context<'_>) -> Poll<Result<Option<Field<'r>>>> {
         debug_assert_eq!(Arc::strong_count(&self.state), 1);
         debug_assert!(self.state.try_lock().is_some(), "expected exlusive lock");
         let mut lock = match self.state.try_lock() {

```
So far, `cargo test --features=multipart` seems to work fine using that patch.
I will open a PR there and check if it makes sense.

### DIFF
--- a/src/filters/multipart.rs
+++ b/src/filters/multipart.rs
@@ -12,8 +12,7 @@ use futures_util::{future, Stream};
 use headers::ContentType;
 use hyper::Body;
 use mime::Mime;
-use multiparty::headers::Headers;
-use multiparty::server::owned_futures03::{FormData as FormDataInner, Part as PartInner};
+use multer::{Field as PartInner, Multipart as FormDataInner};
 
 use crate::filter::{Filter, FilterBase, Internal};
 use crate::reject::{self, Rejection};
@@ -33,15 +32,14 @@ pub struct FormOptions {
 ///
 /// Extracted with a `warp::multipart::form` filter.
 pub struct FormData {
-    inner: FormDataInner<BodyIoError>,
+    inner: FormDataInner<'static>,
 }
 
 /// A single "part" of a multipart/form-data body.
 ///
 /// Yielded from the `FormData` stream.
 pub struct Part {
-    headers: Headers,
-    part: PartInner<BodyIoError>,
+    part: PartInner<'static>,
 }
 
 /// Create a `Filter` to extract a `multipart/form-data` body from a request.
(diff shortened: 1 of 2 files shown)


src/filters/multipart.rs
```diff
@@ -131,17 +123,18 @@ impl Stream for FormData {
 impl Part {
     /// Get the name of this part.
     pub fn name(&self) -> &str {
-        &self.headers.name
+        &self.part.name().unwrap_or("not-set")
```
seanmonstar, line 126: Multipart fields exist without a name?

seanmonstar (commented):
Thanks for writing this up so quickly! This is all set to go already?

src/filters/multipart.rs
jaysonsantos: Hey @seanmonstar, on the code there it points to the docs here https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition which says:

```
Only the value form-data, as well as the optional directive name and filename, can be used in the HTTP context.
```

jaysonsantos:
Hey @seanmonstar I guess so at least on my tests it was all good, any idea of edge cases to be tested there?

src/filters/multipart.rs
seanmonstar: The subheading https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_header_for_a_multipart_body points out:

> The first directive is always form-data, and the header _must_ also include a name parameter to identify the relevant field.

jaysonsantos: True.
Your concern is that if the library is not following the RFC or tat I should put a default that makes more sense on that?

jaysonsantos: Either way, I will open an issue there to try to clarify. I am not sure if it is optional because it is used in both to represent both the header of HTTP and the boundary

seanmonstar: Sorry that I didn't explain more. I wasn't sure myself. So, I'm thinking that perhaps even if `multer` doesn't think so yet, we would treat a lack of a name as an error, and thus if the name does exist, we can just use `self.part.name().expect("checked for name previously")`. What do you think?

jaysonsantos: all good but, the panic is ok there?

seanmonstar: Oh, what I was thinking is that constructing a `Part` would require us to check for a `name`, and if it doesn't exist, then `next_field()` would return an `Error`. We wouldn't want an unchecked panic that a client could trigger, no.

src/filters/multipart.rs
```diff
@@ -131,17 +123,18 @@ impl Stream for FormData {
 impl Part {
     /// Get the name of this part.
     pub fn name(&self) -> &str {
-        &self.headers.name
+        self.part.name().expect("checked for name previously")
```
seanmonstar, line 126: Is the check missing? I'd expect it in `poll_next`, after matching that `Some(part)` was found. If the name is missing, then return `Err()`.

seanmonstar: I would change it to:

```rust
Poll::Ready(Ok(Some(part))) => {
    if part.name().is_some() {
        Poll::Ready(Some(Ok(Part { part })))
    } else {
        Poll::Ready(Some(Err(crate::Error::new(MultipartFieldMissingName))))
    }
}
```

And define lower in the field:

```rust
#[derive(Debug)]
struct MultipartFieldMissingName;

impl std::error::Error for MultipartFieldMissingName {}
```

jaysonsantos: Sorry, I just realized that all filters have their errors contained in their files, I naively looked only on error.rs, thinking they would be grouped there.
I just sent the change.

src/filters/multipart.rs
```diff
@@ -207,3 +208,15 @@ impl Stream for BodyIoError {
         }
     }
 }
+
+/// An error used when a multipart field is missing a name.
+#[derive(Debug)]
+pub struct MultipartFieldMissingName;
```
seanmonstar, line 214:
```suggestion
struct MultipartFieldMissingName;
```

I'd keep it private. _Or_, make it have a private field, so we could extend it latter if need be.

PR #1045, opened 2023-06-01
multipart: fallback to filename as part name

wca:
This is not supported by the standard, but allowed by some implementations for compatibility.


wca:
@seanmonstar Hey, any chance you could comment on this, or refer someone who might be interested?

I wasn't sure if this behavior is something that others might prefer to be hidden behind a feature flag, or wrapped, or otherwise didn't materially change behavior for existing users.

It hasn't impacted us yet, but it'd be nice to not maintain a fork to keep this change around.  Thanks for your time.

seanmonstar (approved):
Sure, sounds good 😊

PR #1046, opened 2023-06-03
updated Part.content_type() to return Mime.as_ref() string

### LINKED ISSUE #1042: Part::content_type only returns the media type, not the subtype
**Version**
0.3.5

**Platform**
Linux 6.1.0-6-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.15-1 (2023-03-05) x86_64 GNU/Linux
(but not relevant here)

**Description**
I'm using the `warp::multipart::form` filter.
When parsing the returned FormData, the part's `content_type` method only returns the media type (that is for instance "image" instead of "image/jpeg").

Snippet of my code:
```rust
pub async fn parse_multipart(mut form: FormData) -> Result<(), Rejection> {
    let rej = || warp::reject::custom(crate::endpoints::BadRequest);

    while let Some(Ok(part)) = form.next().await {
        let name = String::from(part.name());
        let content_type = part.content_type().map(|s| String::from(s));

        let data: Vec<u8> = part
            .stream()
            .map_err(|_| ())
            .map_ok(|buf|
                stream::iter(
                    buf.reader().bytes().map(|b| Result::<u8, ()>::Ok(b.unwrap()))
                )
            )
            .try_flatten()
            .try_collect()
            .await
            .or(Err(rej()))?;

        // [Do something with name, content_type and data]
    }

    Ok(()) // Return actual data instead
}
```

The `content_type` variable on line 6 gets set to eg "image" instead of "image/jpeg".

Relevant code in warp: https://github.com/seanmonstar/warp/blob/4c1f7ba8536baef222bbe0a6fac741b82df66018/src/filters/multipart.rs#L142-L146

Here `self.part` is a [`multer::Field`](https://docs.rs/multer/2.1.0/multer/struct.Field.html). It `content_type` method returns a [`Mime`](https://docs.rs/mime/latest/mime/struct.Mime.html). The `Mime::type_` method only returns the "toplevel media type".

We need a new method to return the whole `Mime` struct.
Also the current `Part::content_type` should probably be changed to return `t.to_string()` instead (or `t.essence_str()`).

**Workaround**
For now, I just deduce the full MIME type from the filename instead.

aedmondson:
Probably resolves #1042.

### DIFF
--- a/src/filters/multipart.rs
+++ b/src/filters/multipart.rs
@@ -140,8 +140,9 @@ impl Part {
     }
 
     /// Get the content-type of this part, if present.
-    pub fn content_type(&self) -> Option<&Mime> {
-        self.part.content_type()
+    pub fn content_type(&self) -> Option<&str> {
+        let content_type = self.part.content_type();
+        content_type.map(|t| t.as_ref())
     }
 
     /// Asynchronously get some of the data for this `Part`.
--- a/src/filters/multipart.rs
+++ b/src/filters/multipart.rs
@@ -140,9 +140,8 @@ impl Part {
     }
 
     /// Get the content-type of this part, if present.
-    pub fn content_type(&self) -> Option<&str> {
-        let content_type = self.part.content_type();
-        content_type.map(|t| t.type_().as_str())
+    pub fn content_type(&self) -> Option<&Mime> {
+        self.part.content_type()
     }
 
     /// Asynchronously get some of the data for this `Part`.


seanmonstar:
Thanks for the PR!

This would be a breaking change, though. A fix could be to change the function to not call `type_()`, just return the full string.

aedmondson:
> 

I see, misunderstood the comments on the issue. I'll change that now.

seanmonstar (approved):

Issue #189, opened 2019-03-14
TLS via native-tls


ignatenkobrain:
In Fedora we don't want to package yet another crypto libraries (in this case I am talking about `ring`) so we want to stick to openssl as our TLS backend. It would be great if you could use `native-tls` for TLS implementation.

Thanks in advance!

sarmahaj:
Currently we are using warp as our web framework in FDO implementation of rust. We are also facing similar problem with 'ring' crate not being FIPS ( Federal Information Processing Standards ) compliant. 
I was following some of the PRs that were created for supporting TLS options. Can someone pls point if 'native-tls' or 'hyper-tls' is planned for support or is there a way where warp can be used so as avoid rustls ?
Thanks !

sarmahaj:
@asonix Thanks ! will try that.

msabansal:
We also had a similar requirement to use openssl as the TLS layer in one of our rust based webservers. We built a compatibility layer for it and have open sourced it at https://github.com/Azure/warp-openssl 

Any feedback is appreciated. Docs will be improved in coming weeks.

PR #1047, opened 2023-06-30
Changes to support openssl for TLS

msabansal:
Added openssl as an alternative library to be used for TLS. This is useful in cases where rustls is not acceptable such as https://github.com/seanmonstar/warp/issues/189


msabansal:
@seanmonstar What do you think about this change? Given that you can achieve something like this with hyper::server::conn::Http::serve_connection along with warp::service?

seanmonstar:
So far, I haven't wanted to make the TLS options in warp that extensive. It's meant as a convenience, but for those wanting more control, or different backends, a warp filter can be converted into a `Service` and used with any TLS acceptor + hyper then.

PR #1049, opened 2023-07-02
Implement `warp::reply::file` for dynamic file service

Molkars:
This pull request implements `warp::reply::file`, `warp::header::conditionals`, and publishes `warp::header::Conditionals`. I also internally moves Conditionals to what I believe is a more fitting place (from fs to header).

Fixes https://github.com/seanmonstar/warp/issues/171
Supersedes https://github.com/seanmonstar/warp/pull/924

### DIFF
--- a/examples/file.rs
+++ b/examples/file.rs
@@ -1,6 +1,7 @@
 #![deny(warnings)]
 
 use warp::Filter;
+use warp::header::Conditionals;
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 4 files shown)


seanmonstar:
Awesome work, thanks! I think I have mainly one suggestion: make `Conditionals` optional. `reply::file(path)` could return `File`, which has an inherent method `conditionals()` that takes the conditional arguments. What do you think?

Molkars:
I'm not sure that's very feasible since we return a future dependent on the value initially passed in. We could have a reply::conditional_file though

mvdschee:
I've been using Molkar's version for a while now, and it works perfectly for my needs. How can we make this PR land into the main version?

Latawiec:
Any updates here? I'd love to have this feature

alexander-camuto:
> Any updates here? I'd love to have this feature

same

nickbp:
I'd independently hit this very issue when trying to build a file server that couldn't serve files. I'd imagined that warp would have the building blocks available to allow DIYing this, but AFAICT the only solution right now is to copy/paste a bunch of `pub(crate)` code out of warp. I got pretty far in doing that, but realized that the exercise had become more hassle than it was worth.

If anyone else stumbles upon this issue after hitting a brick wall like I had, I ended up just switching to using this instead: https://api.rocket.rs/master/rocket/fs/struct.FileServer.html

PR #1056, opened 2023-07-29
More detailed redirects documentation

louiseyousre200:
Added a description, usage and common use cases for each one of the redirects.


seanmonstar (approved):
Thank you!

PR #1061, opened 2023-08-05
Add example (and doc reference) for warp::body::stream()

nickbp:
As of right now it didn't look like there was example usage for this functionality. It took me some trial and error to figure out the typing for the Stream so that I could pass it into a handler function like this. This example hopefully avoids similar headaches for others.


seanmonstar (approved):
Thank you!

PR #1062, opened 2023-08-19
Add references to `Filter` in documentation

belgoking:
A number of functions return structs that (indirectly) implement the `Filter` trait. After not finding an according reference in the documentation for quite a while, I had to look into the code to actually figure out what was going on. In this pull-request I've added some references to the Filter trait in hope it will help point readers into the right direction.


seanmonstar (approved):
Seems like a nice improvement, thanks :)

PR #1075, opened 2023-10-24
Feat add tcp keepalive configuration

### LINKED ISSUE #854: Add 2h keepalive for TCP sessions by default

In many cases unused TCP sessions are left to ESTABLISHED state, e.g. by a firewall that drops the connection. Server endpoint has no way of identifying the situation if no traffic is sent over the TCP session. TCP keepalives [RFC1122, sec. 4.2.3.6](https://datatracker.ietf.org/doc/html/rfc1122#page-101) is an existing mechanism that handles this issue and drops unused connections when they are identified. 2 h keepalive is the minimum time that must be used if enabled by default.

This happens quite often on the internet with public servers so implementing the default is justified to avoid unnecessary resource consumption by hanged connections.

TCP sessions should implement default keepalive as described in RFC 1222.

The alternative would be to implement some kind of monitoring script or similar which tries to identify unused connections from statistics. In most cases, this kind of heuristics is troublesome to implement correctly. Firewalls, like iptables, do not provide the means to drop unused but valid ESTABLISHED TCP connections.

**Additional context**

I have implemented the feature for 0.1.x branch in PR #853 and tested in live. Works great.

bburnichon:
I came across the same issue as #853 (albeit with another keepalive value). It would then also allowing to fix #854

This PR solves it via adding a way to configure tcp keepalive to whatever needed.

For tls, you would need to change the configuration before switching to tls configuration.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ async-compression = { version = "0.3.7", features = ["tokio"], optional = true }
 bytes = "1.0"
 futures-util = { version = "0.3", default-features = false, features = ["sink"] }
 futures-channel = { version = "0.3.17", features = ["sink"]}
-headers = "0.3"
+headers = "0.3.5"
 http = "0.2"
 hyper = { version = "0.14", features = ["stream", "server", "http1", "http2", "tcp", "client"] }
 log = "0.4"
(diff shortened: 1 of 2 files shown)


jq-rs:
Configurable keepalive is way better than any default value, looks good!

jq-rs (approved):

seanmonstar:
With the latest release, warp provides just some simple builder options, and otherwise suggests if you need advanced customization to build a listener manually and just use warp filters as a service.

PR #1092, opened 2024-02-08
Honor SSLKEYLOGFILE by default.

llebout:
tokio_rustls already enables rustls's logging feature therefore we can enable this so that debugging warp TLS servers is simple and easy out of the box. Security concerns of enabling by default are limited since if you can alter environnement you can probably also read warp's server memory and achieve the same thing.


seanmonstar (changes requested):
I dunno about being enabled by default, but a builder option to enable it could be fine.

PR #1099, opened 2024-04-07
deps: update to rustls 0.23

tottoto:
Updates to rustls 0.23.


elichai:
*0.26

elichai:
@seanmonstar Any chance this could get in? 0.23 tore the ecosystem apart and `warp` is one of only highly used libraries that still use 0.22

seanmonstar (approved):

PR #1102, opened 2024-07-23
chore(deps): update handlebars to v6

campeis:


campeis:
looks tls test are failing because the ecc cert is in V1 format and not V3 format.
should be fixed by #1103

seanmonstar (approved):
Thanks so much!

PR #1109, opened 2024-12-16
Provide warp::cloned to simplify passing shared services to routes

andyHa:
Passing shared services to routes is a very common task. Most of the time these implement `Clone`
as they're most probably wrapped as `Arc` anyway. Using `with_cloneable` is a bit shorter and
more concise than `crate::any().map(move || value.clone())`.

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -127,6 +127,29 @@ impl Reply for Json {
     }
 }
 
+/// Converts a [`Stream`](futures_util::Stream) of data chunks into a reply.
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
+/// use futures_util::stream;
+/// use bytes::Bytes;
+///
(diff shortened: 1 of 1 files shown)


seanmonstar:
Sorry, I hadn't been paying attention to the notifications here, I was way too busy with other things. I've finally gotten back to fixing up warp some.

I appreciate the goal here, and I admit I'm slightly torn. On the one hand, since `warp::any().map(move || val.clone())` works, I kind of like it being the "one way". But I recognize it requires understanding several concepts that are somewhat unrelated to just "give me clones of this thing".

If I did merge something like this, I think I'd lean towards a different name. Possibly `warp::cloned(val)`, maybe something else fluid like that. The term `with` is already used for wrapping filters, and `cloneable` is closer to Java-speak than Rust normally is. Anyways, that's my thoughts 🫣

PR #1114, opened 2025-01-22
Fixes Issue#1113: TLS: Ignore certs in pem file when loading private key

jdthomas:
This ignores just certs in the file. Options I considered: 
1. Ignore all the non-key items. replacing with `_ => {},`

2. Ignore explicitly the existing entries in `rustls_pemfile::Item::*` that are not a key. replacing with `rustls_pemfile::Item::X509Certificate(_) | Crl(_) | Csr(_) => {},`

3. Ignore just the certs `X509Certificate(_) => {}, _ => return Err(TlsConfigError::UnknownPrivateKeyFormat),`

Any of those would resolve my issue of certs + keys in same file, happy to adjust diff as needed.


jdthomas:
Looks like rusttls-pemfile also has a `private_key` fn similar to the `certs` fn used at the start of the builder. Perhaps that is what should be used instead of the whole match?

https://docs.rs/rustls-pemfile/latest/rustls_pemfile/fn.private_key.html

PR #1121, opened 2025-08-04
feat: upgrade to hyper v1

seanmonstar:
This upgrades to hyper v1 from v0.14. That in itself causes some breaking changes. But besides that obvious, some of warp's API was changed to better accommodate the upgrade:

- The `tls` feature has been removed (it could be added again, but this was simpler for now.)
- The `Server` type had many of its methods removed in favor of a few simple ones. If you want to customize the behavior significantly, you are better off creating your own custom accept loop. The main selling point of warp is the `Filter` system, not the `Server` type.
- The `warp::addr` module was disabled (will also affect logs) (and the related `Transport` trait).


Gitopolis:
Could you pls add the `tls` feature back? I'm using warp for kubernetes admission webhooks, and tls is required by kubernetes, which means I can't use new versions of warp

PR #1133, opened 2025-08-16
deps: make hyper optional dependency

tottoto:
Makes hyper optional dependency.


seanmonstar (approved):
Thanks! This was an oversight, so I'm fine with the "breakage".

PR #1142, opened 2025-10-02
docs: replace doc_auto_cfg config with doc_cfg

tottoto:
`doc_auto_cfg` was merged into `doc_cfg` in https://github.com/rust-lang/rust/pull/138907.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@
 #![deny(missing_debug_implementations)]
 #![deny(rust_2018_idioms)]
 #![cfg_attr(test, deny(warnings))]
-#![cfg_attr(docsrs, feature(doc_auto_cfg))]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -89,10 +89,8 @@ jobs:
       # check for missing docs
       - name: cargo check
         run: cargo check --all-features
-        env:
-          RUSTFLAGS: "--cfg docsrs"
```
seanmonstar, line 93: I feel like this line was important for something... There's a builder that has docs exposed but otherwise the type is unnameable, and I added this to catch that all methods were documented correctly. I think?

tottoto: Migrated this checking to the building document process. It could be detected as https://github.com/seanmonstar/warp/actions/runs/18195955801/job/51802257485#step:4:202.

PR #1148, opened 2026-01-14
Provides warp::reply::stream

### LINKED ISSUE #1136: Cant return streams or other custom body types anymore in v0.4

At the moment, the returnable body types in 0.4 are severely limited. This prohibits users from i.e. returning streams, which was possible in version 0.3.

I'd like some way to return other data types, like streams, as a reply.

**Additional context**
This is an issue regarding the problem stated in the pull request #1132. I thought an issue might be more fitting to track work on this. Feel free to delete if that is not the case or I overlooked a duplicate issue.

andyHa:
Before 0.4 a stream could by sent as body via
Body::wrap_stream. This has been made
private to not expose any internal workings.

With this change, replying with a stream is
possible again without exposing any
internals.

Fixes: #1136

### DIFF
--- a/src/reply.rs
+++ b/src/reply.rs
@@ -127,6 +127,29 @@ impl Reply for Json {
     }
 }
 
+/// Converts a [`Stream`](futures_util::Stream) of data chunks into a reply.
+///
+/// # Example
+///
+/// ```
+/// use warp::Filter;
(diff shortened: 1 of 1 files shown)


seanmonstar (approved):
Thanks!

PR #1149, opened 2026-01-24
feat: reimplement `warp::addr::remote()` filter

### LINKED ISSUE #1127: Suggest an alternative to `warp::addr::remote` in changelog
I'm trying to upgrade my project to warp 0.4.0. It currently uses `warp::addr::remote`.

It'd be nice if the changelog could suggest an alternative or explain that getting the remote address can't be achieved in warp 0.4.

I found the stub
https://github.com/seanmonstar/warp/blob/53bc811f68545baff8609b80c799e7048f310623/src/filters/addr.rs#L24-L28
which is not leading to a compilation failure because its module definition is commented out
https://github.com/seanmonstar/warp/blob/53bc811f68545baff8609b80c799e7048f310623/src/filters/mod.rs#L6

gamgi:
Reimplements `warp::addr::remote()` filter and enables the `warp::addr` module, which was disabled during the upgrade to hyper v1.

Let me know if there's something missing, or you prefer a different implementation.

Closes https://github.com/seanmonstar/warp/issues/1127.

### DIFF
--- a/src/filters/addr.rs
+++ b/src/filters/addr.rs
@@ -3,6 +3,8 @@
 use std::convert::Infallible;
 use std::net::SocketAddr;
 
+use futures_util::future;
+
 use crate::filter::{filter_fn_one, Filter};
 
 /// Creates a `Filter` to get the remote address of the connection.
@@ -22,7 +24,5 @@ use crate::filter::{filter_fn_one, Filter};
 ///     });
 /// ```
 pub fn remote() -> impl Filter<Extract = (Option<SocketAddr>,), Error = Infallible> + Copy {
-    // TODO: should be replaced with just simple extensions insert by a
-    // make service and then gotten again here.
-    //filter_fn_one(|route| futures_util::future::ok(route.remote_addr()))
+    filter_fn_one(|route| future::ok(route.extensions().get::<SocketAddr>().cloned()))
 }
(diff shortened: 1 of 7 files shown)


src/filters/addr.rs
```diff
@@ -22,7 +24,5 @@ use crate::filter::{filter_fn_one, Filter};
 ///     });
 /// ```
 pub fn remote() -> impl Filter<Extract = (Option<SocketAddr>,), Error = Infallible> + Copy {
-    // TODO: should be replaced with just simple extensions insert by a
-    // make service and then gotten again here.
-    //filter_fn_one(|route| futures_util::future::ok(route.remote_addr()))
+    filter_fn_one(|route| future::ok(route.extensions().get::<SocketAddr>().cloned()))
```
seanmonstar: I think this probably should be a separate newtype, `pub(crate) struct RemoteAddr(SocketAddr)`. That way, it doesn't conflict with anyone else setting that extension, and it allows other named extensions for socket address with other uses.

seanmonstar (commented):
Thanks, looks good, just one thought inline.

src/filters/addr.rs
gamgi: Good point, updated to use a newtype.

seanmonstar (approved):

PR #1151, opened 2026-01-28
feat: reexpose `Info::remote_addr()` in log filter

### LINKED ISSUE #1150: warp::log::custom parameter missing remote_addr() method in version 0.4.2
Environment:
Warp Version: 0.4.2
Hyper Version: 1.0

OS: Windows 10 (Edge/Chromium environment)

Problem: When upgrading to Warp 0.4.2, the warp::log::Info struct no longer provides the remote_addr() method. This prevents custom loggers from capturing the client's IP address directly from the info object. Additionally, attempting to use the warp::addr::remote() filter to manually pass the IP to a handler results in a "failed to resolve" error because the addresses and tls features appear to be missing or reorganized in the 0.4.2 release.

Steps to Reproduce:

Use warp = "0.4.2" with features = ["server"].

Define a custom logger:

Rust
let log = warp::log::custom(|info| {
    println!("{:?}", info.remote_addr()); // Throws: no method named `remote_addr` found
});
Attempt to add warp::addr::remote() to a filter chain.

Try to add features = ["addresses"] or features = ["tls"] to Cargo.toml.

Expected Behavior: warp::log::Info should provide a way to access connection metadata (Remote IP), or the addresses feature should be explicitly available/documented for the 0.4.x branch.

Actual Behavior:

remote_addr() is missing from Info.

warp::addr module is private or unresolved.

Feature flags like addresses are reported as non-existent by Cargo for version 0.4.2.

Current Workaround: Downgrading to Warp 0.3.5 and using the tls feature allows the code to compile and provides access to remote_addr().

gamgi:
Reexposes `Info::remote_addr()` in the default log filter and for custom log filter usage. These were previously disabled during the upgrade to hyper v1.

Let me know if there's something missing.

Closes https://github.com/seanmonstar/warp/issues/1150.

### DIFF
--- a/src/filters/log.rs
+++ b/src/filters/log.rs
@@ -1,6 +1,7 @@
 //! Logger Filters
 
 use std::fmt;
+use std::net::SocketAddr;
 use std::time::{Duration, Instant};
 
 use http::{header, StatusCode};
@@ -35,7 +36,8 @@ pub fn log(name: &'static str) -> Log<impl Fn(Info<'_>) + Copy> {
         // - response content length?
         log::info!(
             target: name,
-            "\"{} {} {:?}\" {} \"{}\" \"{}\" {:?}",
+            "{} \"{} {} {:?}\" {} \"{}\" \"{}\" {:?}",
+            OptFmt(info.route.remote_addr()),
             info.method(),
             info.path(),
             info.route.version(),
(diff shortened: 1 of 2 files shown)


gamgi:
Thanks for the quick turn around previously; decided to open this PR as follow-up to my previous one since it's related, and since someone was asking for these features.

seanmonstar (approved):
Thanks!

PR #1154, opened 2026-03-17
Remove `Sync` requirement on `Body` and `Stream` objects

### LINKED ISSUE #1153: Relax `Sync` requirements on `Body` type
I believe we can remove `Sync` from `Body`'s trait bounds.
This follows the pattern that other libraries have taken such as `tonic` and `axum` which have Body types that do not require `Sync`.

Doing this allows `warp::service` to be more compatible across the Tower ecosystem. For example, if you really wanted to, you could do arcane things like:

```rust 
// Build an Axum router which passes requests through warp filters
let axum_router = axum::Router::new().fallback_service(warp::service(warp_filters));
// Previously this would fail since Axum's `Body` type is not `Sync`
```

macladson:
Fixes #1153

Remove `Sync` from `Body`'s trait bounds.
Also removes `Sync` from the various `Stream` objects as this also appears safe.

### DIFF
--- a/src/bodyt.rs
+++ b/src/bodyt.rs
@@ -5,10 +5,10 @@ use bytes::Buf;
 use bytes::Bytes;
 use futures_util::StreamExt;
 use http_body::Frame;
-use http_body_util::{combinators::BoxBody, BodyExt};
+use http_body_util::{combinators::UnsyncBoxBody, BodyExt};
 
 #[derive(Debug)]
-pub struct Body(BoxBody<Bytes, crate::Error>);
+pub struct Body(UnsyncBoxBody<Bytes, crate::Error>);
 
 impl Default for Body {
     fn default() -> Self {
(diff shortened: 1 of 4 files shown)


src/bodyt.rs
```diff
@@ -5,10 +5,10 @@ use bytes::Buf;
 use bytes::Bytes;
 use futures_util::StreamExt;
 use http_body::Frame;
-use http_body_util::{combinators::BoxBody, BodyExt};
+use http_body_util::{combinators::UnsyncBoxBody, BodyExt};
 
 #[derive(Debug)]
-pub struct Body(BoxBody<Bytes, crate::Error>);
+pub struct Body(UnsyncBoxBody<Bytes, crate::Error>);
```
seanmonstar, line 11: I think this would be a breaking change. `Body` no longer would satisfy `Sync` bounds. So, if somewhere else someone was putting `warp::Body` into a `BoxBody`, this change will cause compiler errors.

macladson: Ah yes, good point. This would have to go into a major release

PR #1158, opened 2026-04-15
fix: Don't treat OPTIONS as CORS preflight if request-method not set

tkanakamalla:
As per the spec (https://fetch.spec.whatwg.org/#http-requests) and the MDN docs (https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request), a CORS preflight request should include at least the Origin as well as Access-Control-Request-Method headers in the OPTIONS request.

So we should not consider an OPTIONS request as a CORS preflight request if it does not include the Access-Control-Request-Method header and instead return Validated:::NotCors, so it will pass through

### DIFF
--- a/src/filters/cors.rs
+++ b/src/filters/cors.rs
@@ -371,9 +371,9 @@ impl Configured {
                     }
                 } else {
                     tracing::trace!(
-                        "preflight request missing access-control-request-method header"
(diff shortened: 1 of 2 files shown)


seanmonstar (approved):
Makes sense, thanks!

PR #1164, opened 2026-05-09
Remove unnecessary clone when creating a new `TlsStream`

FranciscoTGouveia:
Since the `pin` variable will not be used later, it can be moved rather than cloned.

CC @nunoplopes


FranciscoTGouveia:
If possible, could I get a review on this, @seanmonstar?

seanmonstar (approved):
Thanks :)

PR #1167, opened 2026-06-16
Normalize absent path_and_query to root

### LINKED ISSUE #1166: `warp::path::full()` panics for authority-form CONNECT requests with `http` 1.4.1
**Version**
<=0.4.3

**Platform**
x86_64 GNU/Linux

**Description**
`http` 1.4.1 changed `PathAndQuery::from_static()` validation to reject inputs that do not start with `/`:
https://github.com/hyperium/http/blob/v1.4.1/CHANGELOG.md#141-may-25-2026

This exposes a panic in `warp`  path handling for authority-form CONNECT requests.

  - warp falls back to `PathAndQuery::from_static("")` when `uri.path_and_query()` is absent:
    https://github.com/seanmonstar/warp/blob/v0.4.3/src/filters/path.rs#L464-L470

  - `warp::path::full()` reaches that helper:
    https://github.com/seanmonstar/warp/blob/v0.4.3/src/filters/path.rs#L415-L417

  - `warp::service()` is a public API for converting a `Filter` into a `Tower` Service:
    https://github.com/seanmonstar/warp/blob/v0.4.3/src/filter/service.rs#L39-L47

  - `http` 1.4.1 rejects empty path/query input and paths that do not start with `/`, `?`, or `#`:
    https://github.com/hyperium/http/blob/v1.4.1/src/uri/path.rs#L408-L421

  - `http` 1.4.1’s `PathAndQuery::from_static()` panics when scanning rejects the input:
    https://github.com/hyperium/http/blob/v1.4.1/src/uri/path.rs#L60-L72

 **Why I think this is a `warp` issue and not an `http` issue**

An authority-form CONNECT request target has an authority, not a path. In `http`’s public API, that is represented as an absent `path_and_query`.

That representation seems right: there is no path/query component to return, because it has an authority not a path. The problematic part is not that `http` returns `None`, it's that `warp` handles the absence by constructing a synthetic `PathAndQuery` from the empty string

  `PathAndQuery::from_static("")`

After `http` 1.4.1 the empty string is no longer accepted as a path/query value. If `warp` wants path filters to treat an absent path/query as the root path, it likely needs to normalize that absence explicitly to `/` rather than relying on an empty `PathAndQuery` sentinel.

  ## Repro

  `Cargo.toml`:

 ```
 [dependencies]
  bytes = "1"
  http = "=1.4.1"
  http-body-util = "0.1"
  tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
  tower-service = "0.3"
  warp = { version = "=0.4.3", default-features = false }
```

 `src/main.rs`:

 

```rust
  use bytes::Bytes;
  use http::Request;
  use http_body_util::{BodyExt, Empty};
  use tower_service::Service;
  use warp::Filter;

  #[tokio::main]
  async fn main() {
      let route =
          warp::path::full().map(|path: warp::path::FullPath| path.as_str().to_owned());
      let mut service = warp::service(route);

      let request = Request::builder()
          .method("CONNECT")
          .uri("example.com:443")
          .body(Empty::<Bytes>::new())
          .unwrap();

      let uri = request.uri().clone();
      println!(
          "request uri={uri:?} path={:?} path_and_query={:?}",
          uri.path(),
          uri.path_and_query(),
      );

      let response = service.call(request).await.unwrap();
      let status = response.status();
      let body = response.into_body().collect().await.unwrap().to_bytes();

      println!(
          "status={status} body={:?}",
          std::str::from_utf8(&body).unwrap(),
      );
  }
```

Result:

With `http` "=1.4.1", this reaches an authority-form URI with no path/query:

  `request uri=example.com:443 path="" path_and_query=None`

Then warp panics through its fallback to `PathAndQuery::from_static("")`:

  `static str is not valid path`

For comparison, changing only http to =1.4.0 avoids this panic and warp::path::full() observes /.

Should warp normalize an absent path_and_query to / instead of "" in its path filter handling, or is there another layer where you would prefer authority-form requests to be handled?

jlwass:
Fixes `warp::path::full()` panicking for authority-form `CONNECT` requests when using newer `http` (>=1.4.1) versions that reject `PathAndQuery::from_static("")`.

  When a request URI has no `path_and_query`, normalize it explicitly to `/` instead of constructing a `PathAndQuery` from the empty string.

  Closes #1166.

  Tested:
  - `cargo fmt -- --check`
  - `cargo test full_path --features test`

### DIFF
--- a/src/filters/path.rs
+++ b/src/filters/path.rs
@@ -466,7 +466,7 @@ fn path_and_query(route: &Route) -> PathAndQuery {
         .uri()
         .path_and_query()
         .cloned()
-        .unwrap_or_else(|| PathAndQuery::from_static(""))
+        .unwrap_or_else(|| PathAndQuery::from_static("/"))
 }
 
 /// Convenient way to chain multiple path filters together.
--- a/tests/path.rs
+++ b/tests/path.rs
@@ -2,7 +2,11 @@
 #[macro_use]
 extern crate warp;
 
+use bytes::Bytes;
 use futures_util::future;
+use http_body_util::BodyExt;
+use http_body_util::Empty;
+use tower_service::Service;
 use warp::Filter;
 
 #[tokio::test]
@@ -328,6 +332,27 @@ async fn full_path() {
         .await
         .unwrap();
     assert_eq!(ex.as_str(), "/");
+}
+
+#[tokio::test]
+async fn full_path_connect_authority_form() {
+    let route = warp::path::full().map(|path: warp::path::FullPath| path.as_str().to_owned());
+    let mut service = warp::service(route);
+
+    let request = http::Request::builder()
+        .method("CONNECT")
+        .uri("example.com:443")
+        .body(Empty::<Bytes>::new())
+        .unwrap();
+
+    assert_eq!(request.uri().path(), "");
+    assert_eq!(request.uri().path_and_query(), None);
+
+    let response = service.call(request).await.unwrap();
+    assert_eq!(response.status(), http::StatusCode::OK);
+    let body = response.into_body().collect().await.unwrap().to_bytes();
+
+    assert_eq!(&body[..], b"/");
 }
 
 #[tokio::test]


seanmonstar (approved):
Thanks!

PR #1168, opened 2026-07-15
cors: a non-preflight OPTIONS request can still be CORS if it holds an Origin

MathieuDuponchelle:
This is a follow-up to https://github.com/seanmonstar/warp , my actual use case had chrome complaining about the lack of an `Access-Control-Allow-Origin` header in the response of an OPTIONS request sent as part of a [WHEP] exchange: while it is not a preflight request due to the lack of an `Access-Control-Request-Method` it should still be treated as CORS. Fix by defining a new `NotPreflight` variant holding the origin.

[WHEP]: https://datatracker.ietf.org/doc/draft-ietf-wish-whep/

### DIFF
--- a/src/filters/cors.rs
+++ b/src/filters/cors.rs
@@ -349,6 +349,7 @@ enum Validated {
     Preflight(HeaderValue),
     Simple(HeaderValue),
     NotCors,
+    NotPreflight(HeaderValue),
 }
 
 impl Configured {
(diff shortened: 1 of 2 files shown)


seanmonstar:
Thanks! Should this be reflected in the cors integration tests?

MathieuDuponchelle:
yeah, can't hurt I should have thought of that, will do :)

MathieuDuponchelle:
Done :)

